PR #97, opened 2020-10-26
Changes to poms  to support compatibility with JDK9+ module system

bbossola:
In the current build some packages ("javax.xml" and "org.w3c" for example) are accessible via the JRE System Library AND via other jar files pulled by Maven as transitive dependencies, such as xml-apis.jar. Being on the classpath, they are treated as part of the unnamed module, and the "uniquely visible" constraint is violated.

Please see https://issues.alfresco.com/jira/browse/ACS-544 for full detailed explanation

### DIFF
--- a/data-model/pom.xml
+++ b/data-model/pom.xml
@@ -188,7 +188,12 @@
                     <groupId>org.codehaus.woodstox</groupId>
                     <artifactId>woodstox-core-asl</artifactId>
                 </exclusion>
-            </exclusions>
+                <!-- [ACS-544] Excluded as conflicts with JDK9+ as it includes javax.transaction -->
+                <exclusion>
+                  <groupId>org.apache.geronimo.specs</groupId>
+                  <artifactId>geronimo-jta_1.1_spec</artifactId>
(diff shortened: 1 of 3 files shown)


bbossola:
As described in the Jira issue, a patched version of xpp3:xpp3 is also required (the standard one includes the JRE class javax.xml.namespace.QName). It's available as an attachment to the Jira issue and can be easily reproduced removing the offending class from the package.

skopf:
Good catch. Changes look good to me.

I checked the XPP3 project on GitHub and I am afraid we cannot avoid the custom patched XPP3 version. It almost looks like this project is no longer maintained.

However, if we use a custom patched version of XPP3, then we need to have proper documentation and change tracking - even if we just deleted one file.

bbossola:
The pom.xml has been updated: the change for the xpp3:xpp3 package is not part of this PR, it is detailed in https://alfresco.atlassian.net/browse/ACS-823

alandavis (approved):
This should save you and others some pain each time you switch branch.

PR #113, opened 2020-10-30
REPO-5209: Prevent creation of node updated event when property added…

cmshields:
… and removed in same transaction


SaraAspery (approved):
LGTM :)

PR #120, opened 2020-11-03
[ACS-777] Email notification for digital workspace

dhrn:
REST APIs that send email to use an optional client name. The client name is used to lookup the custom email template and its related assets based on the registered configuration. 

https://alfresco.atlassian.net/browse/ACS-777

depends https://github.com/Alfresco/rest-api-explorer/pull/108


.gitignore
```diff
@@ -35,3 +35,6 @@ target
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 
 hs_err_pid*
+
+
+*/alf_data
```
alandavis, line 40: Normally the alf_data directory is not seen. Is this really needed for everyone? Looks like it may be because of local config.

alandavis (commented):
I have had a very fast scan of these changes. However given the amount of new code, I think you really need to go through the PR in detail with someone in your team explaining what it is doing. There is one thing as a reviewer I would want to be sure about.

There should be checks that the new request parameters are not simply used 'as is' in the email. To do so would open us up to security issues such as returning sensitive information on the server or hosting an attack.

Note: As much as don't like the style that does not put a space after key words in Java such as "if(..." "for(...", I can see that that was the style in the files already. You have made the correct choice in not reformatting, as part of this PR. If we were to reformat (please don't) it would be done in a separate PR.

repository/src/main/java/org/alfresco/service/cmr/invitation/Invitation.java
```diff
@@ -97,4 +97,10 @@
     Date getCreatedAt();
     
     Date getModifiedAt();
+
+	/**
+	 * Which client to be sent
+	 * @return the clientName
+	 */
+	String getClientName();
```
jamalkm, line 105: Code standard: It looks like you are using tabs rather than spaces

repository/src/main/java/org/alfresco/service/cmr/invitation/InvitationService.java
```diff
@@ -138,6 +138,19 @@ public NominatedInvitation inviteNominated(String inviteeFirstName, String invit
...
+     *
+     * @param inviteeUserName who is to be invited
+     * @param Invitation.ResourceType resourceType  what resource type ?
+     * @param resourceName which resource
+     * @param inviteeRole which role ?
+     * @param clientName which client
+     * @param workspacePath path of the digital workspace
+     */
+    @Auditable(parameters = { "inviteeComments", "inviteeUserName", "resourceType", "resourceName", "inviteeRole", "clientName", "workspacePath" })
+    public ModeratedInvitation inviteModerated(String inviteeComments, String inviteeUserName, Invitation.ResourceType resourceType, String resourceName, String inviteeRole, String clientName, String workspacePath);
```
jamalkm, line 152: Same as my first comment; There is no need for `workspacePath` parameter.

repository/src/main/resources/alfresco/application-context-highlevel.xml
```diff
@@ -44,4 +44,5 @@
     <import resource="classpath*:alfresco/deprecated-model-bootstrap-context.xml" />
     <import resource="classpath*:alfresco/events-context.xml" />
     <import resource="classpath*:alfresco/messaging-context.xml" />
+    <import resource="classpath:alfresco/invitation-service-context.xml"/>
```
jamalkm, line 47: What is the reason of moving this context file here?

repository/src/main/resources/alfresco/client/config/repo-clients-apps.properties
```diff
@@ -17,4 +17,10 @@ repo.client-app.share.requestResetPasswordTemplatePath=
 # reset password UI page url
 repo.client-app.share.resetPasswordPageUrl=${shareUrl}/page/reset-password
 # reset password confirmation email template path
-repo.client-app.share.confirmResetPasswordTemplatePath=
\ No newline at end of file
+repo.client-app.share.confirmResetPasswordTemplatePath=
+
+### Digital workspace template configurations
+repo.client-app.workspace.inviteModeratedTemplatePath=
+repo.client-app.workspace.sharedLinkBaseUrl=
```
jamalkm, line 24: You don't `sharedLinkBaseUrl` property for  the site invite.

jamalkm (changes requested):

repository/src/main/resources/alfresco/application-context-highlevel.xml
dhrn: This fixes the minimal context run test. this bean is having a couple of dependencies that are not required for the minimal run. so I moved the inviting service here.

eromano (approved):

jamalkm (commented):
The changes look good to me. However, before I approve, I was wondering if you have tested an upgrade scenario? We have to make sure whether a patch is required or not.

remote-api/src/main/java/org/alfresco/rest/api/impl/SiteMembershipRequestsImpl.java
```diff
@@ -270,7 +270,75 @@ public SiteMembershipRequest createSiteMembershipRequest(String inviteeId, final
...
+		String message = siteInvite.getMessage();
+		if(message == null)
+		{
+			// the invitation service ignores null messages so convert to an empty message.
+			message = "";
+		}
+
+		if(siteVisibility.equals(SiteVisibility.MODERATED))
+		{
+			request = inviteToModeratedSite(message, inviteeId, siteId, inviteeRole, client);;
```
jamalkm, line 341: extra semicolon `;;`

dhrn:
I have tested with the upgrade from 6.2.2.1 with the help of an external volume attached to the docker-compose file. I didn't face any issue while going with the latest dev version

jamalkm (approved):

PR #122, opened 2020-11-04
LOC-264 - fixes to punctuation in JAPANESE and SIMP CHINESE

gbroadbent:
Fixes to some strings to correct double byte punctuation


alandavis (approved):
Even with a broken build (tmp problem) I approve this change.

gbroadbent:
Merging JA and SCH bundles with amended double byte punctuation

PR #148, opened 2020-11-12
REPO-5190 Remove Metadata Extractors that are been offloaded in 7.0.0

alandavis:
Removed all the Extractors that now exist in the T-Engines:

- JodConnverterMetadataExtracter
- TikaPoweredMetadataExtracter – the abstract base class used by other extractors
-- MailMetadataExtracter
-- PoiMetadataExtracter
-- TikaAutoMetadataExtracter
-- MP3MetadataExtracter
-- TikaSpringConfiguredMetadataExtracter - removed as it required Spring config and would run in process
-- PdfBoxMetadataExtracter
-- OpenDocumentMetadataExtracter
-- OfficeMetadataExtracter
-- DWGMetadataExtracter
- HtmlMetadataExtracter
- RFC822MetadataExtracter

XmlMetadataExtracter and XPathMetadataExtracter still exist but don't provide any extraction out of the box. The reason they still exist is to support custom transforms (in AMPs) to extract from XML. There are no XML extractors in the T-Engines at the moment, but that is where the custom transformer code really should be moved.

- There are new tests to ensure the async transforms take place as expected.
- Additionally many of the existing tests still exist (those not related to a specific extractor). Some of these have been modified to reflect that the extract is now async and to no longer check the modified value has not changed (it is now expected to change).
- There are also a number of new metadata extract smoke tests that ensure that a selected subset of extracts are supported by the OOTB T-Engines.


jamalkm (approved):
LGTM

PR #149, opened 2020-11-12
MNT-22036 : REST API always applies versioning

Epurashu:
Add versionState form-data property to POST nodes/{nodeId}/children multipart upload.
   Add versionState to MultiPartBuilder.
   Add tests for multipart upload

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
@@ -2914,6 +2914,7 @@ public class NodesImpl implements Nodes
         String versionComment = null;
         String relativePath = null;
         String renditionNames = null;
+        String versioningState = null;
 
         Map<String, Object> qnameStrProps = new HashMap<>();
         Map<QName, Serializable> properties = null;
@@ -2969,6 +2970,10 @@ public class NodesImpl implements Nodes
 
                 case "renditions":
                     renditionNames = getStringOrNull(field.getValue());
+                    break;
+
+            case "versionstate":
+                    versioningState = getStringOrNull(field.getValue());
                     break;
 
                 default:
@@ -3042,11 +3047,27 @@ public class NodesImpl implements Nodes
                     throw new ConstraintViolatedException(fileName + " already exists.");
                 }
             }
-            
+
             // Note: pending REPO-159, we currently auto-enable versioning on new upload (but not when creating empty file)
             if (versionMajor == null)
             {
                 versionMajor = true;
+            }
+
+            // MNT-22036 add versioningState property for newly created nodes.
+            if (null != versioningState)
+            {
+                switch (versioningState)
+                {
+                case "none":
+                    versionMajor = null;
+                    break;
+                case "major":
+                    versionMajor = true;
+                    break;
+                default:
+                    versionMajor = false;
+                }
             }
 
             // Create a new file.
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
@@ -2972,7 +2972,7 @@ public class NodesImpl implements Nodes
                     renditionNames = getStringOrNull(field.getValue());
                     break;
 
-            case "versionstate":
+            case "versioningstate":
                     versioningState = getStringOrNull(field.getValue());
                     break;
 
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
@@ -2984,11 +2984,16 @@ public class NodesImpl implements Nodes
                 case "renditions":
                     renditionNames = getStringOrNull(field.getValue());
                     break;
-
-            case "versioningenabled":
+                case "versioningenabled":
                     String versioningEnabledStringValue = getStringOrNull(field.getValue());
                     if (null != versioningEnabledStringValue)
                     {
+                        // MNT-22036 versioningenabled parameter was added to disable versioning of newly created nodes.
+                        // The default API mechanism should not be changed/affected.
+                        // Versioning is enabled by default when creating a node using form-data.
+                        // To preserve this, versioningEnabled value must be 'true' for any given value typo/valuesNotSupported (except case-insensitive 'false')
+                        // .equalsIgnoreCase("false") will return true only when the input value is 'false'
+                        // !.equalsIgnoreCase("false") will return false only when the input value is 'false'
                         versioningEnabled = !versioningEnabledStringValue.equalsIgnoreCase("false");
                     }
                     break;
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
@@ -1848,6 +1848,19 @@ public class NodesImpl implements Nodes
         {
             versionMajor = Boolean.valueOf(str);
         }
+        String versioningEnabledStringValue = parameters.getParameter("versioningEnabled");
+        if(null != versioningEnabledStringValue)
+        {
+            boolean versioningEnabled = Boolean.parseBoolean(versioningEnabledStringValue);
+            if(versioningEnabled)
+            {
+                versionMajor = (null != versionMajor) ? versionMajor : true;
+            }
+            else
+            {
+                versionMajor = null;
+            }
+        }
         String versionComment = parameters.getParameter(PARAM_VERSION_COMMENT);
 
         // Create the node
@@ -2914,7 +2927,7 @@ public class NodesImpl implements Nodes
         String versionComment = null;
         String relativePath = null;
         String renditionNames = null;
-        String versioningState = null;
+        boolean versioningEnabled = true;
 
         Map<String, Object> qnameStrProps = new HashMap<>();
         Map<QName, Serializable> properties = null;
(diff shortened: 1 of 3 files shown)


gavincornwell (changes requested):
Apologies for this but seeing your unit tests has highlighted the fact that "versioningState" does not makes sense in conjunction with "majorVersion". For example, it's not clear which one takes precedence if you provide both at the same time i.e. versioningState=minor and majorVersion=true.

I propose that we change "versioningState" to "versioningEnabled" instead. This will be a boolean. If versioningEnabled = false the majorVersion is ignored.

I discussed this @Epurashu and we agreed that this feels like the better approach. He will also add the "versioningEnabled" flag to the JSON body and update the API Spec to reflect.

To be clear the default value of "versioningEnabled" for multi-part and JSON will be different to maintain the current behaviour.

Epurashu:
Thank you for your input on implementing this in a very short time @gavincornwell 
I`ve created tests for both Multipart and JSON, at the begging of each test I`ve specified the scenario and expected output.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -1848,6 +1848,19 @@ public Node createNode(String parentFolderNodeId, Node nodeInfo, Parameters para
         {
             versionMajor = Boolean.valueOf(str);
         }
+        String versioningEnabledStringValue = parameters.getParameter("versioningEnabled");
+        if(null != versioningEnabledStringValue)
```
gavincornwell, line 1852: We normally have a space after the "if"

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -2971,6 +2985,14 @@ public Node upload(String parentFolderNodeId, FormData formData, Parameters para
                     renditionNames = getStringOrNull(field.getValue());
                     break;
 
+            case "versioningenabled":
+                    String versioningEnabledStringValue = getStringOrNull(field.getValue());
+                    if(null != versioningEnabledStringValue)
```
gavincornwell, line 2990: We normally have a space after the "if"

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -2971,6 +2985,14 @@ public Node upload(String parentFolderNodeId, FormData formData, Parameters para
...
+                    {
+                        versioningEnabled = !versioningEnabledStringValue.equalsIgnoreCase("false");
```
gavincornwell, line 2992: Any particular reason for testing for "false" and negating the result?

Wouldn't the following be easier to read and understand?

`versioningEnabled = versioningEnabledStringValue.equalsIgnoreCase("true");`

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
```diff
@@ -5612,6 +5612,283 @@ public void updatePropertiesMultivalueTest() throws Exception
...
+        // Test Scenarios:
+        // 1:  majorVersion not set -  versioningEnabled not set  Expect: MAJOR version
+        // 2:  majorVersion not set -  versioningEnabled false    Expect: versioning disabled
+        // 3:  majorVersion true    -  versioningEnabled false    Expect: versioning disabled
+        // 4:  majorVersion false   -  versioningEnabled false    Expect: versioning disabled
+        // 5:  majorVersion not set -  versioningEnabled true     Expect: MAJOR version
+        // 6:  majorVersion true    -  versioningEnabled true     Expect: MAJOR version
+        // 7:  majorVersion false   -  versioningEnabled true     Expect: Minor version
+        // 8:  majorVersion not set -  versioningEnabled False    Expect: versioning disabled
+        // 9:  majorVersion not set -  versioningEnabled False1   Expect: MAJOR version
```
gavincornwell, line 5629: It might be better to use the word invalid here rather than False1 as it makes it look like a typo.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
```diff
@@ -5612,6 +5612,283 @@ public void updatePropertiesMultivalueTest() throws Exception
...
+        // Test Scenarios:
+        // 1: majorVersion not set -  versioningEnabled not set  Expect: versioning disabled
+        // 2: majorVersion not set -  versioningEnabled false    Expect: versioning disabled
+        // 3: majorVersion true    -  versioningEnabled false    Expect: versioning disabled
+        // 4: majorVersion false   -  versioningEnabled false    Expect: versioning disabled
+        // 5: majorVersion not set -  versioningEnabled true     Expect: MAJOR version
+        // 6: majorVersion true    -  versioningEnabled true     Expect: MAJOR version
+        // 7: majorVersion false   -  versioningEnabled true     Expect: Minor version
+        // 8: majorVersion not set -  versioningEnabled False    Expect: versioning disabled
+        // 9: majorVersion not set -  versioningEnabled False1   Expect: versioning disabled
```
gavincornwell, line 5769: It might be better to use the word invalid here rather than False1 as it makes it look like a typo.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
```diff
@@ -5612,6 +5612,283 @@ public void updatePropertiesMultivalueTest() throws Exception
...
+        response = post(getNodeChildrenUrl(myNodeId), reqBody.getBody(), null, reqBody.getContentType(), 201);
+        documentResponse = RestApiUtil.parseRestApiEntry(response.getJsonResponse(), Document.class);
+
+        documentProperties = documentResponse.getProperties();
+        assertNull(documentProperties);
+
+        // Scenario 9:
+        fileName = "myfile" + UUID.randomUUID() + ".txt";
+        multiPartBuilder = MultiPartBuilder.create().setFileData(new FileData(fileName, file));
+        multiPartBuilder.setVersioningEnabled("False1");
```
gavincornwell, line 5743: It might be better to use the word invalid here rather than False1 as it makes it look like a typo.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
```diff
@@ -5612,6 +5612,283 @@ public void updatePropertiesMultivalueTest() throws Exception
...
+        response = post(getNodeChildrenUrl(myNodeId), toJsonAsStringNonNull(d1),requestHeaders, null, null, 201);
+        documentResponse = RestApiUtil.parseRestApiEntry(response.getJsonResponse(), Document.class);
+
+        documentProperties = documentResponse.getProperties();
+        assertNull(documentProperties);
+
+        //Scenario 9:
+        d1.setName("testDoc" + UUID.randomUUID());
+        requestHeaders = new HashMap<>();
+        requestHeaders.put("versioningEnabled","False1");
```
gavincornwell, line 5872: It might be better to use the word invalid here rather than False1 as it makes it look like a typo.

gavincornwell (changes requested):
LGTM other than the few very minor comments to address.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
Epurashu: I've decided to take this approach (still evaluating) because the default behavior for multipart upload is to enable auto-version for content, because of this, every request having an invalid value for `versioningEnabled` should be auto-versioned.

equalsIgnoreCase("true"); will return false for everything except the value `true`

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/test/java/org/alfresco/rest/api/tests/NodeApiTest.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
Epurashu: Thank you for spotting this, fixed it in the last commit.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
gavincornwell: OK, as long as there's a legitimate reason, would be good to add that as a comment though for the next reader ;-)

Epurashu: I`ve added a comment, thank you.

jottley (approved):

PR #154, opened 2020-11-16
[ACA-4122] Sites API response to return relation with groups

dhrn:
https://alfresco.atlassian.net/browse/ACA-4122

depends:- https://github.com/Alfresco/rest-api-explorer/pull/110

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/SitesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/SitesImpl.java
@@ -25,6 +25,8 @@
  */
 package org.alfresco.rest.api.impl;
 
+import static org.alfresco.repo.site.SiteServiceImpl.GROUP_SITE_PREFIX;
+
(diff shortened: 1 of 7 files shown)


eromano (approved):

gavincornwell (commented):
To be honest I feel very nervous about this PR but I might not be fully understanding the purpose of the changes.

If I understand correctly you're trying to return the list of groups each site member belongs to and show that group's role within this site!?

Personally, I think we need to review the use case/requirements here, I'm happy to have a discussion if you put something in my calendar.

skopf (changes requested):
We cannot do this. The UX should have been rejected in the first place. You can expect that a user is in 10s of Ks of groups, each giving the user access to this site.

eromano:
can we add this request as an option ?

eromano:
I am closing this PR for now I think we can archive a similar experience in the FE also if is not perfect under the performance side

PR #162, opened 2020-11-17
Read performance improvements for ValueProtectingMap

AFaust:
This PR contains a suggested improvement of the ValueProtectingMap class to be more consistent with expected behaviour (consistent copy-on-write vs. copy-on-write-and-some-reads) as well as - more importantly - provide better read performance on average and specifically for `keySet`, `entrySet` and `values` operations.

Relevant changes:

- removal of protection of map keys - keys can / must never be mutable in any way as this would break map lookup, so protecting keys is unnecessary overhead
- use lazy protection during key / entry / value set handling only on actual write operations instead of bulk cloning / copying "in case" mutable values are accessed
- add finer grained handling of some well-known mutable complex types to avoid/reduce serialisation overhead
- use `protectValue` for access to single property value in `AbstractNodeDAOImpl.getNodeProperty` as finer grained, already publicly accessible API instead of full wrapper instantiation

Comparisons using the micro-benchmarks added to the unit test of course have to be taken with a grain of salt due to slight fluctuations resulting from concurrently running processes / CPU load. In general though, I have observed the following relative changes in my tests (map of mixed 50:50 mutable/immutable values):

- iterating over entry set with read-only access to immutable values: -90 % (baseline of ~75 ns)
- iterating over entry set with read-only access to mutable values: ~ -10% (baseline of ~65 ns)
- iterating over entry set with single update: +25 % (unlikeliest map use; baseline of ~54 ns)
- iterating over key set: -90 % (baseline of ~62 ns)
- iterating over values: +30 % (baseline of ~61 ns)
- getting a single mutable value: between -20 and -40 % (baseline of ~12 ns)
- getting a single immutable value: between +20 and +35 % (baseline of ~0.3-0.4 ns)
- putting a single value: +25 % (baseline of ~65 ns)

Why is this PR potentially relevant? The `MLPropertyInterceptor` sits in front of the `NodeService` and *almost always* (unless thread is marked via `MLPropertyInterceptor.setMLAware(boolean)`) iterates over the complete node property values map (read-only access with mixed mutable / immutable values), and creates a copy of the map with `MLText` instances replaced with the best value for the current `I18NUtil.getContentLocaleLang()`. Assuming a 50:50 or - more likely - 4:1 split between immutable and mutable values in the properties of a node, the changes of this PR with the -90 % / -10 % relative change to read-only entry set iteration access to those types of values would have significant impact on *almost every* regular code flow using `NodeService.getProperties(NodeRef)`.
The instances of performance degradation affect either the more unlikely uses of the map (i.e. puts against the map almost never occur because of the copy created on retrieval by `MLPropertyInterceptor`) or affect the operation with the lowest baseline of all, where fluctuations due to concurrent load/operations alone make a reliable measurement problematic, and likely have more effect in real life cases than the change in question.


jottley:
@tpage-alfresco @AFaust I am doing a review of all open PRs that are over 6 months old.  Is this still relevant or needed?

PR #163, opened 2020-11-17
ACS-778: Fixed IDS authentication component.

jamalkm:
The IDS Auth component may fail with other exceptions (e.g. connection error) not just with user authentication, therefore the fix captures those exceptions and re-throw AuthenticationException so other authentication components in the chain can have a go at authenticating the user.


gavincornwell (approved):
Nice catch Jamal! 

LGTM

PR #164, opened 2020-11-18
ACS-896 Fix community PR builds

CezarLeahu:
- **scripts/travis/init.sh**: execute the docker logins only on branch builds
- **~/.m2/settings.xml**: remove `quay.io` & `docker.io` server definitions
  (the docker logins in the init.sh script should be enough)
- **.travis.yml**: use DockerHub images instead of Quay.io images

### DIFF
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -31,7 +31,6 @@
             </properties>
         </profile>
         <profile>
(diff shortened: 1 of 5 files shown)


cristianturlica (approved):
Changes look good.

PR #169, opened 2020-11-19
REPO-5219 Allow AGS AMP to specify metadata extract mapping (REPO)

alandavis:
Added an optional extractMapping transform option to all metadata extractors to override the default one in the T-Engine.

In the case of the AGS AMP it extends the RFC822MetadataExtracter with its own class to specify a different set of document to system mappings. The class in the repo no longer does extractions, but is now used by the AsynchronousExtractor, which offloads extractions to T-Engines to obtain the mappings if it has been extended that are then passed to the T-Engine.


cmshields (approved):
LGTM!

PR #183, opened 2020-11-30
ACS-953 Unable to upgrade from ACS-7.0-M1 or ACS-7.0-M2 to ACS-7.0-A12

alandavis:
Now that we create internal releases of the form M.m.r-M9 (milestone), M.m.r-A9 (alpha) and M.m.r-RC9 (release candidate) during development and testing, we need to ensure we can upgrade from any of these to any other, as they may occur in any order and also to the final external release M.m.r. We also will allow a change from the final release back to an internal one for regression testing. The code within ModuleComponentHelper which calls this method, checks if it is the same version (0) and then if it is downgrading (> 0), so if they share the same M.m.r part matches AND is one of these formats, we return <0.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/module/ModuleVersionNumber.java
+++ b/repository/src/main/java/org/alfresco/repo/module/ModuleVersionNumber.java
@@ -54,7 +54,7 @@ public class ModuleVersionNumber implements Externalizable
 
     // Matches versions with 3 or 4 parts to their basic number such as 1.2.3 or 1.2.3.4
     // that also optionally have a -A9 -M9 or -RC9 suffix whe 9 is an integer.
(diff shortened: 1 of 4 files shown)


repository/src/main/java/org/alfresco/repo/module/ModuleVersionNumber.java
```diff
@@ -67,11 +73,56 @@ public ModuleVersionNumber(VersionNumber versionCurrent)
         this(versionCurrent.toString());
     }
 
+    /**
+     * Now that we create internal releases of the form {@code M.m.r-M9} (milestone), {@code M.m.r-A9} (alpha) and
+     * {@code M.m.r-RC9} (release candidate) during development and testing, we need to ensure we can upgrade form any
```
cmshields, line 78: Typo: upgrade form any -> upgrade from any

cmshields (approved):
LGTM! Only issue I found was the typo - see comment.

PR #189, opened 2020-12-02
ACS-986 : Azure Connector 1.3 Alpha Release

tzclucian:


cristianturlica (approved):

alandavis (commented):
********************
You also need to instruct dependabot not to upgrade it again or add a test that will fail if upgraded.
********************

tzclucian:
> You also need to instruct dependabot not to upgrade it again or add a test that will fail if upgraded.

Thank you Alan, notified in [this comment](https://github.com/Alfresco/alfresco-community-repo/pull/74#issuecomment-737752561)

PR #190, opened 2020-12-02
ACS-976 Metadata extraction NPE thrown in multi-node cluster using ATS

alandavis:
In a multi repo node cluster, it is possible one node requests a metadata extract via the ATS and
that another responds to it. In this case the responder needs to recreate the TransformRequest. In
this case the target mimetype was wrong, resulting in the TransformRequest being sent to the wrong
code which then through a NPE because there was no replyQueue specified on the TransformRequest.


KikoUK (approved):
LGTM

PR #191, opened 2020-12-02
Fix MNT-21991 customizable personLookupProperties for REST API

nbarithel:
Copy of https://github.com/Alfresco/alfresco-remote-api/pull/866


jottley (approved):

gavincornwell (approved):
Presuming the response is identical i.e. backwards compatible it looks good to me.

nbarithel:
> Presuming the response is identical i.e. backwards compatible it looks good to me.

Yes it is, the same QNames are now defined using xml instead of Java.

PR #193, opened 2020-12-03
Fix/search 2602 missed information search page

aborroy:


repository/src/main/java/org/alfresco/repo/search/impl/solr/SolrChildApplicationContextFactory.java
```diff
@@ -101,118 +114,96 @@ public boolean isUpdateable(String name)
...
     }
-
+    
+    private String getPropertyValue(JSONObject core, String name)
+    {
+        switch (name)
+        {
+            case SolrChildApplicationContextFactory.ALFRESCO_ACTIVE:
+            case SolrChildApplicationContextFactory.ARCHIVE_ACTIVE:
+                return core.getBoolean("Active") ? "true" : "false";
```
killerboot, line 124: Does it make sense to use static string variables for these?

killerboot (approved):

repository/src/main/java/org/alfresco/repo/search/impl/solr/SolrChildApplicationContextFactory.java
aborroy: Refactored this Strings to Constants.

PR #202, opened 2020-12-08
APPS-688: Fixed the issue by using System user instead of the admin user

bogdanioan-gabor:
Fixed the issue by using System user instead of the admin user


alexbalmus (approved):
LGTM

ancutam (approved):

PR #227, opened 2021-01-04
REPO-5191 Bug: T-Engine should provide mapping rather than the repo.

alandavis:
Bug found while reviewing documents on how to create a custom metadata extractor. The original refactor had left the repo doing the mapping. It should have been passing the fully qualified repo properties to the T-Engine to do the mapping.

Linked to:
* https://github.com/Alfresco/acs-packaging/pull/1826
* https://github.com/Alfresco/alfresco-transform-core/pull/316

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/content/metadata/AbstractMappingMetadataExtracter.java
+++ b/repository/src/main/java/org/alfresco/repo/content/metadata/AbstractMappingMetadataExtracter.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2020 Alfresco Software Limited
+ * Copyright (C) 2005 - 2021 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software. 
  * If the software was purchased under a paid Alfresco license, the terms of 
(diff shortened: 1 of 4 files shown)


repository/src/main/java/org/alfresco/repo/content/metadata/AbstractMappingMetadataExtracter.java
```diff
@@ -1462,7 +1463,7 @@ public void embed(
      * @param systemMetadata   Metadata keyed by system properties
      * @return              Returns the metadata keyed by the content file metadata properties
      */
-    private Map<String, Serializable> mapSystemToRaw(Map<QName, Serializable> systemMetadata)
```
alandavis, line 1465: Made protected so it could be overridden in AsynchronousExtractor, where we wish to map to fully qualified QNames name rather than the raw name.

repository/src/main/java/org/alfresco/repo/content/metadata/AbstractMappingMetadataExtracter.java
```diff
@@ -2261,45 +2262,4 @@ protected void embedInternal(Map<String, Serializable> metadata, ContentReader r
     {
         // TODO make this an abstract method once more extracters support embedding
     }
-
-    public static Map<String, String> convertMetadataToStrings(Map<String, Serializable> properties)
```
alandavis, line 2265: This method was introduced in ACS 7, so it is okay to remove it.

eknizat (approved):
LGTM

PR #233, opened 2021-01-08
ACS-1110 replace fabric8 with docker-maven-plugin

mteodori:


CezarLeahu:
Why must the `squash` option be removed?

PR #239, opened 2021-01-13
SEARCH-2624 add an argument to identify a phrase query

jesty:
This PR will be used by https://github.com/Alfresco/alfresco-enterprise-repo/pull/180

### DIFF
--- a/data-model/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/functions/FTSPhrase.java
+++ b/data-model/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/functions/FTSPhrase.java
@@ -47,6 +47,8 @@ public class FTSPhrase extends BaseFunction
     public final static String ARG_PHRASE = "Phrase";
 
     public final static String ARG_PROPERTY = "Property";
+
+    public final static String ARG_IS_PHRASE = "IsPhrase";
     
     public final static String ARG_SLOP = "Slop";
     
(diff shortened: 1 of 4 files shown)


data-model/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/functions/FTSPhrase.java
```diff
@@ -47,6 +47,8 @@
     public final static String ARG_PHRASE = "Phrase";
 
     public final static String ARG_PROPERTY = "Property";
+
+    public final static String ARG_IS_PHRASE = "IsPhrase";
```
killerboot, line 51: This name clashes with ARG_PHRASE. Please change it or add a clear JavaDoc explaining what it means.

killerboot (changes requested):
Please add tests for your change.

data-model/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/functions/FTSPhrase.java
jesty: What do you think about ARG_HAS_QUOTATION_MARKS? Meanwhile, I will add a JavaDoc.

killerboot: I'm not sure, perhaps you can use a short version ARG_HAS_QUOTES.

jesty:
> Please add tests for your change.

@killerboot I didn't found any class where to add tests, so I created a new one: FTSQueryParserTest

jesty:
I close the PR because the change is not more required.

PR #241, opened 2021-01-14
LOC-282 - Partial fix.  12 language bundles updated. 

gbroadbent:
3 languages pending: DA CS and PL


gbroadbent:
Updated bundles with localised UI in 12 languages.

PR #255, opened 2021-01-26
MNT-20775 : deleted-nodes with maxitems parameter take longer

Epurashu:
GetArchivedNodesCannedQuery now will return only the required results (page) based on offset(skipCount) and limit(maxItems).
Offset and limit are set as rowBounds for mybatis.
Added count query for GetArchivedNodesCannedQuery in order to preserve the totalItems.
Because GetArchivedNodesCannedQuery returns only the required results, an extra condition was added to establish `hasMoreItems`

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/node/archive/GetArchivedNodesCannedQuery.java
+++ b/repository/src/main/java/org/alfresco/repo/node/archive/GetArchivedNodesCannedQuery.java
@@ -57,7 +57,8 @@ public class GetArchivedNodesCannedQuery extends AbstractCannedQueryPermissions<
     private CannedQueryDAO cannedQueryDAO;
     private NodeDAO nodeDAO;
 
-    public GetArchivedNodesCannedQuery(CannedQueryDAO cannedQueryDAO, NodeDAO nodeDAO, MethodSecurityBean<ArchivedNodeEntity> methodSecurity, CannedQueryParameters params)
+    public GetArchivedNodesCannedQuery(CannedQueryDAO cannedQueryDAO, NodeDAO nodeDAO,
+                MethodSecurityBean<ArchivedNodeEntity> methodSecurity, CannedQueryParameters params)
     {
         super(params, methodSecurity);
         this.cannedQueryDAO = cannedQueryDAO;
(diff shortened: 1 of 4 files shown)


adinapitul (approved):

ancutam:
Changes look good! I suggest creating a branch in alfresco-entersprise-repo with the same name, to run a build with the changes on enterprise as well.

repository/src/main/java/org/alfresco/repo/node/archive/GetArchivedNodesCannedQuery.java
```diff
@@ -119,4 +120,26 @@ private void preload(List<NodeRef> nodeRefs)
                         + (System.currentTimeMillis() - start) + " msecs");
         }
     }
+
+    @Override
+    protected Pair<Integer, Integer> getTotalResultCount(List<ArchivedNodeEntity> results)
+    {
+        Object paramBeanObj = super.getParameters().getParameterBean();
+        if (paramBeanObj == null)
+            throw new NullPointerException("Null GetArchivedNodes query params");
```
mauriziovitale, line 129: I think the log is generic or misleading

repository/src/main/java/org/alfresco/repo/node/archive/GetArchivedNodesCannedQuery.java
```diff
@@ -119,4 +120,26 @@ private void preload(List<NodeRef> nodeRefs)
...
+
+        // Get parameters
```
mauriziovitale, line 131: do we need this comment ?

mauriziovitale (changes requested):
check the comments, please

repository/src/main/java/org/alfresco/repo/node/archive/GetArchivedNodesCannedQuery.java
jottley: Since the query parameters are required...Maybe it could read as "Required parameters not provided for GetArchivedNodes canned query".

Epurashu:
Thank you @mauriziovitale for suggesting improvements, I have removed the unnecessary comment, improved the error messages a bit and treated a possible `Overflow`

mauriziovitale (approved):

PR #261, opened 2021-01-27
 Aspect and Type api implementation 

dhrn:
https://alfresco.atlassian.net/browse/REPO-5440
https://alfresco.atlassian.net/browse/REPO-5552

depends https://github.com/Alfresco/rest-api-explorer/pull/119


skopf (changes requested):
LGTM, just one small naming issue:
The abstract base class for the "Aspect" in the rest api package is "AbstractModel". What was the idea behind this base class? Where else are you planning to re-use this base class?
The common parent of aspects and types would be a "class" in alfresco speak. So "AbstractClass" would be a better fit.
Also I am trying to understand how you decided what parts should go into the "AbstractModel" and what parts should go into the "Aspect"

remote-api/src/main/java/org/alfresco/rest/api/Aspects.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * #%L
+ * Alfresco Remote API
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
gavincornwell, line 5: Please update the copyright year from 2016 to 2021, same comment for all new files below.

remote-api/src/main/java/org/alfresco/rest/api/impl/AspectsImpl.java
```diff
@@ -0,0 +1,321 @@
...
+        if (queryParameters != null)
+        {
+            AspectQueryWalker propertyWalker = new AspectQueryWalker();
+            QueryHelper.walk(queryParameters, propertyWalker);
+            return new AspectsFilter(propertyWalker.getModelIds(), propertyWalker.getParentIds(), propertyWalker.getMatchedPrefix(), propertyWalker.getNotMatchedPrefix());
+        }
+        return null;
+    }
+
+    private Collection<QName> getModelAspects(String modelId) {
```
gavincornwell, line 156: Please be consistent with the coding standards of this project.

remote-api/src/main/java/org/alfresco/rest/api/impl/AspectsImpl.java
```diff
@@ -0,0 +1,321 @@
...
+        ModelDefinition modelDefinition =  null;
+
+        if(modelId == null)
```
gavincornwell, line 159: Please add a space between the if and opening bracket to be consistent with coding standards for this project.

gavincornwell (changes requested):
Nothing to add to what Stefan said except for a couple of minor coding standards items.

dhrn:
Similar pattern was followed for the CMM end point. i just followed it. 

Now we have `AbstractClass` that have common properties for aspect and type. 
now it is look all the properties are same for aspect and types. 

let me know the comments
Thanks

dhrn:
Here i have adding the slack discussion for other reviewers

current representation  contains `id` `title` `descrption`  `parentId`. previously the convention was `name` `parentName` . so  i can't use the old model. so I have created the new abstract class.

As per the above  i have renamed the AbstractModel to AbstractClass. now after this Here we have two models in the same package

-  `AbstractClass.java`
-  `AbstractClassModel.java` -> used in CMM API


This is little confusing to user. so we renamed `AbstractClassModel` to `AbstractCustomClass` to keep the consistency

after that, we found the `ClassDefinition` was used in Node. so it has properties that can be shared across the 
Aspect and Type . so we have extended the `AbstractClass` from `ClassDefinition` .

here is final implementation of the [AbstractClass](https://github.com/Alfresco/alfresco-community-repo/blob/dev-dhrn-REPO-5440/remote-api/src/main/java/org/alfresco/rest/api/model/AbstractClass.java)

gavincornwell (approved):
LGTM

remote-api/src/main/java/org/alfresco/rest/api/impl/AbstractClassImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        else
+        {
+            int end = Math.min(skipCount + maxItems, totalItems);
+            boolean hasMoreItems = totalItems > end;
+            list = list.subList(skipCount, end);
+            return CollectionWithPagingInfo.asPaged(paging, list, hasMoreItems, totalItems);
+        }
+    }
+
+    public boolean filterByNamespace(ModelApiFilter query, QName type)
```
gavincornwell, line 73: "type" is a mis-leading name for this parameter, "qname" would be better.

skopf (changes requested):
As Jan and Gav already mentioned on the api-explorer commit, we need to be really carful with naming. Since we are using the term "prefix" for this registered namespace prefix (e.g. "cm"), we can easily confuse readers if we use this term "prefix" as well for the prefix of the namespace URI.

remote-api/src/main/java/org/alfresco/rest/api/impl/AbstractClassImpl.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.List;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class AbstractClassImpl<T extends AbstractClass> {
+    static String PARAM_MODEL_IDS = "modelIds";
+    static String PARAM_PARENT_IDS = "parentIds";
+    static String PARAM_URI_PREFIX = "uriPrefix";
```
gavincornwell, line 49: As mentioned in the review of the open api spec, this should really be "namespaceUri) as that's what we're checking see lines 82 and 86.

remote-api/src/main/java/org/alfresco/rest/api/impl/CustomModelsImpl.java
```diff
@@ -1,28 +1,28 @@
...
- * 
- * You should have received a copy of the GNU Lesser General Public License
- * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
- * #L%
- */
+/*
+ * #%L
+ * Alfresco Remote API
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
gavincornwell, line 5: Remember your copyright dates ;-)

gavincornwell (changes requested):

gavincornwell (approved):
LGTM

skopf (approved):

PR #263, opened 2021-01-28
MNT-14308: CMIS - optionally request rendition(s) when uploading docs

montgolfiere:
- add new configurable option to request set of renditions when uploading via CMIS (none by default)
- note: in the (unlikely) event a customer *relied* on previous behaviour (rather than dynamic requests) they can override & set to "doclib"

### DIFF
--- a/repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
+++ b/repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
@@ -231,7 +231,11 @@ public class AlfrescoCmisServiceFactory extends AbstractServiceFactory
     protected AlfrescoCmisService getCmisServiceTarget(CMISConnector connector)
     {
         AlfrescoCmisServiceImpl cmisService = new AlfrescoCmisServiceImpl(connector);
-        cmisService.setCmisRequestRenditionsOnCreateDoc(parseCommaSeparatedSet(getCmisCreateDocRequestRenditionsSet()));
+
+        Set<String> stringSet = parseCommaSeparatedSet(getCmisCreateDocRequestRenditionsSet());
+        logger.info("getCmisServiceTarget: cmis.create.doc.request.renditions.set="+stringSet);
+        cmisService.setCmisRequestRenditionsOnCreateDoc(stringSet);
+
         return cmisService;
     }
 
--- a/repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
+++ b/repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2016 Alfresco Software Limited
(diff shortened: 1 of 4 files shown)


stevenblair-al:
approved by me.

stevenblair-al (approved):

repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
```diff
@@ -58,6 +61,8 @@
     private CMISTransactionAwareHolderInterceptor cmisHolder;
     private AuthorityService authorityService;
 
+    private String cmisCreateDocRequestRenditionsSet = null;
```
nithinnambiar, line 64: String default value is null

repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
```diff
@@ -217,6 +230,26 @@ public CmisService getService(final CallContext context)
...
+            {
+                String entry = st.nextToken().trim();
+                if (!entry.isEmpty())
+                {
+                    stringSet.add(entry);
+                }
+            }
+        }
+        return stringSet;
     }
```
nithinnambiar, line 254: Looks good to me, is it possible to add a unit test case for this?

eknizat (approved):
Looks good. As already mentioned a test would be good.

amohammedalfresco (approved):

repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
montgolfiere: Yes, why not ;-)

repository/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceFactory.java
montgolfiere: I didn't notice any existing CMIS tests for checking *async* renditions. 

@nithinnambiar How would you propose to unit test that option is passed in ok (when not empty) ? Note: I have checked manually via debug. 

Instead, I suggest we add a log info message allowing easy verification both ways. 

Does that sound reasonable for an optional setting (and no other change in functionality) ?

nithinnambiar: Perhaps a small unit test to check if  `cmisRequestRenditionsOnCreateDoc`  in `AlfrescoCmisServiceImpl` is getting populated with right values from the config?

montgolfiere: Seems overkill. Do you have examples of a similar test elsewhere in the codebase - if so, happy to add ;-)

montgolfiere:
As per https://github.com/Alfresco/alfresco-community-repo/pull/263#discussion_r566783094 pending some other similar unit tests examples. 

For now, I've added the INFO message to show what values were populated (including empty set by default).

PR #264, opened 2021-01-28
REPO-5549 Fix: Local transformer names must exist and be unique docli…

alandavis:
…b Read from ... when overriding a Local transform.

* Problem was that there were two conflicting pieces of code in play in the LocalTransformRegistry.register method. One that checked for duplicate T-Engine names and the other that allowed transforms to be overridden if they had the same name.
* The code checking for duplicate names needed an extra clause to only look at T-Engines (they have a T-Engine url associated with them).
* The code that overrode transforms then worked, but still had issues as the supported source to target mimetypes, priorities and max sizes were not cleared.
* It turned out to be simpler to split the original LocalTransformRegistry.register method into two. Extraction a new method into CombinedConfig.removeOverriddenOrInvalidTransformers that discarded invalid or overridden config before the list of supported source to target mimetypes was created rather than try to fix them up later. This is why there appear to be quite a few changes. The reality is that there are not.
* More extensive unit tests were also added.

### DIFF
--- a/data-model/src/main/java/org/alfresco/util/ConfigScheduler.java
+++ b/data-model/src/main/java/org/alfresco/util/ConfigScheduler.java
@@ -57,6 +57,9 @@ public abstract class ConfigScheduler<Data>
     public static class ConfigSchedulerJob implements Job
     {
         @Override
+        // Synchronized has little effect in normal operation, but on laptops that are suspended, there can be a number
+        // of Threads calling execute concurrently without it, resulting in errors in the log. Theoretically possible in
+        // production but not very likely.
         public synchronized void execute(JobExecutionContext context) throws JobExecutionException
         {
             JobDataMap dataMap = context.getJobDetail().getJobDataMap();
--- a/data-model/src/main/java/org/alfresco/util/ConfigScheduler.java
+++ b/data-model/src/main/java/org/alfresco/util/ConfigScheduler.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2020 Alfresco Software Limited
+ * Copyright (C) 2005 - 2021 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software.
  * If the software was purchased under a paid Alfresco license, the terms of
@@ -57,7 +57,7 @@ public abstract class ConfigScheduler<Data>
     public static class ConfigSchedulerJob implements Job
     {
         @Override
-        public void execute(JobExecutionContext context) throws JobExecutionException
+        public synchronized void execute(JobExecutionContext context) throws JobExecutionException
         {
             JobDataMap dataMap = context.getJobDetail().getJobDataMap();
             ConfigScheduler configScheduler = (ConfigScheduler)dataMap.get(CONFIG_SCHEDULER);
--- a/repository/src/main/java/org/alfresco/repo/content/transform/LocalTransformImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/content/transform/LocalTransformImpl.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2020 Alfresco Software Limited
+ * Copyright (C) 2005 - 2021 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software.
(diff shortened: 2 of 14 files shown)


data-model/src/main/java/org/alfresco/util/ConfigScheduler.java
```diff
@@ -57,7 +57,7 @@
     public static class ConfigSchedulerJob implements Job
     {
         @Override
-        public void execute(JobExecutionContext context) throws JobExecutionException
+        public synchronized void execute(JobExecutionContext context) throws JobExecutionException
```
alandavis, line 60: Has no effect in normal operation. On a laptop which is put to sleep, there can be a number of Threads calling execute is quick succession. Make debug a little difficult.

repository/src/main/java/org/alfresco/repo/content/transform/LocalTransformImpl.java
```diff
@@ -70,7 +70,7 @@ public LocalTransformImpl(String name, TransformerDebug transformerDebug,
         checkAvailability();
     }
 
-    private boolean remoteTransformerClientConfigured()
+    public boolean remoteTransformerClientConfigured()
```
alandavis, line 73: Made public for use in a test

repository/src/main/java/org/alfresco/repo/content/transform/LocalTransformServiceRegistry.java
```diff
@@ -187,20 +182,8 @@ public void register(Transformer transformer, Map<String, Set<TransformOption>>
...
             else if (!isPipeline && !isFailover)
             {
-                baseUrl = getBaseUrlIfTesting(name, baseUrl);
-                if (baseUrl == null)
-                {
-                    throw new IllegalArgumentException("Local transformer " + name +
-                            " must have its baseUrl set in " + LOCAL_TRANSFORMER+name+URL+
-                            " Read from "+readFrom);
-                }
                 int startupRetryPeriodSeconds = getStartupRetryPeriodSeconds(name);
```
alandavis, line 187: The 3 removed error check sections above have been moved to CombinedConfig removeOverriddenOrInvalidTransformers which is run before this method.

repository/src/main/java/org/alfresco/repo/content/transform/LocalTransformServiceRegistry.java
```diff
@@ -333,7 +316,7 @@ else if (isPipeline)
     }
 
     // When testing, we need to be able to set the baseUrl when reading from a file.
-    protected String getBaseUrlIfTesting(String name, String baseUrl)
+    public String getBaseUrlIfTesting(String name, String baseUrl)
     {
```
alandavis, line 320: Needed for testing

repository/src/main/java/org/alfresco/transform/client/registry/CombinedConfig.java
```diff
@@ -267,15 +270,118 @@ public void register(TransformServiceRegistryImpl registry)
...
+                registry.logError(msg);
+                combinedTransformers.remove(i--);
+            }
+        }
+        return combinedTransformers;
+    }
+
+    private static int lastIndexOf(String name, List<TransformAndItsOrigin> combinedTransformers, int toIndex)
+    {
+        for (int j = toIndex-1; j >=0; j--)
```
alandavis, line 372: Lists are short (< 100) entries and this is not a frequent or time critical step.

repository/src/test/java/org/alfresco/transform/client/registry/LocalTransformServiceRegistryConfigTest.java
```diff
@@ -339,16 +387,10 @@ public void testJsonConfig() throws IOException
         internalTestJsonConfig(64, 70);
     }
 
-    @Test
```
alandavis, line 342: No longer overrides anything in super class, so remove it.

repository/src/test/java/org/alfresco/transform/client/registry/LocalTransformServiceRegistryConfigTest.java
```diff
@@ -339,16 +387,10 @@ public void testJsonConfig() throws IOException
         internalTestJsonConfig(64, 70);
     }
```
alandavis, line 389: Removing Test as it no longer overrides a super method.

PR #268, opened 2021-02-01
DESKTOPAPP-539 Fix camel config

killerboot:
This is an approach that fixes the SyncServiceAPITests in acs-packaging.
The main point is to instantiate org.apache.camel.component.activemq.ActiveMQComponent with the right connection config.
This PR needs more work:
* The config does not allow SSL connection
* It is not clear if the connection to the broker is correctly pooled and the pool is configured. I noticed 360+ activemq threads in the repository JVM. This is way too much.
* As far as I understand the dependencies contain a full embedded broker (org.apache.activemq:activemq-broker) which we should not pack into repository
* The config in repository/src/main/resources/alfresco/subsystems/Messaging/default/messaging-context.xml needs to be reviewed and cleaned up according to latest Camel lib config. After that the review should be done in enterprise counterpart and dsync amp.


alandavis:
Alex is the point of these changes to replace our own 'special' config (since 2015) with a standard ActiveMQConnectionFactory that is now available with the newer version of Camel?

killerboot:
@alandavis not entirely. It seems that now it is required to instantiate ActiveMQComponent that controls the connection factory. The factory on it's own is not autowired to this component, so it is using it's own default connection factory pointing to localhost all the time.
The reason why I transferred the connection factory config into xml is just to prove that my assumption was correct. I'm not saying that this approach is the best way forwards. All of this config needs to be reviewed and refactored according to the new camel recommendations and best practices.
I did remove a bunch of config that can spin up an embedded ActiveMQ (repository/src/main/resources/alfresco/subsystems/Messaging/default/activemq/*). I think this is not required and we should remove this functionality and dependencies for it completely.

repository/src/main/java/org/alfresco/messaging/camel/configuration/ConnectionFactoryConfiguration.java
```diff
@@ -1,96 +0,0 @@
...
-import org.springframework.context.annotation.Configuration;
-
-/**
- * Configures the ActiveMQ connection factory for use.
- *
- * @author Gethin James
- */
-
-@Configuration
-public class ConnectionFactoryConfiguration
```
jamalkm, line 49: This config class is necessary for other components, like event2 or even sync service. So why we removing this?

jamalkm (changes requested):

repository/src/main/java/org/alfresco/messaging/camel/configuration/ConnectionFactoryConfiguration.java
killerboot: @jamalkm please see my comment: https://github.com/Alfresco/alfresco-community-repo/pull/268#issuecomment-771534030

jamalkm: @killerboot sorry, I don't get the comment!
That config class is only creating a connection using `ActiveMQConnectionFactory` for best performance rather the generic JMS connection factory or others. It point to localhost because this is how the property is set:
`messaging.broker.url=failover:(tcp://localhost:61616)?timeout=3000&jms.useCompression=true`
So in production we override this. For example, when using Amazon ActiveMQ.

This connection factory then plugged in into an ActiveMQ `PooledConnectionFactory`, then we use the `PooledConnectionFactory` to create Camel component like `amqp`. So repository `event2` is using the `amqp` component to configure Camel route.

killerboot: I don't disagree with any of that, but this is how it worked with previous camel. What happens now is that there is a separate component class ActiveMQComponent that controls the pool and connection factory. When dsync starts to use the events, those are going to localhost regardless of what is set into messaging.broker.url as the ActiveMQComponent is not using any of ConnectionFactoryConfiguration by default. The same applies to messaging.broker.connections.max=8, The thread dump shows 360+ threads for activemq, so the pool is not configured correctly.

You can observe the behaviour if you build the latest repository and then try to run SyncServiceAPITests in acs-packaging.
My changes are just the illustration of that. I'm not claiming that this class should be removed.

As a side note there is a complete copy of the configuration that creates another connection factory for desktop sync. Do we need 2 of those that are using the same ActiveMQ? https://github.com/Alfresco/dsync-services/blob/master/alfresco-device-sync-repo/src/main/java/org/alfresco/messaging/camel/configuration/DeviceSyncConnectionFactoryConfiguration.java

jamalkm: I see, thanks for the clarification.
I think we need to simplify the ActiveMQ connection and use that connection throughout the Repository and any AMP that requires it.  
To answer your question re: 
> Do we need 2 of those that are using the same ActiveMQ? 

I think it's not required in the Desktop Sync AMP as long as they use the one from the repository.

ciprianchelaru: Also, I'll do a test without that class and I'll let you know the result.

ciprianchelaru: The redundant JMS configuration can be removed safely, this is a green build test https://travis-ci.com/github/Alfresco/dsync-services/builds/216040343

CezarLeahu:
I've created PR https://github.com/Alfresco/alfresco-community-repo/pull/274 with the updated messaging configuration. We should continue the conversation there @killerboot, @jamalkm, @alandavis.

We can probably close this Draft PR.

PR #274, opened 2021-02-05
ACS-1215 Update JMS/ActiveMQ configuration after the Camel 3.7 upgrade

CezarLeahu:
- add an **activemq** messaging component bean (of type `org.apache.camel.component.activemq.ActiveMQComponent`)
- add a common **jmsConfig** bean to be used by all the three messaging component beans: **jms**, **activemq** and **amqp**
- exclude `activemq-broker` dependencies
- remove unnecessary messaging configuration


CezarLeahu:
I've took the changes from https://github.com/Alfresco/alfresco-community-repo/pull/268 and modified them to meet our requirements.
The best solution I've found so far was to use one common **JmsConfiguration** bean in all our messaging components. This way the connection/thread pools should be shared, regardless of the used protocol.
I've ended up keeping the [ConnectionFactoryConfiguration](https://github.com/Alfresco/alfresco-community-repo/blob/master/repository/src/main/java/org/alfresco/messaging/camel/configuration/ConnectionFactoryConfiguration.java) class, as that's needed for the SSL configuration (no better alternatives).

And yes, we do have three messaging components now: **amqp**, **jms** and **activemq**. Before the _Camel_ upgrade we only had **amqp**. It's still unclear to me why we now need all three - particularly why we need both **jms** and **activemq**. Anyway, with the proposed configuration all the test suites in the ACS builds pass - green branch builds:
* https://travis-ci.com/github/Alfresco/alfresco-community-repo/builds/216190605
* https://travis-ci.com/github/Alfresco/alfresco-enterprise-repo/builds/216192850
* https://travis-ci.com/github/Alfresco/acs-packaging/builds/216189362
The DSync client also seems to be working fine with the new configuration.

I still have to inspect the ACS JVM and the connection pools at runtime and make sure the large number of ActiveMQ threads mentioned in https://github.com/Alfresco/alfresco-community-repo/pull/268 by @killerboot is no longer an issue, but for now, this configuration seems to solve our DSync compatibility issue.

killerboot (approved):
LGTM

killerboot:
Does it make sense to add any additional tests for community-repo to ensure that the messaging setups works fine?
Perhaps to help catch issues before they reach dsync or anything that is using events.

jamalkm (approved):
LGTM. Good work for encapsulating the config with JmsConfiguration!

CezarLeahu:
> Does it make sense to add any additional tests for community-repo to ensure that the messaging setups works fine?
> Perhaps to help catch issues before they reach dsync or anything that is using events.

Yes, I think we could probably add a few short tests to ensure that camel communication through the `jms:`, `activemq:` and `amqp:` protocols/URIs is working.
It might take me bit to write them (I'll probably open a different PR for that), but I've added another AC for this on the JIRA ticket (https://alfresco.atlassian.net/browse/ACS-1215).

PR #311, opened 2021-02-24
REPO-5571

tzclucian:
## Fixed spammy logs

@rgherghelas found these logs while running the `TempFileCleaner`.
The cleaner had the same behaviour up until now. Now, we don't know whether or not the initial requirements were to display how many files were deleted per folder but if we don't like those logs this PR is ready to address the issue, by only logging when the poller actually deleted something in a folder.

image

### DIFF
--- a/core/src/main/java/org/alfresco/util/TempFileProvider.java
+++ b/core/src/main/java/org/alfresco/util/TempFileProvider.java
@@ -493,7 +493,7 @@ public class TempFileProvider
                     int countRemoved = removeFiles(file,
                         isLongLifeTempDir(file) ? longLifeBefore : removeBefore, longLifeBefore,
                         true);
-                    logger.debug("Removed " + countRemoved + " files from temp directory: " + file);
+                    logger.debug("Removed " + countRemoved + " files from " + (isLongLifeTempDir(file) ? "temp " : "") + "directory: " + file);
                 }
                 else
                 {
--- a/core/src/main/java/org/alfresco/util/TempFileProvider.java
+++ b/core/src/main/java/org/alfresco/util/TempFileProvider.java
@@ -481,18 +481,19 @@ public class TempFileProvider
             int count = 0;
             for (File file : filesToIterate)
             {
+                if (shouldTheDeletionStop())
+                {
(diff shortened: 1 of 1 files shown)


ancutam (approved):

core/src/main/java/org/alfresco/util/TempFileProvider.java
```diff
@@ -518,32 +505,24 @@ private static int removeFiles(File directory, long removeBefore, long longLifeB
                     try
                     {
                         // only delete if the limits allow
-                        if (maxFilesToDelete != null && maxFilesToDelete.get() <= 0 ||
-                            maxTimeToRun != null && ((jobStartTime + maxTimeToRun.toMillis()) < System.currentTimeMillis()))
+                        if (shouldTheDeletionContinue())
```
DenisGabriela, line 508: The name of the method is a bit confusing to me. 

>  if shouldTheDeletionContinue() is true then return count

I suggest renaming it to shouldTheDeleteStop() or something similar.

tzclucian: That's correct! Thanks.

core/src/main/java/org/alfresco/util/TempFileProvider.java
```diff
@@ -485,26 +481,19 @@ private static int removeFiles(File directory, long removeBefore, long longLifeB
...
-                            logger.debug("Removed " + countRemoved + " files from directory: " + file);
-                        }
-                    }
+                    // long life for this folder and its children
+                    // OR
+                    // enter subdirectory and clean it out and remove itsynetics
+                    int countRemoved = removeFiles(file,
+                        isLongLifeTempDir(file) ? longLifeBefore : removeBefore, longLifeBefore,
+                        true);
+                    logger.debug("Removed " + countRemoved + " files from temp directory: " + file);
```
cristianturlica, line 496: this changes removes the ability to know in the logs if we have along lived file or not...

tzclucian: Addressed.

PR #316, opened 2021-02-25
Update the ACS build scripts

CezarLeahu:
- mirror the `build_functions.sh` script from the other ACS repos
- no practical functional change


alandavis (approved):
LGTM

PR #317, opened 2021-02-26
[MNT-20006] Obtaining the site as system in order to allow the activity record when an user is not member of a site

tiagosalvado10:


jottley (approved):

repository/src/main/java/org/alfresco/repo/site/SiteServiceImpl.java
```diff
@@ -1419,7 +1419,7 @@ private NodeRef getSiteNodeRef(NodeRef nodeRef)
         }
         else
         {
-            ChildAssociationRef primaryParent = nodeService.getPrimaryParent(nodeRef);
+            ChildAssociationRef primaryParent = directNodeService.getPrimaryParent(nodeRef);
```
skopf, line 1422: Changing the behaviour of this method is very dangerous. It is currently throwing an Exception if the calling user is not a member of this Site. Other callers might rely on this behaviour. For example, it is currently used in the deleteSite() and stopping execution immediately at the beginning.

skopf (changes requested):
We cannot change the (implicit) contract of the getSiteNodeRef method.

repository/src/main/java/org/alfresco/repo/site/SiteServiceImpl.java
skopf: This change might be OK, but we need to check if this change of behaviour in this method has unexpected side effects.

tiagosalvado10: **Analysis:**

The changed method **getSiteNodeRef(NodeRef nodeRef)** is used by:

1) getSite(NodeRef)
2) getSiteShortName(NodeRef)

The **deleteSite(String)** and **updateSite(String)** methods don't use the changed method but the one below:

3) getSiteNodeRef(String)

Using the directNodeService bean here will affect all the calls to methods 1) and 2) that supply, e.g., a docLib nodeRef (or any other site sub-container) instead of the actual site nodeRef.

As there are too many calls to the methods 1) and 2), it would be danger to change it indeed.

Would it be an option to add a new method to SiteServiceImpl class for this purpose (not exposed in the Public Api) and then change the siteService type in ActivityPosterImpl class to SiteServiceImpl ?

skopf:
ActivityPosterImpl should only rely to the SiteService interface.

Another option is to use AuthenticationUtil.runAsSystem( { ... } ). This is easy if you just need to it read info. Be careful when using it to modify values, as it will influence the auditing properties.

tiagosalvado10:
Reverted the initial changes so the proxy beans are used again. Changed the ActivityPosterImpl class to obtain the site as system and allow the activity record when metadata is updated.

skopf (approved):
lgtm

PR #323, opened 2021-03-02
MNT-22184 Add security header for admin console

nithinnambiar:
Added X-Frame-Options security header for admin console.
https://alfresco.atlassian.net/browse/MNT-22184


eknizat (approved):
LGTM

PR #326, opened 2021-03-03
Externalize the nodeServiceCleanup CRON expression

nbarithel:
This would allow to update the cron expression, especially to put a year 2099 on read-only/Solr tracking repositories


skopf (approved):
LGTM. Externalizing cron expressions makes a lot of sense.

tzclucian:
@nbarithel Thank you for your contribution!

PR #334, opened 2021-03-04
ACS-1217 Additional option for SOLR to authenticate with a shared secret

skopf:


killerboot (approved):
@skopf I added config to the repository.properties and a small test. I think it is good to go, unless you would like to also add a REST API level test.

skopf:
Thanks for your contribution @killerboot . I will merge it once the build is green.

PR #356, opened 2021-03-23
ACS-1252 : Add tests for the JMS/ActiveMQ/Camel configuration in *alfresco-community-repo*

DenisGabriela:
- fix messaging context for existing tests by adding missing mocked beans and properties 
   - add tests for amqp, jms, activemq protocols
   - add travis job to run CamelRoutesTest


CezarLeahu:
I went through the PR and... I have some thoughts:

The primary purpose of ACS-1252 is to check that the **Camel _JMS/AMQP/Activemq_ components** exist in the Alfresco context and are usable. I'm sure the current test setup does check that, but it does so in a rather indirect way, in that it relies a lot on routes and additional beans (i.e., _RequestProcessor_).

I think that starting from the current test setup in this PR we could simplify the new tests a bit, so their purpose is more clear:
* delete the `RequestProcessor.java` class/bean + all its usages (`CamelRoutesTest`, `test-messaging-context.xml` and `testRoutes.xml` - leave the `direct-vm` routes though, we still need those)
* rename the tests from `test<Protocol>Route()` to `test<Protocol>Component()` to better convey their purpose - to that end the new tests could also be moved into a separate class in the same package with `CamelRoutesTest` (but that can be done later)
* modify the content of each test method to something along this pattern:
```
final String msg = "ping <protocol>";

<protocol>Template.sendBody(msg);

final Object reply = camelContext
    .createConsumerTemplate()
    .receiveBody("<protocol>:queue:alfresco.test", 2000);

assertEquals(msg, reply);
```
where `<protocol>` is supposed to be `activemq`, `amqp` or `jms`.

Now, this solution doesn't modify the message body between request and reply, but I think we can live without that.
This solution tests both read and write through the Camel components. We should run these tests once on a local workstation, with a fresh container of ActiveMQ and check the number of messages on the `alfresco.test` queue (in the ActiveMQ Web Console). If we get a difference of 3 (one for each protocol), I think we're OK.

Also, this solution could be expanded with a plain ActiveMQ Java client which would read from the `alfresco.test` queue, and reply with a modified message in a `alfresco.test.reply` queue, and this latter queue would be the one used by the tests to read the replies. I don't think it's worth the effort, though...

CezarLeahu (approved):
LGTM

.travis.yml
```diff
@@ -210,6 +210,11 @@ jobs:
         - docker run -d -p 61616:61616 -p 5672:5672 alfresco/alfresco-activemq:5.16.1
       script: travis_wait 20 mvn -B test -pl remote-api -Dtest=AppContextExtraTestSuite -Ddb.driver=org.postgresql.Driver -Ddb.name=alfresco -Ddb.url=jdbc:postgresql://localhost:5433/alfresco -Ddb.username=alfresco -Ddb.password=alfresco
 
+    - name: "Messaging tests"
```
ancutam, line 213: Maybe add Repository to the name of the job, to keep the same pattern, `Repository - Messaging Tests`.

ancutam (approved):

PR #357, opened 2021-03-24
SEARCH-2768 Add callback feature for asynchronous ACL updates.

tpage-alfresco:
Change default for user filter to empty, as changes from all users could affect metadata
or permissions which might need to be indexing by the enterprise Elasticsearch Connector.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterListener.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterListener.java
@@ -31,5 +31,5 @@ import org.alfresco.service.cmr.repository.NodeRef;
 public interface FixedAclUpdaterListener
 {
     /** Callback method for when permissions have been updated by the FixedAclUpdater. */
-    void permissionsUpdated(NodeRef nodeRef);
+    void permissionsUpdatedAsynchronously(NodeRef nodeRef);
 }
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterListener.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterListener.java
@@ -1,0 +1,35 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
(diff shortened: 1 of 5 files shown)


repository/src/main/resources/alfresco/repository.properties
```diff
@@ -1202,7 +1202,7 @@ repo.event2.filter.nodeAspects=sys:*
 repo.event2.filter.childAssocTypes=rn:rendition
 # Comma separated list of users which should be excluded
 # Note: username's case-sensitivity depends on the {user.name.caseSensitive} setting
-repo.event2.filter.users=System, null
+repo.event2.filter.users=
```
tpage-alfresco, line 1205: I didn't work out why the user sending making the changes shows up as "null" - I could work around this by using a "run as system", but I figured it's safer to avoid filtering any events by default so that Elasticsearch has the potential to receive any updates to metadata/permissions.

evasques (approved):
LGTM

jesty (commented):
LGTM

repository/src/main/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterListener.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.alfresco.repo.domain.permissions;
+
+import org.alfresco.service.cmr.repository.NodeRef;
+
+/** Listener to receive callback events when permissions are asynchronously updated with the {@link FixedAclUpdater}. */
+public interface FixedAclUpdaterListener
+{
+    /** Callback method for when permissions have been updated by the FixedAclUpdater. */
+    void permissionsUpdated(NodeRef nodeRef);
```
jesty, line 34: Could be better to highlight that this method is invoked only when the update is asynchronous? We can call it something like permissionsUpdatedAsynchronously. The comment is pretty clear but when I read the implementation on the other PR isn't so clear to me at first glance.

jesty (approved):

repository/src/test/java/org/alfresco/repo/domain/permissions/FixedAclUpdaterUnitTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.util.Pair;
+import org.alfresco.util.PolicyIgnoreUtil;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+
+/** Mock-based unit tests for {@link FixedAclUpdater}. */
+public class FixedAclUpdaterUnitTest
```
tpage-alfresco, line 49: These tests won't run unless they're manually added to a test suite.

PR #360, opened 2021-03-25
Bugfix/repo 5610 events are not actually sent to activemq

n-insaidoo:
A fix is applied to collect the user name in the calling thread in `EventGenerator::sendEvent` so that it doesn't silently fail.
As part of detecting this issue, changes were made to verify the reception of events on the `alfresco.repo.event2` event topic through end-to-end tests. Please see enterprise companion PR [here](https://github.com/Alfresco/alfresco-enterprise-repo/pull/344).

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
+++ b/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
@@ -44,13 +44,13 @@ import org.springframework.beans.factory.InitializingBean;
  */
 public class EventGeneratorQueue implements InitializingBean
 {
-    private static final Log LOGGER = LogFactory.getLog(EventGeneratorQueue.class);
+	protected static final Log LOGGER = LogFactory.getLog(EventGeneratorQueue.class);
     
-    private Executor enqueueThreadPoolExecutor;
-    private Executor dequeueThreadPoolExecutor;
-    private Event2MessageProducer event2MessageProducer;
-    private BlockingQueue<EventInMaking> queue = new LinkedBlockingQueue<>();
-    private Runnable listener = createListener();
+    protected Executor enqueueThreadPoolExecutor;
+    protected Executor dequeueThreadPoolExecutor;
+    protected Event2MessageProducer event2MessageProducer;
+    protected BlockingQueue<EventInMaking> queue = new LinkedBlockingQueue<>();
+    protected Runnable listener = createListener();
 
     @Override
     public void afterPropertiesSet() throws Exception
--- a/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
+++ b/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
@@ -37,14 +37,12 @@ import org.alfresco.util.PropertyCheck;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.context.ApplicationEvent;
-import org.springframework.extensions.surf.util.AbstractLifecycleBean;
 
 /*
  * This queue allows to create asynchronously the RepoEvent offloading the work to a ThreadPool but
  * at the same time it preserves the order of the events
  */
-public class EventGeneratorQueue extends AbstractLifecycleBean implements InitializingBean
+public class EventGeneratorQueue implements InitializingBean
 {
     private static final Log LOGGER = LogFactory.getLog(EventGeneratorQueue.class);
     
@@ -77,16 +75,6 @@ public class EventGeneratorQueue extends AbstractLifecycleBean implements Initia
         this.dequeueThreadPoolExecutor = dequeueThreadPoolExecutor;
         dequeueThreadPoolExecutor.execute(listener);
     }
-    @Override
-    protected void onBootstrap(ApplicationEvent event)
-    {
-    }
-    
-    @Override
-    protected void onShutdown(ApplicationEvent event)
-    {
-    }
-
 
     /**
      * Procedure to enqueue the callback functions that creates an event.
--- a/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
+++ b/repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
@@ -1,0 +1,191 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2021 Alfresco Software Limited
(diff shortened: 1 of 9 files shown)


repository/src/main/resources/alfresco/repository.properties
```diff
@@ -1209,6 +1209,10 @@ repo.event2.filter.childAssocTypes=rn:rendition
 repo.event2.filter.users=System, null
 # Topic name
 repo.event2.topic.endpoint=amqp:topic:alfresco.repo.event2
+# Thread pool for async delivery
+repo.event2.threadPool.priority=1
+repo.event2.threadPool.coreSize=8
+repo.event2.threadPool.maximumSize=10
```
bbossola, line 1216: Check if you can get rid of this ^M characters

bbossola (changes requested):

repository/src/main/resources/alfresco/repository.properties
n-insaidoo: The file is filled with that line termination character, I'd leave it as is otherwise I will have to remove it entirely.

repository/src/test/java/org/alfresco/repo/event2/EventGeneratorTest.java
```diff
@@ -0,0 +1,221 @@
...
+            ActiveMQTextMessage am = (ActiveMQTextMessage) message;
+            return am.getText();
+        }
+        catch (JMSException e)
+        {
+            return null;
+        }
+    }
+
+    // a simple main to investigate he contents of the local broker
```
killerboot, line 152: a small typo "...to investigate **t**he contents..."

repository/src/test/java/org/alfresco/repo/event2/EventGeneratorTest.java
```diff
@@ -0,0 +1,221 @@
...
+            {
+                @Override
+                public void onMessage(Message message)
+                {
+                    String text = getText(message);
+                    System.out.println("Received message " + message + "\n" + text+"\n");
+                }
+            });
+            
+            Thread.sleep(50000000l);
```
killerboot, line 214: Can you please make this sleep configurable and passed through the "main" as a parameter?

killerboot (approved):
Overall looks ok to me. I just noted some minor changes.

Is there any concern about message order that may have changed because of the async behaviour plus a executor thread pool?

repository/src/test/java/org/alfresco/repo/event2/EventGeneratorTest.java
n-insaidoo: Sorted out.

repository/src/test/java/org/alfresco/repo/event2/EventGeneratorTest.java
n-insaidoo: Sure, I will do that.

robfrank (approved):

n-insaidoo:
> Overall looks ok to me. I just noted some minor changes.
> 
> Is there any concern about message order that may have changed because of the async behaviour plus a executor thread pool?

@killerboot It's a fair point. A thread pool won't necessarily guarantee *exact* execution order (although it will be close). Being the messaging bus asynchronous (and the fact we are using a pub/sub model) I do not necessarily see this as an issue. However, all depends on the contract we already published. @alandavis can you fill in here please?

n-insaidoo:
Changes have been applied with the latest commit to guarantee the message order despite the newly introduced async behaviour.

repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
```diff
@@ -0,0 +1,191 @@
...
+    private Runnable listener = createListener();
+
+    @Override
+    public void afterPropertiesSet() throws Exception
+    {
+        PropertyCheck.mandatory(this, "enqueueThreadPoolExecutor", enqueueThreadPoolExecutor);
+        PropertyCheck.mandatory(this, "dequeueThreadPoolExecutor", dequeueThreadPoolExecutor);
+        PropertyCheck.mandatory(this, "event2MessageProducer", event2MessageProducer);
+    }
+
```
robfrank, line 64: Is it possible to write a  constructor instead of getters/setters? 
Spring supports constructor with named params on XML too

repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
```diff
@@ -0,0 +1,191 @@
...
+    public void setDequeueThreadPoolExecutor(Executor dequeueThreadPoolExecutor)
+    {
+        this.dequeueThreadPoolExecutor = dequeueThreadPoolExecutor;
+        dequeueThreadPoolExecutor.execute(listener);
+    }
+    @Override
+    protected void onBootstrap(ApplicationEvent event)
+    {
+    }
+
```
robfrank, line 84: could this empty method be avoided?

repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
```diff
@@ -0,0 +1,191 @@
...
+    @Override
+    protected void onBootstrap(ApplicationEvent event)
+    {
+    }
+    
+    @Override
+    protected void onShutdown(ApplicationEvent event)
+    {
+    }
+
```
robfrank, line 89: same as previous method

robfrank (approved):
Code looks good to me, using constructor instead of getters/setters is a wish

repository/src/main/java/org/alfresco/repo/event2/EventGeneratorQueue.java
n-insaidoo: Yes those empty methods can be avoided, I will remove that and the other mentioned in the above comment.

bbossola (approved):

robfrank (approved):

repository/src/main/resources/alfresco/events2-context.xml
```diff
@@ -38,7 +38,6 @@
         <property name="dictionaryService" ref="dictionaryService"/>
         <property name="descriptorService" ref="descriptorComponent"/>
         <property name="eventFilterRegistry" ref="event2FilterRegistry"/>
-        <property name="event2MessageProducer" ref="event2MessageProducer"/>
```
jamalkm, line 41: See my comment in the enterprise PR. Please add `eventGeneratorQueue` here to avoid the need to set it in the enterprise.

jamalkm (approved):
LGTM

bbossola (approved):

PR #371, opened 2021-04-06
SEARCH-2782 use current transaction timestamp instead of current time as event time

jesty:


eliaporciani (commented):
I think we should take the transaction timestamp of the transaction in which the node has been updated/create/deleted. It is not guaranteed that the last transaction is the right one.

jesty:
> I think we should take the transaction timestamp of the transaction in which the node has been updated/create/deleted. It is not guaranteed that the last transaction is the right one.

I'm using the **current transaction**, not the **last transaction**. I think it is the same used to update/create/delete the node. Anyway, I'll do a check on it.

eliaporciani (approved):

PR #377, opened 2021-04-08
SEARCH-2782 commit time as event time

jesty:
This PR is updated because we face an issue when the transaction entry object is not initialized and we cannot retrieve the commit time. This happens when no nodes were added to the transaction because there aren't changes to any node, but we need to produce some events (e.g. permission update).

Before I’m getting the commit time using the nodeDao:
```
Long currentTransactionId = nodeDAO.getCurrentTransactionId(true);
Transaction transaction = nodeDAO.getTxnById(currentTransactionId);
Instant commitTimeMs = Instant.ofEpochMilli(transaction.getCommitTimeMs());
```
The PermissionService does not invoke always NodeDAOImpl.getCurrentTransaction() method, so even if we are inside a transaction the Transaction entry will not be initialized and the method nodeDAO.getCurrentTransactionId() will return null or it will throw an exception, depends on the boolean parameter value. In this case, also the ACL transaction is null. For instance, if you remove an ACL from a node and that ACL type is shared the transaction entry is never initialized.

In this PR I'm purposing a solution checking if the current transaction id is null, and then using the current time.

I prefer the solution above because it avoids exceptions for a trivial reason. Otherwise, we can force the transaction entry initialization invoking nodeDAO.getCurrentTransactionId(true) in PermissionEventGenerator but in this case, we are creating empty transactions and we are relying on a side effect (the getter that inits an object).

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
@@ -563,6 +563,7 @@ public abstract class AbstractNodeDAOImpl implements NodeDAO, BatchingDAO
             Long txnId = txn.getId();
             // Update it
             Long now = System.currentTimeMillis();
+            txn.setCommitTimeMs(now);
             updateTransaction(txnId, now);
         }
     }
@@ -602,6 +603,17 @@ public abstract class AbstractNodeDAOImpl implements NodeDAO, BatchingDAO
         AlfrescoTransactionSupport.bindDaoService(updateTransactionListener);
         // Done
         return txn;
+    }
+    
+    public Long getCurrentTransactionCommitTime()
+    {
+        Long commitTime = null;
+        TransactionEntity resource = AlfrescoTransactionSupport.getResource(KEY_TRANSACTION);
+        if(resource != null)
+        {
+            commitTime = resource.getCommitTimeMs();
+        }
+        return commitTime;
     }
     
     public Long getCurrentTransactionId(boolean ensureNew)
--- a/repository/src/main/java/org/alfresco/repo/domain/node/NodeDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/node/NodeDAO.java
@@ -75,6 +75,13 @@ public interface NodeDAO extends NodeBulkLoader
     /*
      * Transaction
      */
+
+    /**
+     * @return                 the commit time of the current transaction entry or <tt>null</tt> if
+     *                         there have not been any modifications to nodes registered in the
+     *                         transaction.
+     */
+    Long getCurrentTransactionCommitTime();
     
     /**
      * @param ensureNew          <tt>true</tt> to ensure that a new transaction entry is created
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
@@ -4,21 +4,21 @@
  * %%
  * Copyright (C) 2005 - 2016 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software.
- * If the software was purchased under a paid Alfresco license, the terms of
- * the paid license agreement will prevail.  Otherwise, the software is
+ * This file is part of the Alfresco software. 
+ * If the software was purchased under a paid Alfresco license, the terms of 
+ * the paid license agreement will prevail.  Otherwise, the software is 
  * provided under the following open source license terms:
- *
+ * 
  * Alfresco is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * 
  * Alfresco is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- *
+ * 
  * You should have received a copy of the GNU Lesser General Public License
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
@@ -4,21 +4,21 @@
  * %%
  * Copyright (C) 2005 - 2016 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software. 
- * If the software was purchased under a paid Alfresco license, the terms of 
- * the paid license agreement will prevail.  Otherwise, the software is 
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
  * provided under the following open source license terms:
- * 
+ *
  * Alfresco is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * Alfresco is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU Lesser General Public License
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
@@ -186,7 +186,7 @@ public interface AclDAO
     /**
     * @return                 the commit time of the current ACL change set entry or <tt>null</tt> if
     *                         there have not been any modifications.
-    */ 
-    Long getCurrentChangeSetCommitTime();
+    */
+    public Long getCurrentChangeSetCommitTime();
 
 }
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
@@ -182,4 +182,9 @@ public interface AclDAO
      * @return Long
      */
     public Long getMaxChangeSetIdByCommitTime(long maxCommitTime);
+
+    /**
+     * @return the current AclChangeSet created for the current ACL change 
+     */
+    AclChangeSet getCurrentACLChangeSet();
 }
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
@@ -184,7 +184,9 @@ public interface AclDAO
     public Long getMaxChangeSetIdByCommitTime(long maxCommitTime);
 
     /**
-     * @return the current AclChangeSet created for the current ACL change 
-     */
-    AclChangeSet getCurrentACLChangeSet();
+    * @return                 the commit time of the current ACL change set entry or <tt>null</tt> if
+    *                         there have not been any modifications.
+    */ 
+    Long getCurrentChangeSetCommitTime();
+
 }
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
@@ -4,21 +4,21 @@
  * %%
  * Copyright (C) 2005 - 2016 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software.
- * If the software was purchased under a paid Alfresco license, the terms of
- * the paid license agreement will prevail.  Otherwise, the software is
+ * This file is part of the Alfresco software. 
+ * If the software was purchased under a paid Alfresco license, the terms of 
+ * the paid license agreement will prevail.  Otherwise, the software is 
  * provided under the following open source license terms:
- *
+ * 
  * Alfresco is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * 
  * Alfresco is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- *
+ * 
  * You should have received a copy of the GNU Lesser General Public License
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
--- a/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
@@ -4,21 +4,21 @@
  * %%
  * Copyright (C) 2005 - 2016 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software. 
- * If the software was purchased under a paid Alfresco license, the terms of 
- * the paid license agreement will prevail.  Otherwise, the software is 
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
  * provided under the following open source license terms:
- * 
+ *
  * Alfresco is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * Alfresco is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU Lesser General Public License
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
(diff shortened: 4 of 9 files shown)


repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
```diff
@@ -370,13 +381,35 @@ private boolean isFilteredChildAssociation(QName childAssocType, String user)
...
+        return new EventInfo().setTimestamp(getCurrentTransactionTimestamp())
                               .setId(UUID.randomUUID().toString())
                               .setTxnId(AlfrescoTransactionSupport.getTransactionId())
                               .setPrincipal(user)
                               .setSource(URI.create("/" + descriptorService.getCurrentRepositoryDescriptor().getId()));
     }
 
+    private ZonedDateTime getCurrentTransactionTimestamp()
+    {
+        Long currentTransactionId = nodeDAO.getCurrentTransactionId(false);
```
aborroy, line 393: This method works relying on the execution of this method inside the same Thread that is executing the transaction. If that changes at same point, an alternative method should be adopted.

aborroy (approved):
LGTM

jesty:
> This PR is updated because we face an issue when the transaction entry object is not initialized and we cannot retrieve the commit time. This happens when no nodes were added to the transaction because there aren't changes to any node, but we need to produce some events (e.g. permission update).
> 
> Before I’m getting the commit time using the nodeDao:
> 
> ```
> Long currentTransactionId = nodeDAO.getCurrentTransactionId(true);
> Transaction transaction = nodeDAO.getTxnById(currentTransactionId);
> Instant commitTimeMs = Instant.ofEpochMilli(transaction.getCommitTimeMs());
> ```
> 
> The PermissionService does not invoke always NodeDAOImpl.getCurrentTransaction() method, so even if we are inside a transaction the Transaction entry will not be initialized and the method nodeDAO.getCurrentTransactionId() will return null or it will throw an exception, depends on the boolean parameter value. In this case, also the ACL transaction is null. For instance, if you remove an ACL from a node and that ACL type is shared the transaction entry is never initialized.
> 
> In this PR I'm purposing a solution checking if the current transaction id is null, and then using the current time.
> 
> I prefer the solution above because it avoids exceptions for a trivial reason. Otherwise, we can force the transaction entry initialization invoking nodeDAO.getCurrentTransactionId(true) in PermissionEventGenerator but in this case, we are creating empty transactions and we are relying on a side effect (the getter that inits an object).

It’s not true that the PermissionService never creates a transaction entry. It creates a new transaction if it is required, I mean, only if there is an actual ACL change to the node. The test was failing because the transaction is null, but the transaction was null because the permission isn’t changed, it is the same before, so we can even avoid producing the event.
IMHO this is an important thing because we can try to make the system lighter avoiding useless events. For instance, two tests have been failed because of:
the test is removing all permissions from a node that hasn’t permission. So the ACL it’s the same.
the test is moving a folder to a new parent with the same ACL as the old parent. So the ACL it’s the same.
In both scenarios, the system is not starting a transaction and relying on that we can avoid producing useless events.

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
killerboot: Afaik the idea is to produce messages from the events (serialise the java object into text and etc) by a separate thread pool of workers. The events are still generated after the transaction commit stage.

killerboot (approved):

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
```diff
@@ -368,13 +379,36 @@ private boolean isFilteredChildAssociation(QName childAssocType, String user)
...
+        ZonedDateTime timestamp;
+        if(currentTransactionId != null)
+        {
+            Transaction transaction = nodeDAO.getTxnById(currentTransactionId);
+            Instant commitTimeMs = Instant.ofEpochMilli(transaction.getCommitTimeMs());
+            timestamp = ZonedDateTime.ofInstant(commitTimeMs, ZoneOffset.UTC);
+        }
+        else 
+        {
+            // Sometimes the transaction entry is not created because no nodes were added to the transaction, but we are producing events anywhere.
```
killerboot, line 401: If the transaction is not active and nothing was committed to the DB, then there should be no events produced.
I think you can check the transaction state by AlfrescoTransactionSupport.isActualTransactionActive().

killerboot (changes requested):
Can you please check why the events are produced even if the transaction is not committed?
I suspect there is a bug in the event generator or consolidator.

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
jesty: Totally agree, I already did some experiments about it. Should I do it in this PR or in a separate PR/Issue?

jesty: Anyway, isActualTransactionActive doesn't work in this case because the transaction it is active, but it is never used so AlfrescoTransactionSupport.getResource(KEY_TRANSACTION) is null.

jesty:
> Can you please check why the events are produced even if the transaction is not committed?
> I suspect there is a bug in the event generator or consolidator.

The transaction is committed, but the TransactionEntity is never created and stored in the database because there isn't a real change. For instance, if I remove all permission from a node that hasn't permission, or if I change permission to a node setting the same ACL it has before the change, the transaction is created and committed, but TransactionEntity will never create and stored in the database. This happens because TransactionEntity will be initialized only when the system adds something to a transaction and if there aren't real updates, the system doesn't add anything to the transaction.

killerboot:
> > Can you please check why the events are produced even if the transaction is not committed?
> > I suspect there is a bug in the event generator or consolidator.
> 
> The transaction is committed, but the TransactionEntity is never created and stored in the database because there isn't a real change. For instance, if I remove all permission from a node that hasn't permission, or if I change permission to a node setting the same ACL it has before the change, the transaction is created and committed, but TransactionEntity will never create and stored in the database. This happens because TransactionEntity will be initialized only when the system adds something to a transaction and if there aren't real updates, the system doesn't add anything to the transaction.

In this case the event consolidator should not produce any events as there was no change to nodes or permissions.

jesty:
> > > Can you please check why the events are produced even if the transaction is not committed?
> > > I suspect there is a bug in the event generator or consolidator.
> > 
> > 
> > The transaction is committed, but the TransactionEntity is never created and stored in the database because there isn't a real change. For instance, if I remove all permission from a node that hasn't permission, or if I change permission to a node setting the same ACL it has before the change, the transaction is created and committed, but TransactionEntity will never create and stored in the database. This happens because TransactionEntity will be initialized only when the system adds something to a transaction and if there aren't real updates, the system doesn't add anything to the transaction.
> 
> In this case the event consolidator should not produce any events as there was no change to nodes or permissions.

So, I identified 4 scenarios:

- Node Transaction
- ACL changeset (Enterprise only)
- Node Transaction and ACL Changeset (Enterprise only)
- No transaction, in my test this happens only in few corner cases, I think no so common, for instance when you remove permissions from a node that hasn't any permission.

I updated this PR by adding a method to retrieve the ACL changeset object, this method will be used in the enterprise version to retrieve the ACL Changeset commit time.

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
```diff
@@ -1,28 +1,28 @@
...
+ * 
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
```
aborroy, line 25: What is the change in these lines?

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
```diff
@@ -1,28 +1,28 @@
...
+ * 
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
```
aborroy, line 25: Again, it's confusing adding this kind of changes to a commit...

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
```diff
@@ -1662,9 +1663,29 @@ public void beforeCommit(boolean readOnly)
...
             // Update it
             long commitTimeMs = System.currentTimeMillis();
+            AlfrescoTransactionSupport.bindResource(RESOURCE_KEY_ACL_CHANGE_SET_COMMIT_TIME_MS, commitTimeMs);
             aclCrudDAO.updateAclChangeSet(changeSetId, commitTimeMs);
         }
     }
+
+    /**
+     * Retrieve the value from the current thread local, avoiding the access to the database.
+     * @return
```
aborroy, line 1673: What happens if this method returns null?

aborroy (approved):
Minor comments

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
```diff
@@ -182,4 +182,9 @@
      * @return Long
      */
     public Long getMaxChangeSetIdByCommitTime(long maxCommitTime);
+
+    /**
+     * @return the current AclChangeSet created for the current ACL change
```
killerboot, line 187: This javadoc is not clear. "Current" means for current transaction, if it is present. The return value can be null.

The AclChangeSet is never used in this interface, do you know why?

killerboot (changes requested):
The events should **not** be generated for empty transactions. Please fix that.

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
jesty: I can change to: "@return the current AclChangeSet created for the current transaction. It can be null if a transaction is not active."
AclChangeSet is used only inside the implementation. In the interface it uses only field of this object, for instance in getMaxChangeSetIdByCommitTime and getMaxChangeSetCommitTime.

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
jesty: sorry, it is a formatting change, I'll remove both

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
killerboot: AclChangeSet is a DAO layer entity and should not be exposed via the interface :)

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
jesty: sorry, it is a formatting change and IntelliJ doesn't show to me, I'll remove both

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
jesty: I'll change the Javadoc in the interface, if null it means that there isn't a transaction active.

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
jesty: I apologize because they are in the same package the layering wasn't clear to me. Anyway I did a refactoring in order to have a method like  getMaxChangeSetIdByCommitTime

repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
```diff
@@ -604,6 +605,17 @@ private TransactionEntity getCurrentTransaction()
         return txn;
     }
     
+    public Long getCurrentTransactionCommitTime()
```
jesty, line 608: In order to avoid useless access to the database, I decide to use ThreadLocal in order to retrieve the current commit time, as did for the id, instead of reading it from the database.

repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
```diff
@@ -563,6 +563,7 @@ public void beforeCommit(boolean readOnly)
             Long txnId = txn.getId();
             // Update it
             Long now = System.currentTimeMillis();
+            txn.setCommitTimeMs(now);
```
killerboot, line 566: I think this is done in updateTransaction method.

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
```diff
@@ -182,4 +182,11 @@
      * @return Long
      */
     public Long getMaxChangeSetIdByCommitTime(long maxCommitTime);
+
+    /**
+    * @return                 the commit time of the current ACL change set entry or <tt>null</tt> if
+    *                         there have not been any modifications.
+    */ 
+    Long getCurrentChangeSetCommitTime();
```
killerboot, line 190: The method should be "public", right?

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
```diff
@@ -1636,7 +1636,8 @@ public Acl createLayeredAcl(Long indirectedAcl)
         return changes;
     }
 
-    private static final String RESOURCE_KEY_ACL_CHANGE_SET_ID = "acl.change.set.id";
+    public static final String RESOURCE_KEY_ACL_CHANGE_SET_ID = "acl.change.set.id";
```
killerboot, line 1639: Why did you make them public?

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
```diff
@@ -392,36 +425,41 @@ protected void onShutdown(ApplicationEvent applicationEvent)
...
+                    // Node events
+                    for (Map.Entry<NodeRef, EventConsolidator> entry : consolidators.getNodes().entrySet())
+                    {
+                        EventConsolidator eventConsolidator = entry.getValue();
+                        sendEvent(entry.getKey(), eventConsolidator);
+                    }
+
+                    // Child assoc events
+                    for (Map.Entry<ChildAssociationRef, ChildAssociationEventConsolidator> entry : consolidators.getChildAssocs()
+                                                                                                           .entrySet())
```
killerboot, line 443: This formatting looks very odd, please move everything in one line.

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
```diff
@@ -392,36 +425,41 @@ protected void onShutdown(ApplicationEvent applicationEvent)
...
+                    {
+                        ChildAssociationEventConsolidator eventConsolidator = entry.getValue();
+                        sendEvent(entry.getKey(), eventConsolidator);
+                    }
+
+                    // Peer assoc events
+                    for (Map.Entry<AssociationRef, PeerAssociationEventConsolidator> entry : consolidators.getPeerAssocs()
+                                                                                                     .entrySet())
```
killerboot, line 451: This formatting looks very odd, please move everything in one line.

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
killerboot: Is it correct that after the changes in EventTransactionListener in this class this code will never be executed? If so, can this if-else block be removed?

killerboot (changes requested):

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
jesty: in interfaces, the default modifier is public, so as IntelliJ uses to say: "Modified ‘public’ is redundant for interface methods" :)

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
jesty: removed

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAOImpl.java
jesty: I used RESOURCE_KEY_ACL_CHANGE_SET_COMMIT_TIME_MS in a test in the enterprise repository. In order to don't have similar constants with different modifiers, I used the same. Anyway, especially because they are in two different repositories, I prefer to avoid the coupling in the test, so I replaced the public with private and I'll use the string in the test.

repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
jesty: not exactly, updateTransaction method creates a new object and uses it for the update, while I need to update the object in memory.
```
  @Override
    protected int updateTransaction(Long txnId, Long commitTimeMs)
    {
        TransactionEntity transaction = new TransactionEntity();
        transaction.setId(txnId);
        transaction.setCommitTimeMs(commitTimeMs);
        return template.update(UPDATE_TRANSACTION_COMMIT_TIME, transaction);
    }
```

repository/src/main/java/org/alfresco/repo/event2/EventGenerator.java
jesty: done

repository/src/main/java/org/alfresco/repo/domain/permissions/AclDAO.java
killerboot: Please make it consistent with the rest of the interface.

repository/src/main/java/org/alfresco/repo/domain/node/NodeDAO.java
```diff
@@ -75,6 +75,13 @@
     /*
      * Transaction
      */
+
+    /**
+     * @return                 the commit time of the current transaction entry or <tt>null</tt> if
+     *                         there have not been any modifications to nodes registered in the
+     *                         transaction.
+     */
+    Long getCurrentTransactionCommitTime();
```
killerboot, line 84: Please make the method visibility consistent with the rest of the interface.

killerboot (approved):

PR #382, opened 2021-04-13
Feature/search 2802 shared secret auth

aborroy:


core/src/main/java/org/alfresco/httpclient/HttpClientFactory.java
```diff
@@ -75,23 +69,25 @@
...
-            {
-                return NONE;
-            }
-            else
+            switch (type.toLowerCase())
             {
-                throw new IllegalArgumentException("Invalid communications type");
+                case "https": return HTTPS;
+                case "none": return NONE;
+                case "secret": return SECRET;
```
tpage-alfresco, line 88: This is a lot more concise :+1: You could even use `SecureCommsType.valueOf(type.toUpperCase())`, which also happens to throw an IllegalArgumentException.

tpage-alfresco (approved):
LGTM

PR #383, opened 2021-04-13
[MNT-22183] Use absolute paths for stylesheets

davidcanonieto:
https://alfresco.atlassian.net/browse/MNT-22183

Relative URLs can be dangerous since browser may not determine the correct directory. If the HTML uses path-relative CSS links, it may be susceptible to path- relative stylesheet import (PRSSI) vulnerabilities. This could allow an attacker to take advantage of CSS imports with relative URLs by overwriting their target file.

### DIFF
--- a/packaging/war/src/main/webapp/error.jsp
+++ b/packaging/war/src/main/webapp/error.jsp
@@ -57,8 +57,8 @@ ModuleDetails shareServicesModule = moduleService.getModule("alfresco-share-serv
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
    <title>Alfresco</title>
-   <link rel="stylesheet" type="text/css" href="./css/reset.css" />
-   <link rel="stylesheet" type="text/css" href="./css/alfresco.css" />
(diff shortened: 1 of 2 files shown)


packaging/war/src/main/webapp/error.jsp
```diff
@@ -57,8 +57,8 @@ ModuleDetails shareServicesModule = moduleService.getModule("alfresco-share-serv
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
    <title>Alfresco</title>
-   <link rel="stylesheet" type="text/css" href="./css/reset.css" />
-   <link rel="stylesheet" type="text/css" href="./css/alfresco.css" />
+   <link rel="stylesheet" type="text/css" href="packaging/war/src/main/webapp/css/reset.css" />
```
adinapitul, line 60: have you tested this and are you sure this is actually working?
this path doesn't mean anything for the exploded war on tomcat for example

skopf (changes requested):
wrong path. This is pointing to the non existing /css folder in the server root

eromano (approved):

PR #387, opened 2021-04-16
ACS-895 - User email should not be necessary for joining a site

skopf:
This behaviour has been inconsistent anyway: It failed and prevented the user from being added to the site if there was no email address. But it proceeded if the email could not be sent for other reasons, like an invalid server name or non existing account on that server.


SaraAspery (approved):
LGTM

PR #413, opened 2021-04-27
ATS-893 Add Perl Licenses

PunchedChimera:


KikoUK (approved):

montgolfiere (changes requested):
Should this go into alfresco-transform-core project instead of this one ?

PunchedChimera:
> Should this go into alfresco-transform-core project instead of this one ?

They can be added to transform core, however, the other licences used by alfresco-transform-core are all stored here too. 
This seemed to more appropriate place to store them.

PunchedChimera:
I can see that the transformer licences have been removed. I will close and add the licences to alfresco-transform-core.

PR #427, opened 2021-05-04
ATS-892 / ATS-909: Add IPTC model to core Repo

montgolfiere:
- add existing IPTC model (& message files) alongside other core models, eg. Exif & DublinCore

- see also ATS-909

- note: one available in GA releases of 7.0.x.y (& 6.2.x.y) we will remove the copies from Transform Core project


PunchedChimera (approved):

KikoUK (approved):

repository/src/main/resources/alfresco/core-services-context.xml
```diff
@@ -719,6 +719,9 @@
                 <!-- Smart Folders model -->
                 <value>alfresco/model/smartfolder-model.xml</value>
 
+                <!-- IPTC model -->
+                <value>alfresco/model/contentModel.xml</value>
```
skopf, line 723: Shouldn't the name be "iptcModel.xml" ?

montgolfiere: @skopf Great, catch. Doh !

PR #435, opened 2021-05-06
REPO-5624 Reimplement SchemaBootstrapTest

n-insaidoo:
Now automating the update of the test schema reference file so that it is not required to be done manually whenever the main one is updated.


SaraAspery (approved):
LGTM!

PR #455, opened 2021-05-13
[MNT-22183] Use absolute paths for stylesheets (#383)

dhrn:
Use absolute paths for stylesheets


dhrn:
addressed https://github.com/Alfresco/alfresco-community-repo/pull/458

PR #472, opened 2021-05-19
REPO-5631 Investigate different types of results for optional patches difference detection

n-insaidoo:
Changes made to support other `Result` implementations when detecting optional patches problems descriptions.


bbossola (approved):
LGTM but make sure the build passes.

PR #473, opened 2021-05-20
ACS-1600 : Error when running propTablesCleanupJob on an env with 100million records in alf_prop_value

DenisGabriela:
- commit changes to db anytime `deleteBatchSize` of deleted data is reached


cristianturlica (approved):
LGTM

PR #489, opened 2021-05-24
[MNT-22419] Allow to save null leading values in a multiValued property

tiagosalvado10:
Usecase:

1. Create a model content property (e.g., d:date or d:text) as multi-value property
2. Upload a document, e.g., through REST api and add the previous property with [null, null, "2021-05-24", null, null]
3. The value ["2021-05-24", null, null] is assigned to document property, excluding the null values at the beginning

This PR has a proposal which change the behavior described above, storing all the supplied values: [null, null, "2021-05-24", null, null]


skopf (changes requested):
This will break existing applications. Let's have a discussion in the Jira ticket.

tiagosalvado10:
PR won't be merged as this behavior exists since always and changing it could lead to breaking changes.

PR #498, opened 2021-05-25
ACS-1602 : Update & tests for AbstractRoutingContentStore

DenisGabriela:


repository/src/main/java/org/alfresco/repo/content/AbstractRoutingContentStore.java
```diff
@@ -449,4 +451,51 @@ public boolean isStorageClassesSupported(Set<String> storageClasses)
...
+
+    @Override
+    public Set<String> findStorageClasses(String contentUrl)
+    {
+        Set<String> storageClasses = new HashSet<>();
+        for (ContentStore store : getAllStores())
+        {
+            storageClasses.addAll(store.findStorageClasses(contentUrl));
+        }
+        return storageClasses;
```
cristianturlica, line 477: This should be similar to `public ContentReader getReader(String contentUrl) throws ContentIOException` e.g.

```java
        ContentStore store = selectReadStore(contentUrl);
        if (store != null)
        {
            if (logger.isDebugEnabled())
            {
                logger.debug("Getting storage classes from store: \n" +
                        "   Content URL: " + contentUrl + "\n" +
                        "   Store:       " + store);
            }
            return store.findStorageClasses(contentUrl);
        }
        else
        {
            if (logger.isDebugEnabled())
            {
                logger.debug("Storage classes not found for content URL: " + contentUrl);
            }
           // TODO: return empty set
            return ;
        }
```

repository/src/main/java/org/alfresco/repo/content/AbstractRoutingContentStore.java
```diff
@@ -449,4 +451,51 @@ public boolean isStorageClassesSupported(Set<String> storageClasses)
...
+    @Override
+    public void updateStorageClasses(String contentUrl, Set<String> storageClasses,
+        Map<String, Object> parameters)
+    {
+        List<ContentStore> stores = getAllStores();
+        for (ContentStore store : stores)
+        {
+            if (store.isStorageClassesSupported(storageClasses))
+            {
+                store.updateStorageClasses(contentUrl, storageClasses, parameters);
```
tzclucian, line 464: There might be cases when the `storageClasses` are supported but the content is not in this store and the update will fail.

As it is the code will rethrow the exception further on.
I'm just think out loud whether or not we should address those exceptions in a way or another.

tzclucian (approved):

ancutam (approved):

PR #499, opened 2021-05-25
REPO-5632 Investigate intermittent failing tests in AppContext05TestSuite

n-insaidoo:
As a result of the fixedACL job finish off right before a few assertions take place, some tests that call `FixedAclUpdaterTest::getFirstNodeWithAclPending(QName nodeType, NodeRef parentRef)` occasionally fail as a sought after node with the `ASPECT_PENDING_FIX_ACL` aspect and grandchildren nodes is not always found after setting new permissions on the test data (a set of subfolders and files). An increase of the test data solves this issue.


SaraAspery (approved):
LGTM!

PR #500, opened 2021-05-25
REPO-5234: Metadata extraction fails for Embedded files

amohammedalfresco:
- Bump Tika to `1.26`


montgolfiere:
@amohammedalfresco If/when bumping Tika, I suggest checking other related dependencies (eg. pdfbox & poi) so that we have a consistent set.

KikoUK (approved):

PR #511, opened 2021-05-31
APPS-986: [AGS] Export of records with size> 4 MB fails

ramunteanu:
- Changed deleteTempFileOnClose

### DIFF
--- a/remote-api/src/main/java/org/alfresco/repo/web/scripts/RepositoryContainer.java
+++ b/remote-api/src/main/java/org/alfresco/repo/web/scripts/RepositoryContainer.java
@@ -109,7 +109,8 @@ public class RepositoryContainer extends AbstractRuntimeContainer
     {
         File tempDirectory = TempFileProvider.getTempDir(tempDirectoryName);
         this.streamFactory = new TempOutputStreamFactory(tempDirectory, memoryThreshold, maxContentSize, encryptTempFiles, false);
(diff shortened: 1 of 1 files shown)


ramunteanu:
As indicated in the documentation of the TempOutputStream's constructor:
     * @param deleteTempFileOnClose
     *            true if temp files should be deleted on output stream close
     *            (useful if we need to cache the content for further reads). If
     *            this is false then we need to make sure we call
     *            {@link TempOutputStream}.destroy to clean up properly.
     
The destroy is called in the BufferedResponse#writeResponse.
Since there where some changes in this area, will wait for the proposed fix to be agreed or make another change before adding a test.

cristianturlica (approved):
Changes look good to me, as the temp file is removed in 
` public void writeResponse()
    {
        try
        {
            if (logger.isDebugEnabled() && outputStream != null)
            {
                logger.debug("Writing Transactional response: size=" + outputStream.getLength());
            }
            
            if (outputWriter != null)
            {
                outputWriter.flush();
                res.getWriter().write(outputWriter.toString());
            }
            else if (outputStream != null)
            {
                if (logger.isDebugEnabled())
                    logger.debug("Writing Transactional response: size=" + outputStream.getLength());

                try
                {
                    outputStream.flush();
                    FileCopyUtils.copy(outputStream.getInputStream(), res.getOutputStream());
                }
                finally
                {
                    outputStream.destroy();
                }
            }
        }
        catch (IOException e)
        {
            throw new AlfrescoRuntimeException("Failed to commit buffered response", e);
        }
    }`

CezarLeahu (approved):

PR #518, opened 2021-06-08
ATS-886: IPTC content model (with message files)

montgolfiere:
Already part of ACS 7.1.0-M1 (and higher).

Cherry-pick to 7.0.N to significantly simplify OX. 

Workaround: manual scripts to statically bootstrap from classpath &/or custom AMP.


skopf (approved):
TECHNICALLY, this change is good. I think I already reviewed the original PR against master.

But we need to decide if we want to include this change in a SP. This is not a decision for me, but for Steve and Indy.

stevenblair-al (approved):
Approved to merge into 7.1.0 release train but **not approved to** go into 7.0.1 SP

montgolfiere:
@stevenblair-al That's unfortunate. It's a missed opportunity to simplify deployment for our customers using ACS 7.0.x.

However, It's already part of 7.1.0-M1 (so nothing else to merge) ;-)

montgolfiere:
I will close without merge for now. We can re-open, if needed in future, eg. based on customer request for a future SP (&/or HF).

PR #534, opened 2021-06-16
RM-7171 Remove reference to http://swagger.io/v2/schema.json to avoid downloading schema.

tpage-alfresco:
This prevents the unit tests from failing when the webpage is unavailable. It has the side
effect of not picking up any updates to the schema.


SaraAspery (approved):
LGTM!

tpage-alfresco:
The branch build was successful, the PR build failure has nothing to do with this change.

PR #536, opened 2021-06-16
MNT-20500 - Admin console breaks with serialised objects

evasques:
In node browser:

- Added macro convertToJSON to recursively parse hashes and enumerables
- Added attempt/recover in macros to handle errors and not break the page
- Changed the output of serialized objects to JSON format

In both admin console and node browser:

- Adjusted consistency of the output between both pages when an error occurs
- Validate the depth of each hash. When we find a hash with over 1000 elements, we throw an error instead of displaying the object. Used the stop tag to effectively force an abort of the template processing preventing performance or security issues regarding very large objects.

Related PRs:
 - Original PR in repo that fixes the serialised objects: https://github.com/Alfresco/alfresco-community-repo/pull/291
 - Original PR in share (when share-services was still in there) where the depth problem was pointed out: https://github.com/Alfresco/share/pull/397


jottley (approved):

pieCit87 (approved):
LGTM

PR #551, opened 2021-06-22
LOC-312 FIXED - Updated UI in 16 languages including the AGS amps fil…

gbroadbent:
…es in 5 new languages. Also a fix to l10n.properties to exclude the dictionary-model.properties from localisation.


alandavis (approved):
Review as limited to the l10n.properties  file. All other files are just standard L10n updates

PR #552, opened 2021-06-22
[ACS-1681] Remove reset password APIs

Giovanni007:
As discussed in MNT-22018, this is for removing the reset-password and request-password-reset API.


jamalkm (changes requested):
I know the Share part of this feature was not implemented, but the API has been published and we might have customers/partners using this API. Removing this public API would cause a lot of issues for those customers implemented a solution based on this API.

The lifecycle of V1 REST API is not dependant on Share or ADW. So, I think we shouldn't remove this API. If there is really a need to remove it, then first we have to deprecate it. 
@montgolfiere what do you think?

skopf:
@jamalkm How would a customer use this feature? When you call this API endpoint, then the system is sending out an email with a link to a non existing Share page. And that's it.

jamalkm:
@skopf customers need to register their own clients. The API does a lookup to get the registered client and generates the email based on the registered client and its email templates. The Share client mentioned in the Docs is for demo purposes only.

For example, suppose you want to register a client, named: `MyClient` and have your own email templates, in that case you need to set the following properties:
```
# Assets (images/logo) for the email template
repo.client-app.MyClient.templateAssetsUrl=${MyClientUrl}/your-images-path
# reset password request email template path. Can be:
#   an XPATH (e.g. app:company_home/app:dictionary/app:email_templates/example-email.ftl)
#   a NodeRef of the template
#   a class path of the template
repo.client-app.MyClient.requestResetPasswordTemplatePath=
# reset password UI page url
repo.client-app.MyClient.resetPasswordPageUrl=${MyClientUrl}/page/reset-password
# reset password confirmation email template path
repo.client-app.MyClient.confirmResetPasswordTemplatePath=
```
 
See `ClientAppConfig.java` JavaDocs for more details how it makes sense of these properties. Also, `EmailHelper.java` for how it retrieves the template.

Giovanni007:
I don't have a strong opinion about this, but If some customers are using them, I'm sure could be a problem to delete the API without deprecating them before. Actually, our doubts were not only about the missing implementation. Here some of the points we discussed and we were not sure about:

- **Security**: Is there any way to disable this functionality? I suppose some of the customers would prefer not to have it at all for security concern, mails are not really secure and the APIs don't require authentication (as expected from this kind of functionality)
- **Configuration**: would these APIs work whit all the possible Authentication configuration that we support? I guess in some case they would be useless and not really able to reset the password
- **Documentation**: I didn't find any documentation besides what we have in the code and in the API-explorer. Are you aware of documentation that the customer could use where we explain that the default configuration is not supposed to work and how to correctly implement the functionality?
- **Possible implementation**: we don't want to add new functionality to Share so the default configuration will never work

So, what is the best way to proceed at this point?
@skopf @jamalkm do you think can we proceed to deprecate them from the api-explorer without deleting the code for the moment?

jamalkm:
I'm not sure whether there is a documentation or not. Regarding configuration, this feature only works if the Repository is responsible for user management. In other words, it won't work if users being managed by LDAP or other external mechanisms.
I think it makes sense to deprecate the API from the api-explorer (similar to [listFavoriteSitesForPerson](https://api-explorer.alfresco.com/api-explorer/#/favorites/listFavoriteSitesForPerson)) and also, add deprecated annotation on the code itself for now, and remove the code in feature releases.

PR #560, opened 2021-06-25
SEARCH-2965 avoid exception for case sensitive for fields: TEXT, name, content, and description

jesty:
I also removed some useless toString() methods invocations.

In the past, there was a similar behaviour for the Score field, so I included also it in my fix.

### DIFF
--- a/data-model/src/main/java/org/alfresco/repo/search/impl/parsers/AlfrescoFunctionEvaluationContext.java
+++ b/data-model/src/main/java/org/alfresco/repo/search/impl/parsers/AlfrescoFunctionEvaluationContext.java
@@ -30,6 +30,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
 
+import java.util.Set;
 import org.alfresco.repo.dictionary.IndexTokenisationMode;
 import org.alfresco.repo.search.adaptor.LuceneFunction;
 import org.alfresco.repo.search.adaptor.QueryParserAdaptor;
@@ -54,7 +55,11 @@ import org.alfresco.service.namespace.QName;
 @SuppressWarnings("deprecation")
 public class AlfrescoFunctionEvaluationContext implements FunctionEvaluationContext
 {
-    private static HashSet<String> EXPOSED_FIELDS = new HashSet<String>();
+    private static Set<String> EXPOSED_FIELDS = new HashSet<>();
+
+    //The field names in the two collections below are fixed in getLuceneFieldName in order to allow users to use case insensitive field name
+    private static Set<String> CASE_INSENSITIVE_FIELDS_UPPERCASE = Set.of(QueryConstants.FIELD_TEXT);
+    private static Set<String> CASE_INSENSITIVE_FIELDS_LOWERCASE = Set.of("content", "name", "description", "score");
 
     private NamespacePrefixResolver namespacePrefixResolver;
 
@@ -233,7 +238,7 @@ public class AlfrescoFunctionEvaluationContext implements FunctionEvaluationCont
                 if (propertyDef.getDataType().getName().equals(DataTypeDefinition.CONTENT))
                 {
                     throw new FTSQueryException("Order on content properties is not curently supported");
-                }
+                } 
                 else if (propertyDef.getDataType().getName().equals(DataTypeDefinition.TEXT))
                 {
                     if(propertyDef.getIndexTokenisationMode() == IndexTokenisationMode.FALSE)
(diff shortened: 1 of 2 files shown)


data-model/src/main/java/org/alfresco/repo/search/impl/parsers/AlfrescoFunctionEvaluationContext.java
```diff
@@ -387,20 +392,26 @@ else if(dictionaryService.getDataType(fullQName) != null)
...
         else
         {
-            if(propertyName.equalsIgnoreCase("Score"))
+
+            if (CASE_INSENSITIVE_FIELDS_LOWERCASE.contains(propertyName.toLowerCase()))
             {
                 return propertyName.toLowerCase();
             }
+
+            if (CASE_INSENSITIVE_FIELDS_UPPERCASE.contains(propertyName.toUpperCase()))
```
alessandrobenedetti, line 409: Question shooting in the dark:
what was happening before with here "TEXT" ?
From the original code, it doesn't seem there was any special treatment for such pseudo-field, but no exception was thrown.
Are we sure this is the right place the lowercasing/uppercasing should happen?

jesty: Yes, before my change there was a similar behavior implemented for the score field:

```
 if(propertyName.equalsIgnoreCase("Score"))
{
    return propertyName.toLowerCase();
}
```

And when performing the query the exception is thrown at line 403 (now 414) by:
```
 throw new FTSQueryException("Unknown property: " + fullQName);
```

This method is where the name expansions happen, for instance when you search for cm:name (line 348).

Moreover, before pushing I tried if it works building a local docker image and running some searches using a mixed case.

jesty:
I missed "title" field in the list, I just added it

data-model/src/main/java/org/alfresco/repo/search/impl/parsers/AlfrescoFunctionEvaluationContext.java
alessandrobenedetti: I'll rephrase:

<Before your changes>
Using 'TEXT' was working fine -> this means TEXT didn't reach that conditional? what was happening in that case? was just returned as it was? 
Using 'text' was returning the exception. -> so on the other hand 'text' reached that point.

alessandrobenedetti: ok, line 386:

`if (EXPOSED_FIELDS.contains(propertyName))
        {
            return propertyName;
        }`

So We are actually case sensitive for a huge number of properties.
Why the few present in this ticket should be different?
my opinion is that we should be consistent and <be> or <not be> case sensitive for all the "special fields".
I guess we need Martin feedback on this before we refine this pull request

jesty: Ok, I see your point, basically, in this issue, we are allowing the mixed case for TEXT and every default text field (org.alfresco.repo.search.impl.elasticsearch.query.AlfrescoDefaultTextFields). Martin, is it enough?

mjstanford: I agree that the behavior should be consistent. There is no need to treat TEXT differently from other fields with regard to date sensitivity.

However, the priority right now is to release M3, and progress with our work for the next release cycle. This change is not a must-have. I am not overly concerned if the behavior change only applies to TEXT fields for now. This is still an improvement over the current situation on Solr. Further work to treat all fields consistently should only be undertaken at this point if trivial. If not, a separate ticket should be created.

jesty: @mjstanford I don't think that the further work to treat all fields consistently is trivial. So, now we can:
1) merge this PR as is and open a new ticket for other fields
2) Because this change is not a must-have we can freeze this PR and add the consistent behavior after the release.

mjstanford: I prefer option 1

alessandrobenedetti: My recommendation is to go with 2, I don't see much benefits nor any urgent priority in splitting up the work

PR #586, opened 2021-07-08
[ACS-1776] Add missing pipeline outlookMsgToLibreofficeViaText

tiagosalvado10:
This PR has been raised to ensure both master, 7.0.N and 6.2.N have the same 0100-basePipelines.json files as they all use the same versions of T-Engines.


alandavis (commented):
LGTM - identical to 7.0.N and master

PR #603, opened 2021-07-16
MNT-21534: Alfresco does not pull correct name for several scheduled jobs

amohammedalfresco:
- Add friendly name to scheduled jobs that are currently showing class names


alandavis (approved):
LGTM

PR #629, opened 2021-07-28
Node API create 2 versions instead of 1 when using cm:versionable

Giovanni007:
From MNT-22462/ACS-1842:
When a custom model is defined with` cm:versionable `mandatory aspect, this cause the node creation REST  API service to create 2 versions instead of one. The new document created has versions _1.0_ and _2.0_ created at the same time in the version history.

We should consider that:
- creating an empty file without content (via POST json) is **not** auto-versioned, and any subsequent file content updates are not versioned
- creating a file with content (via POST of multipart/form-data) is auto-versioned, and any subsequent file content updates are versioned
- If `versioningEnabled` is false and the node is **not** using cm:versionable, the version history is not created. 
- Right now, when the node is cm:versionable, setting versioningEnabled to false will avoid the issue of having version 2.0 but the API will still create version 1.0 in the history because of the cm:versionable aspect.

The goal of the fix is to disable the cm:versionable behaviour when the versioning is managed through the API (versioningEnabled=true or default behaviour when using POST of multipart/form-data calls)


Giovanni007:
Archived with https://github.com/Alfresco/alfresco-community-repo/pull/993

PR #649, opened 2021-08-05
Feature/acs 1601 Node cleanup job improvements for Postgres

nithinnambiar:
https://alfresco.atlassian.net/browse/ACS-1601
Implementation of a new algorithm to cleanup nodes and transactions in fixed sized batches. The default algorithm uses a time window approach which doesn't scale well for high volume node entries.
If `system.node_table_cleaner.algorithm` is set to V2 then node clean up job would use the new algorithm.


skopf (changes requested):
I can see multiple problems with this implementation:

1. It violates our separation of concerns pattern
In our architecture, only the DAOs are aware of the actual data model. All services and jobs above the DAOs implement the domain logic, but are agnostic of the actual storage of metadata. You can see in the current `DeletedNodeCleanupWorker` that it uses methods in the nodeDAO to get transaction ids and to purge nodes.
The class `DeletedNodeBatchCleanup` however is getting a DataSource and starts talking to the DB directly.

2. It violates our SQL templates pattern
Within a DAO, we do not use SQL statements directly. Instead, we use MyBatis to build SQL statements from templates. This pattern allows us to easily adopt queries for specific database dialects.
You can see this for example in `NodeDAOImpl.deleteTxnsUnused()` where we use the MyBatis templating engine to build and run the statement.
The class `DeletedNodeBatchCleanup` however is compiling SQL statements from hard coded strings. This makes it impossible to adopt to new SQL dialects or to handle problems with specific supported dialects, e.g. if a query is performing badly only with MSSQL.
MyBatis does have patterns to support these `IN (...)` patterns as needed here. Also it is likely that we already have templates for the required select statements.

Notes on dialect specific coding:
There are some very rare, exceptional cases where the MyBatis templating mechanism is not sufficient. For example if bugs in the JDBC driver prevent us from using plain clean SQL.
In these cases, we implement a dialect specific DAO and load it with a `HierarchicalBeanLoader` in spring. See `dao-context.xml` for multiple examples. Also a good reference are the recent improvements done by team 7.

3. Algorithm naming of "V2"
This naming suggests that this is a newer, or enhanced version of the old algorithm. But this is not the case. Instead, this is just an alternative version optimised for very large repositories.

4. Readability
The approach to have just one purge method which is able to handle the separate concerns of nodes and transactions all together makes the code really hard to understand.

5. Missing design documentation
The comment in the ticket only states "This would delete alf_node, alf_node_properties and alf_transaction table entries in fixed size batches.". The changes in this branch - if I read the code correct (see above) - seem to go beyond that. Can we please document exactly how the new strategy would look like. Team7 has already documented the existing strategy in comments on REPO-5379. We need a similar documentation which describes the strategy behind this housekeeping job.

6. Mix of different approaches
We have been discussing different approaches to enhance the entire cleanup job. These focused on individual problems with the existing implementation which have been identified in large scale performance tests. See REPO-5379 for details. This branch seems to mix up different concepts.

7. proof of efficency

8. Tests with different databases

9. keeping statement active
The approach used here where we keep one select stmt open and fire multiple delete statements wasn't working in other scenarios due to bugs in different JDBC drivers

nithinnambiar:
> I can see multiple problems with this implementation:
> 
> 1. It violates our separation of concerns pattern
>    In our architecture, only the DAOs are aware of the actual data model. All services and jobs above the DAOs implement the domain logic, but are agnostic of the actual storage of metadata. You can see in the current `DeletedNodeCleanupWorker` that it uses methods in the nodeDAO to get transaction ids and to purge nodes.
>    The class `DeletedNodeBatchCleanup` however is getting a DataSource and starts talking to the DB directly.
> 2. It violates our SQL templates pattern
>    Within a DAO, we do not use SQL statements directly. Instead, we use MyBatis to build SQL statements from templates. This pattern allows us to easily adopt queries for specific database dialects.
>    You can see this for example in `NodeDAOImpl.deleteTxnsUnused()` where we use the MyBatis templating engine to build and run the statement.
>    The class `DeletedNodeBatchCleanup` however is compiling SQL statements from hard coded strings. This makes it impossible to adopt to new SQL dialects or to handle problems with specific supported dialects, e.g. if a query is performing badly only with MSSQL.
>    MyBatis does have patterns to support these `IN (...)` patterns as needed here. Also it is likely that we already have templates for the required select statements.
> 
> Notes on dialect specific coding:
> There are some very rare, exceptional cases where the MyBatis templating mechanism is not sufficient. For example if bugs in the JDBC driver prevent us from using plain clean SQL.
> In these cases, we implement a dialect specific DAO and load it with a `HierarchicalBeanLoader` in spring. See `dao-context.xml` for multiple examples. Also a good reference are the recent improvements done by team 7.
> 
> 1. Algorithm naming of "V2"
>    This naming suggests that this is a newer, or enhanced version of the old algorithm. But this is not the case. Instead, this is just an alternative version optimised for very large repositories.
> 2. Readability
>    The approach to have just one purge method which is able to handle the separate concerns of nodes and transactions all together makes the code really hard to understand.
> 3. Missing design documentation
>    The comment in the ticket only states "This would delete alf_node, alf_node_properties and alf_transaction table entries in fixed size batches.". The changes in this branch - if I read the code correct (see above) - seem to go beyond that. Can we please document exactly how the new strategy would look like. Team7 has already documented the existing strategy in comments on [REPO-5379](https://alfresco.atlassian.net/browse/REPO-5379). We need a similar documentation which describes the strategy behind this housekeeping job.
> 4. Mix of different approaches
>    We have been discussing different approaches to enhance the entire cleanup job. These focused on individual problems with the existing implementation which have been identified in large scale performance tests. See [REPO-5379](https://alfresco.atlassian.net/browse/REPO-5379) for details. This branch seems to mix up different concepts.
> 5. proof of efficency
> 6. Tests with different databases
> 7. keeping statement active
>    The approach used here where we keep one select stmt open and fire multiple delete statements wasn't working in other scenarios due to bugs in different JDBC drivers

Points 1 and 2.
This design is inline with the changes Team 7 did for prop table cleanup jobs. Cristi suggested following a similar approach. Please refer to `org.alfresco.repo.domain.schema.script.DeleteNotExistsExecutor`
It makes direct jdbc calls to do batch deletion in a similar fashion. They do have a different implementation for MySql using `LIMIT` and `OFFSET` approach as MySql driver ignores the batch size.
see `org.alfresco.repo.domain.schema.script.MySQLDeleteNotExistsExecutor`. 
This Epic is specific to Postgres DB and hence the cursor based approach works well. I believe there is separate Epic for Mysql and that should take care of MySql specific changes.

Point 3.
What would you suggest as an alternative to "V2", do you recommend using "batch" or something else?

Point 4.
The class has two public methods,` purgeOldDeletedNodes()` and `purgeOldEmptyTransactions()` which invokes a private method `purge()`. Since the code was pretty much identical except few parameters
for node and transaction cleanup it would make sense to have one single `purge()` method to avoid code duplication. The IDE itself recommended it.

Point 5.
The low level design details are captured in the javadoc comments for the class. I have added some more detail to it. I can add more notes in jira if required.

Point 6.
I think this is what we agreed at a high level after the initial discussions. see https://alfresco.atlassian.net/browse/ACS-1601?focusedCommentId=1088567

Point 7.
The performance test results for Postgres DB could be found here - > https://alfresco.atlassian.net/browse/REPO-5380

Point 8.
As mentioned above this Epic changes are specific to Postgres DB. 

Point 9.
MySql would require a slightly different implementation based on `OFFSET` and `LIMIT`.

nithinnambiar:
DeletedNodeCleanupWorker has been refactored to use Mybatis Cursor to delete Nodes in fixed sized batches. This has been tested for functional correctness. 
This requires latest version of Mybatis  in Repo and Activiti.

jamalkm (approved):
LGTM

PR #671, opened 2021-08-19
[ACS-1885] libraries version bump

kcichonczyk:
Bumped versions of libraries to prevent security vulnerabilities:
* 'keycloak' version to 15.0.1
* 'mybaitis' version to 3.5.7

### DIFF
--- a/repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
+++ b/repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
@@ -194,21 +194,45 @@ public class SqlSessionMetricsWrapper implements SqlSession
     }
 
     @Override
-    public <T> Cursor<T> selectCursor(String s)
+    public <T> Cursor<T> selectCursor(String statement)
     {
-        return null;
+        long startTime = System.currentTimeMillis();
+        try
+        {
+            return this.sqlSession.selectCursor(statement);
+        }
+        finally
+        {
+            reportQueryExecuted(startTime, SELECT_LABEL, statement);
+        }
     }
 
     @Override
-    public <T> Cursor<T> selectCursor(String s, Object o)
+    public <T> Cursor<T> selectCursor(String statement, Object parameter)
     {
-        return null;
+        long startTime = System.currentTimeMillis();
+        try
+        {
+            return this.sqlSession.selectCursor(statement, parameter);
+        }
+        finally
+        {
+            reportQueryExecuted(startTime, SELECT_LABEL, statement);
+        }
     }
 
     @Override
-    public <T> Cursor<T> selectCursor(String s, Object o, RowBounds rowBounds)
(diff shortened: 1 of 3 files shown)


repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
```diff
@@ -192,6 +193,24 @@ private void reportQueryExecuted(final long startTime, final String queryTypeTag
         }
     }
 
+    @Override
+    public <T> Cursor<T> selectCursor(String s)
+    {
+        return null;
```
jamalkm, line 199: This requires implementation. See other methods in this class how they delegate to the default implementation of mybatis.

repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
```diff
@@ -192,6 +193,24 @@ private void reportQueryExecuted(final long startTime, final String queryTypeTag
...
+    }
+
+    @Override
+    public <T> Cursor<T> selectCursor(String s, Object o)
+    {
+        return null;
```
jamalkm, line 205: Same as above. Requires implementation.

repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
```diff
@@ -192,6 +193,24 @@ private void reportQueryExecuted(final long startTime, final String queryTypeTag
...
+    }
+
+    @Override
+    public <T> Cursor<T> selectCursor(String s, Object o, RowBounds rowBounds)
+    {
+        return null;
```
jamalkm, line 211: Same here. Requires implementation.

jamalkm (changes requested):
I left a few comments. However, to summarise:

- Keycloak client shouldn't be upgraded yet
- mybatis overridden methods should delegate to the default implementation rather simply returning `null`

repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
kcichonczyk: Implemented.

repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
kcichonczyk: Implemented.

repository/src/main/java/org/alfresco/ibatis/SqlSessionMetricsWrapper.java
kcichonczyk: Implemented.

alandavis:
The base issue is now closed

PR #678, opened 2021-08-24
SEARCH-2833 Add field for association Qname in NodeResource and ChildAssociationResource

jesty:
In order to implement the path live indexing, we need to retrieve the prefix string and the local name used in the association. 

In this PR we are setting the value to fields added in

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
+++ b/repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
@@ -46,6 +46,7 @@ import org.alfresco.repo.event2.filter.NodePropertyFilter;
 import org.alfresco.repo.security.authentication.AuthenticationUtil;
 import org.alfresco.repo.security.permissions.AccessDeniedException;
 import org.alfresco.service.cmr.dictionary.DictionaryService;
+import org.alfresco.service.cmr.repository.ChildAssociationRef;
 import org.alfresco.service.cmr.repository.ContentData;
 import org.alfresco.service.cmr.repository.MLText;
 import org.alfresco.service.cmr.repository.NodeRef;
@@ -148,6 +149,7 @@ public class NodeResourceHelper implements InitializingBean
                            .setModifiedByUser(getUserInfo((String) properties.get(ContentModel.PROP_MODIFIER), mapUserCache))
                            .setModifiedAt(getZonedDateTime((Date)properties.get(ContentModel.PROP_MODIFIED)))
                            .setContent(getContentInfo(properties))
+                           .setAssocQName(getAssocQName(nodeRef))
                            .setPrimaryHierarchy(PathUtil.getNodeIdsInReverse(path, false))
                            .setProperties(mapToNodeProperties(properties))
                            .setAspectNames(getMappedAspects(nodeRef));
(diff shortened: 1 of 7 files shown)


repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
```diff
@@ -158,6 +160,15 @@ private boolean isSubClass(QName className, QName ofClassQName)
         return dictionaryService.isSubClass(className, ofClassQName);
     }
 
+    private String getAssocQName(NodeRef nodeRef) {
```
killerboot, line 163: Please check formatting.

repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
```diff
@@ -158,6 +160,15 @@ private boolean isSubClass(QName className, QName ofClassQName)
...
+        String result = null;
+        ChildAssociationRef primaryParent = nodeService.getPrimaryParent(nodeRef);
```
killerboot, line 165: This call never returns null, see org.alfresco.repo.node.db.DbNodeServiceImpl#getPrimaryParent

repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
killerboot, line 163: Does it make sense to change the naming a bit as the assoc type here is only the primary parent-child assoc? There are also secondary parent-child and peer assocs that we are planning to implement later on.

jamalkm, line 163: Please use Alfresco format. 
Also, it makes sense to rename the method or add JavaDoc to indicate this only returns the primary parent-child association.

jesty: Should I call it primaryAssocQName? Do you mean also in acs-event-model, not only the private method, right?

jesty: @killerboot wrote a similar comment, Can I change the name to primaryAssocQName in order to clarify it?

jesty: done

repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
jesty: done

repository/src/test/java/org/alfresco/repo/event2/CreateRepoEventIT.java
```diff
@@ -32,6 +32,7 @@
 import java.util.List;
 
 import org.alfresco.model.ContentModel;
+import org.alfresco.repo.dictionary.NamespaceDAO;
```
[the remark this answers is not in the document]
jesty: removed, thanks

repository/src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
killerboot: yes, I think it should be ok to use "getPrimaryAssocQName"

jamalkm: > Do you mean also in acs-event-model, not only the private method

I think it makes sense to change that only in NodeResource. I'll add the comment on that PR. I think we should replicate our V1 REST API.

killerboot (approved):

jamalkm (approved):

PR #679, opened 2021-08-24
Fix ACS-1907

skopf:
- Added alternative approach for DBQueryEngine to page through the entire query using "LIMIT" requests instead of result set streaming
- Use the alternative approach for MySql


repository/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/db/DBQueryEngine.java
```diff
@@ -399,6 +411,54 @@ private boolean shouldCache(QueryOptions options, List<Node> nodes, int required
...
+        {
+            performTmdqSelectStreaming(statement, dbQuery, handler);
+        }
+    }
+
+    private void performTmdqSelectStreaming(String statement, DBQuery dbQuery, ResultHandler<Node> handler)
+    {
+        template.select(statement, dbQuery, handler);
+    }
+
```
alandavis, line 430: Straight forward but a small test that returns 2 to 3 pages of data would be good.

alandavis (approved):
I think we need a test that returns results from 2 to 3 pages. Other than that it is fine. I see you spotted my minor format changes.

repository/src/main/resources/alfresco/subsystems/Search/common-search-context.xml
```diff
@@ -105,8 +105,35 @@
             <ref bean="metadataQueryIndexesCheck2" />
         </property>
     </bean>
```
alandavis, line 107: Similar change needed in https://github.com/Alfresco/alfresco-enterprise-repo/blob/master/repository/src/main/resources/alfresco/subsystems/Search/common-search-enterprise-context.xml#L31-L55

repository/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/db/DBQueryEngine.java
```diff
@@ -399,6 +411,54 @@ private boolean shouldCache(QueryOptions options, List<Node> nodes, int required
...
+    private void performTmdqSelectPaging(String statement, DBQuery dbQuery, int requiredNodes, ResultHandler<Node> handler)
+    {
+        int batchStart = 0;
+        int batchSize = requiredNodes * 2;
+        batchSize = Math.max(Math.min(batchSize, MIN_PAGING_BATCH_SIZE), MAX_PAGING_BATCH_SIZE);
```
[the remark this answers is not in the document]
skopf: Of course. good spot. Where is the facepalm emoji when you need it!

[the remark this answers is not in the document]
skopf: fixed

repository/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/db/DBQueryEngine.java
skopf: Added test cases comparing results at different page sizes to show that paging has np impact on the overall result

alandavis (approved):
LGTM - merge once the enterprise PR is ready

PR #680, opened 2021-08-25
Feature/acs 1904 fix node ref vs content url

SaraAspery:
Ensure the content url rather than the node ref iss passed to the adapter. Also, remove deprecated methods

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
@@ -27,7 +27,6 @@ package org.alfresco.repo.content;
 
 import java.io.Serializable;
 import java.util.Collection;
-import java.util.Date;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
```diff
@@ -612,7 +605,7 @@ public boolean isContentDirectUrlEnabled(NodeRef nodeRef)
                 throw new IllegalArgumentException("The supplied nodeRef " + nodeRef + " has no content.");
             }
 
-            contentDirectUrlEnabled = (store.isContentDirectUrlEnabled(nodeRef));
+            contentDirectUrlEnabled = (store.isContentDirectUrlEnabled(getContentUrl(nodeRef)));
```
alandavis, line 608: Minor: Generally it is better to not nest method calls to simplify debug. However leave it this time.

alandavis (approved):
LGTM

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
SaraAspery: Fair point. Thanks

PR #688, opened 2021-09-01
Feature/acs 1924 remove version renditions end point

SaraAspery:
- Removed direct access url version renditions endpoint from NodeVersionsRelation
- Removed its associated test
- Also, correct WebApiParam names for all direct access url endpoints


alandavis (approved):
When we put this code back, we need to remember not to just revert the whole commit as we would want to keep the name and description changes. Can you please put that in the commit message.

PR #692, opened 2021-09-07
ACS-849_incorrect_probes added db check

nithinnambiar:
The changes related to https://alfresco.atlassian.net/browse/ACS-849
1) The liveness probe to return true if the spring context is up and running
2) Add db connection check as part of readiness probe.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
@@ -23,8 +23,10 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
+
 package org.alfresco.rest.api.probes;
 
+import org.alfresco.repo.admin.RepoHealthChecker;
 import org.alfresco.rest.api.discovery.DiscoveryApiWebscript;
 import org.alfresco.rest.api.model.Probe;
 import org.alfresco.rest.framework.WebApiDescription;
@@ -41,22 +43,24 @@ import org.apache.commons.logging.LogFactory;
 /**
  * An implementation of an Entity Resource for Probes.
  */
-@EntityResource(name="probes", title = "Probes")
-public class ProbeEntityResource implements EntityResourceAction.ReadById<Probe>
+@EntityResource(name = "probes", title = "Probes") public class ProbeEntityResource
+            implements EntityResourceAction.ReadById<Probe>
 {
     public static final String LIVE = "-live-";
     public static final String READY = "-ready-";
 
     public static final long CHECK_PERIOD = 10 * 1000; // Maximum of only one checkResult every 10 seconds.
 
-    protected static Log logger = LogFactory.getLog(ProbeEntityResource.class);;
-
+    protected static Log logger = LogFactory.getLog(ProbeEntityResource.class);
+    ;
+    private final Object lock = new Object();
+    private final Probe liveProbe = new Probe(getMessage(true, true, "Tested"));
     private long nextCheckTime = 0;
+    private long lastCheckTime = 0;
     private Boolean checkResult;
-    private Boolean checking = false;
-    private boolean readySent;
-
     private DiscoveryApiWebscript discovery;
+
+    private RepoHealthChecker repoHealthChecker;
 
     public DiscoveryApiWebscript setDiscovery(DiscoveryApiWebscript discovery)
     {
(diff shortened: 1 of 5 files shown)


remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -41,123 +43,104 @@
...
      */
-    @Override
-    @WebApiDescription(title="Get probe status", description = "Returns 200 if valid")
-    @WebApiParam(name = "probeName", title = "The probe's name")
-    @WebApiNoAuth
-    public Probe readById(String name, Parameters parameters)
+    @Override @WebApiDescription(title = "Get probe status", description = "Returns 200 if valid") @WebApiParam(name = "probeName", title = "The probe's name") @WebApiNoAuth public Probe readById(
+                String name, Parameters parameters)
     {
         boolean isLiveProbe = LIVE.equalsIgnoreCase(name);
```
tpage-alfresco, line 83: This looks like a better fit for an enum:
```ProbeType{ LIVE, READY }```

File: remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java [hunk 2]
```diff
@@ -41,123 +43,104 @@
...
     public static final String LIVE = "-live-";
     public static final String READY = "-ready-";
 
     public static final long CHECK_PERIOD = 10 * 1000; // Maximum of only one checkResult every 10 seconds.
 
-    protected static Log logger = LogFactory.getLog(ProbeEntityResource.class);;
-
-    private long nextCheckTime = 0;
+    protected static Log logger = LogFactory.getLog(ProbeEntityResource.class);
+    ;
```
tpage-alfresco, line 55: This colon is now better formatted, but still unnecessary! :yum:

remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -177,33 +160,31 @@ else if (!result && (isLiveProbe || readySent)) // Are we sick
 
     private String getMessage(boolean isLiveProbe, boolean result, String message)
```
tpage-alfresco, line 161: This method removes a small amount of redundancy, but adds complexity to the logic elsewhere. I think it would be better to get rid of this method completely and just log the message desired.

remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -41,123 +43,104 @@
...
+            else
+            {
+                // if no check is performed, use previous check result
+                result = checkResult;
+
+            }
         }
 
-        message = getMessage(isLiveProbe, result, message);
+        message = getMessage(false, result, message);
```
tpage-alfresco, line 143: If we get rid of the `getMessage` method then we can also use slf4j tokens to avoid string concatenation when debug logging is turned off:
```
LOGGER.debug("readyProbe: Success - {}", message);
```

remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -41,123 +43,104 @@
...
+
+            now = System.currentTimeMillis();
+            if (checkResult == null || isAfterCheckPeriod(now))
             {
-                result = checkResult;
+                doCheck = true;
             }
 
-            if (checking) // Is another thread is checking?
+            if (doCheck)
```
tpage-alfresco, line 114: The `doCheck` variable seems unnecessary. We should just say:
```
if (checkResult == null || isAfterCheckPeriod(now))
```

remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -41,123 +43,104 @@
...
-            }
-            else // This thread will do a check
-            {
-                now = System.currentTimeMillis();
-                if (checkResult == null || nextCheckTime <= now)
+                catch (Exception e)
                 {
-                    doCheck = true;
-                    checking = true;
+                    result = false;
```
tpage-alfresco, line 124: Is it worth debug or trace logging the exception?

remote-api/src/main/java/org/alfresco/rest/api/probes/ProbeEntityResource.java
```diff
@@ -177,33 +160,31 @@ else if (!result && (isLiveProbe || readySent)) // Are we sick
...
-            nextCheckTime += CHECK_PERIOD;
-        }
-        while (nextCheckTime <= now);
-
+        this.lastCheckTime = time;
+        long nextCheckTime = lastCheckTime + CHECK_PERIOD;
         if (logger.isTraceEnabled())
         {
-            logger.trace("nextCheckTime: " + nextCheckTime + " (+" + ((nextCheckTime - oldValue) / 1000) + " secs)");
+            logger.trace("nextCheckTime: " + nextCheckTime + " (+" + ((nextCheckTime) / 1000) + " secs)");
```
tpage-alfresco, line 181: We could use slf4j tokenisation here too.

remote-api/src/test/java/org/alfresco/rest/api/tests/ProbeApiTest.java
```diff
@@ -128,32 +136,28 @@ private void assertResponse(String probeName, Boolean ready, String expected, in
     public void testProbes() throws Exception
     {
         // Live first
-        assertResponse(LIVE, ERROR, "liveProbe: Failure - Tested", 503);
-        assertResponse(LIVE, null, "liveProbe: Failure - No test", 503); // Need to wait 10 seconds.
-        assertResponse(LIVE, null, "liveProbe: Failure - No test", 503);
-        assertResponse(READY, null, "readyProbe: Failure - No test", 503);
+        assertResponse(LIVE, OK, "liveProbe: Success - Tested", 200);
```
tpage-alfresco, line 139: I don't understand how this test works. Please can we split it up into lot of simpler unit tests?

tpage-alfresco (changes requested):
This already looks like you've improved things quite a bit. There are a few more things that I found it hard to understand at the moment.

PR #756, opened 2021-10-18
Fix/mnt 21506 sanitation create people

vrmoreira:


vrmoreira:
Thanks for the input, Stefan. The data-model already had some HTML escaping there, that's the reason I've changed in that way.
You're right, the change should be done at the presentation-layer. Reverted all changes on the data-model (on this PR) and I'll create a new PR for the presentation-layer on Share.

PR #763, opened 2021-10-19
Apply autoformat with prettier

mteodori:


jottley:
This was a draft that was never completed.  Closing.

PR #764, opened 2021-10-21
ACS-2147: Adding Experimental annotation.

mpichura:
As per @skopf 's comment in PR#761 - moving addition of `@Experimental` annotation to a separate PR.


skopf:
@montgolfiere Not sure about the package. Should we have this in org.alfresco.api next to the AlfrescoPublicAPI annotation? Feels similar.

skopf:
But isn't this annotation API related?

PR #766, opened 2021-10-21
Fix small probelm in README

skopf:
Fix maven command for acs-community-packaging


jottley:
@skopf Should this still be addressed?

skopf:
If nobody else ran into the same issue, I think this is a none problem. Closing PR and deleting the branch

PR #789, opened 2021-11-05
[ACS-2158] Added additional retries for getting a node rendition cont…

kcichonczyk:
…ent in the rendition tests (due to timeouts in the terraform-alfresco-pipeline tests).


alandavis:
Closing - change made directly to master - the original commit on this branch was reverted so the issue would have been wrong.

PR #790, opened 2021-11-05
MNT-21446 - Prevented XSS attacks by encoding the minor and major sym…

antoniojfelix:
…bols (< >)

### DIFF
--- a/remote-api/src/main/resources/alfresco/templates/webscripts/org/alfresco/repository/upload/upload.post.js
+++ b/remote-api/src/main/resources/alfresco/templates/webscripts/org/alfresco/repository/upload/upload.post.js
@@ -91,6 +91,13 @@ function escapeRegExp(string) {
    return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+
+function escapeHTML(str){
+  return String(str).replace(/[<>]/gi, function(c){
+     return '&#'+c.charCodeAt(0)+';';
+  });
+}
+
 function main()
 {
    try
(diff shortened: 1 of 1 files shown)


antoniojfelix:
This attack happens in a javascript file on the server side. There’s a validation for special characters regarding the data that is being transmitted to the server, and the Integrity Violation error still occurs for the fields with special characters. But after that, a popup appears in the current page.

I was able to prevent this by replacing every minor (<) and major (>) symbol for its correspondent ascii code. Since these are special chars they shouldn’t be allowed anyway, and because of that there’s no need to replace them back as well. And the attack is prevented.

jottley (approved):

skopf (changes requested):
First of all, the HTML escaping is not sufficient. You are just handling two characters.

But mainly, I am wondering why you need to escape the filename on a POST. The filename is not the problem. The problem is the dialog that tries to display the filename.

antoniojfelix:
Hi @skopf, thanks for your input.

I thought that replacing these two chars would be enough to avoid the XSS attack, but I can improve the RegEx if needed.

The problem is that this attack is only possible by intercepting the request of the upload POST with an external tool (Burp suite in this case) and adding the vulnerability code in any of the form-data fields. For example:

------WebKitFormBoundaryFE1LgroisF9Q30XZ
Content-Disposition: form-data; name="filename"

a.csv``

In this case the exploit is applied to the filename but it can be applied to any of the form-data fields, like **uploaddirectory**, **destination**, **containerid**, etc.

When the dialog tries to display the filename it gives **Integrity Violation** error and that was already covered. But in the response the html tag is applied in the current page causing the alert to appear.

skopf:
Exactly. We need to escape the filename and probably other fields too where they are added to the response HTML. Presumably, this webscript has a FTL file associated. I guess in the ftl file, we just have ${filename} instead of ${filename?html}

antoniojfelix:
Hi, after a chat with @skopf I found out a better way to solve this issue from the client side. So instead of having changes in the alfresco-community-repo project we will have changes in the alfresco-enterprise-share project. Therefore I will close this PR without merging any changes. Thanks @skopf for your time.

PR #802, opened 2021-11-16
ACS-2222 Add ArchivedIOException

PunchedChimera:
Adds a Precondition Failed (412) error response when attempting the get content that is Archived.

Discovered a bug:
When an error is thrown while attempting to access content the Content-Length header is set but no content is returned (obviously we've had a error so there shouldn't be any 😄). This causes requesting clients to hang until they timeout, as they are expecting content to be there when none will be returned.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/framework/webscripts/AbstractResourceWebScript.java
+++ b/remote-api/src/main/java/org/alfresco/rest/framework/webscripts/AbstractResourceWebScript.java
@@ -230,7 +230,7 @@ public abstract class AbstractResourceWebScript extends ApiWebScript implements 
         res.setHeader(HEADER_CONTENT_LENGTH, String.valueOf(-1));
         if (exception instanceof ArchivedIOException)
         {
-            renderException(new ArchivedContentException(exception.getLocalizedMessage(), exception), res, assistant);
+            renderException(new ArchivedContentException(exception.getMsgId(), exception), res, assistant);
         }
         else
         {
--- a/remote-api/src/main/java/org/alfresco/rest/framework/webscripts/AbstractResourceWebScript.java
+++ b/remote-api/src/main/java/org/alfresco/rest/framework/webscripts/AbstractResourceWebScript.java
@@ -42,6 +42,7 @@ import org.alfresco.rest.framework.core.ResourceLocator;
 import org.alfresco.rest.framework.core.ResourceOperation;
 import org.alfresco.rest.framework.core.ResourceWithMetadata;
 import org.alfresco.rest.framework.core.exceptions.ApiException;
+import org.alfresco.rest.framework.core.exceptions.ArchivedContentException;
 import org.alfresco.rest.framework.resource.actions.ActionExecutor;
 import org.alfresco.rest.framework.resource.actions.interfaces.BinaryResourceAction;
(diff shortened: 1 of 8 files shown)


aymanthefirst (approved):
Looks good to me. You could add the reasons why you are doing this to the PR description in case someone who is checking does not have access to Jira.

KikoUK (approved):

mpichura (approved):

remote-api/src/main/java/org/alfresco/rest/framework/webscripts/AbstractResourceWebScript.java
```diff
@@ -173,6 +176,12 @@ public Void doWork() throws Exception
             }
 
         }
+        catch (ContentIOException cioe)
+        {
+            // If the Content-Length is not set back to -1 any client will expect to receive binary and will hang until it times out
```
montgolfiere, line 181: Nice catch 👍

remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -174,6 +174,7 @@
                 <entry key="org.springframework.http.InvalidMediaTypeException" value="#{T(org.springframework.extensions.webscripts.Status).STATUS_UNSUPPORTED_MEDIA_TYPE}" />
                 <entry key="org.alfresco.rest.framework.core.exceptions.ServiceUnavailableException" value="503" />
                 <entry key="org.alfresco.service.cmr.dictionary.InvalidTypeException" value="409" />
+                <entry key="org.alfresco.service.cmr.repository.ArchivedIOException" value="#{T(org.springframework.extensions.webscripts.Status).STATUS_PRECONDITION_FAILED}" />
```
montgolfiere, line 177: Minor: As discussed,  n addition to 412 status code will this return unique errorKey or ...ApiDefault for now. 

If latter, could / should we resolve & remap to a new org.alfresco.rest.framework.core.exceptions.ArchivedIOException (or similar).

montgolfiere (approved):

KikoUK (approved):

PunchedChimera:
Expected error msg will now be something like:
```json
{
    "error": {
        "errorKey": "framework.exception.ArchivedContent",
        "statusCode": 412,
        "briefSummary": "10180000 Content is currently in an Archived state and is inaccessible",
        "stackTrace": "For security reasons the stack trace is no longer displayed, but the property is kept for previous versions",
        "descriptionURL": "https://api-explorer.alfresco.com"
    }
}
```

remote-api/src/main/resources/alfresco/public-rest-context.xml
PunchedChimera: Thanks @montgolfiere I've updated so that we now ultimately throw a `org.alfresco.rest.framework.core.exceptions.ArchivedContentException` after catching the `ArchivedIOException`, which extends `ApiException`. This allows us to control the errorKey in the response. 
To provide greater control I have also provided additional constructors within `ApiException` to allow us to have different errorKey and briefSummary.
The response will now look like the [comment below](https://github.com/Alfresco/alfresco-community-repo/pull/802#issuecomment-973171371), with errorKey being a default string provided by `ArchivedContentException` and the breifSummary being provided by the thrown `ArchivedIOException`.

PR #807, opened 2021-11-19
ACS-2151 Get StorageProps endpoint

mpichura:
This PR is still work in progress.
All below changes are annotated `@Experimental` as they are subject to change.

- Added `NodeStorageInfoRelation` to handle REST request for get StorageProps 
- Added `ContentStorageInfo` model class to expose StorageProps iin REST response
- Added `ContentStorageInformation` interface and its implementation (along with unit test) as a proxy to get StorageProps

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/ContentStorageInformationImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/ContentStorageInformationImpl.java
@@ -1,0 +1,82 @@
+/*
+ * #%L
+ * Alfresco Remote API
+ * %%
+ * Copyright (C) 2005 - 2021 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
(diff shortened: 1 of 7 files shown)


remote-api/src/main/java/org/alfresco/rest/api/impl/ContentStorageInformationImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        final Map<String, String> storageProperties = contentService.getStorageProperties(nodeRef, propQName);
+        final ContentStorageInfo storageInfo = new ContentStorageInfo();
+        storageInfo.setId(propQName.toPrefixString(namespaceService));
+        storageInfo.setStorageProperties(storageProperties);
+        return storageInfo;
+    }
+
+    private QName getQName(final String contentPropName)
+    {
+        //this splitting may be gone when we decide to use colon as prefix separatot
```
montgolfiere, line 79: Can this comment line be removed ? 

For now, we can simply document (in REST API Explorer specs) what we do support - ie. similar to the JavaDoc example (ie. "cm_content").

montgolfiere (approved):
If not already, we should raise jira to implement integration test to call this REST API when deployed with S3 (eg. within S3 Connector project).

PunchedChimera (approved):
Looks good

PR #813, opened 2021-11-22
ACS-2216: Fix incorrect ContentService Java Interface - should accept content propQName

aymanthefirst:
-  Added  new methods to use propQName instead of PROP_CONTENT to be consistent with the rest of the project. 
-  Also added their corresponding tests.
-  Deprecated methods that use PROP_CONTENT

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
@@ -3420,7 +3420,7 @@ public class NodesImpl implements Nodes
     @Override
     public DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachment, Long validFor)
     {
-        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(nodeRef, attachment, validFor);
+        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(nodeRef, ContentModel.PROP_CONTENT, attachment, validFor);
         if (directAccessUrl == null)
         {
             throw new DisabledServiceException("Direct access url isn't available.");
--- a/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
@@ -592,7 +592,8 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
      */
     @Override
     @Deprecated
-    public boolean isContentDirectUrlEnabled(NodeRef nodeRef) {
+    public boolean isContentDirectUrlEnabled(NodeRef nodeRef)
+    {
         return isContentDirectUrlEnabled(nodeRef, ContentModel.PROP_CONTENT);
     }
 
--- a/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
@@ -4,21 +4,21 @@
  * %%
  * Copyright (C) 2005 - 2021 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software.
- * If the software was purchased under a paid Alfresco license, the terms of
- * the paid license agreement will prevail.  Otherwise, the software is
+ * This file is part of the Alfresco software. 
+ * If the software was purchased under a paid Alfresco license, the terms of 
+ * the paid license agreement will prevail.  Otherwise, the software is 
  * provided under the following open source license terms:
- *
+ * 
  * Alfresco is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * 
  * Alfresco is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- *
+ * 
  * You should have received a copy of the GNU Lesser General Public License
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
@@ -79,14 +79,14 @@ import org.springframework.extensions.surf.util.I18NUtil;
  * Note: This class was formerly the {@link RoutingContentService} but the
  * 'routing' functionality has been pushed into the {@link AbstractRoutingContentStore store}
  * implementations.
- *
+ * 
  * @author Derek Hulley
  * @since 3.2
  */
 public class ContentServiceImpl implements ContentService, ApplicationContextAware
 {
     private static final Log logger = LogFactory.getLog(ContentServiceImpl.class);
-
+    
     private DictionaryService dictionaryService;
     private NodeService nodeService;
     private MimetypeService mimetypeService;
@@ -108,29 +108,29 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
      * The policy component
      */
     private PolicyComponent policyComponent;
-
+    
     /*
      * Policies delegates
      */
     ClassPolicyDelegate<ContentServicePolicies.OnContentUpdatePolicy> onContentUpdateDelegate;
     ClassPolicyDelegate<ContentServicePolicies.OnContentPropertyUpdatePolicy> onContentPropertyUpdateDelegate;
     ClassPolicyDelegate<ContentServicePolicies.OnContentReadPolicy> onContentReadDelegate;
-
+    
     public void setRetryingTransactionHelper(RetryingTransactionHelper helper)
     {
         this.transactionHelper = helper;
     }
-
+    
     public void setDictionaryService(DictionaryService dictionaryService)
     {
         this.dictionaryService = dictionaryService;
     }
-
+    
     public void setNodeService(NodeService nodeService)
     {
         this.nodeService = nodeService;
     }
-
+    
     public void setMimetypeService(MimetypeService mimetypeService)
     {
         this.mimetypeService = mimetypeService;
@@ -185,16 +185,16 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
                 NodeServicePolicies.OnUpdatePropertiesPolicy.QNAME,
                 this,
                 new JavaBehaviour(this, "onUpdateProperties"));
-
+        
         // Register on content update policy
         this.onContentUpdateDelegate = this.policyComponent.registerClassPolicy(OnContentUpdatePolicy.class);
         this.onContentPropertyUpdateDelegate = this.policyComponent.registerClassPolicy(OnContentPropertyUpdatePolicy.class);
         this.onContentReadDelegate = this.policyComponent.registerClassPolicy(OnContentReadPolicy.class);
     }
-
+    
     /**
      * Update properties policy behaviour
-     *
+     * 
      * @param nodeRef    the node reference
      * @param before    the before values of the properties
      * @param after        the after values of the properties
@@ -209,7 +209,7 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
         {
             return;
         }
-
+    	
         // Don't duplicate work when firing multiple policies
         Set<QName> types = null;
         OnContentPropertyUpdatePolicy propertyPolicy = null;            // Doesn't change for the node instance
@@ -236,7 +236,7 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
                 // We don't fire notifications for multi-valued content properties
                 continue;
             }
-
+            
             try
             {
                 ContentData beforeValue = (ContentData) before.get(propertyQName);
@@ -245,7 +245,7 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
                         && (!ignoreEmptyContent || beforeValue.getSize() > 0);
                 boolean hasContentAfter = ContentData.hasContent(afterValue)
                         && (!ignoreEmptyContent || afterValue.getSize() > 0);
-
+                
                 // There are some shortcuts here
                 if (!hasContentBefore && !hasContentAfter)
                 {
@@ -257,10 +257,10 @@ public class ContentServiceImpl implements ContentService, ApplicationContextAwa
                     // Still, nothing happening
                     continue;
                 }
-
+                
                 // Check for new content
                 isNewContent = isNewContent || !hasContentBefore && hasContentAfter;
-
+                
                 // Make it clear when there's no content before or after
                 if (!hasContentBefore)
                 {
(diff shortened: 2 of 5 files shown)


repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
```diff
@@ -591,14 +591,24 @@ public boolean isContentDirectUrlEnabled()
      * {@inheritDoc}
      */
     @Override
+    @Deprecated
     public boolean isContentDirectUrlEnabled(NodeRef nodeRef)
+    {
+        return isContentDirectUrlEnabled(nodeRef, ContentModel.PROP_CONTENT);
+    }
```
PunchedChimera, line 598: This should be moved to ContentService.java as a `default` method call.
See `ContentService.requestContentDirectUrl(NodeRef nodeRef, boolean attachment)` for an example.

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -170,8 +170,18 @@ public ContentWriter getWriter(NodeRef nodeRef, QName propertyQName, boolean upd
      *
      * @return {@code true} if direct access URLs retrieving is supported for the node, {@code false} otherwise
      */
+    @Deprecated
     boolean isContentDirectUrlEnabled(NodeRef nodeRef);
```
PunchedChimera, line 174: As mentioned above. Should be changed to a default method, calling the new implementation, something like this:
```suggestion
    @Deprecated
    default boolean isContentDirectUrlEnabled(NodeRef nodeRef)
    {
        return isContentDirectUrlEnabled(nodeRef, ContentModel.PROP_CONTENT);
    }
```

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
```diff
@@ -615,14 +625,25 @@ public boolean isContentDirectUrlEnabled(NodeRef nodeRef)
     /**
      * {@inheritDoc}
      */
+    @Override
+    @Deprecated
     public DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachment, Long validFor)
+    {
+        return requestContentDirectUrl(nodeRef, ContentModel.PROP_CONTENT, attachment, validFor);
+    }
```
PunchedChimera, line 633: As above.

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -197,8 +223,24 @@ default DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachm
      * @throws UnsupportedOperationException if the store is unable to provide the information.
      */
     @Auditable(parameters = {"nodeRef", "validFor"})
+    @Deprecated
     DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachment, Long validFor);
```
PunchedChimera, line 227: Needs a to be a default method (as mentioned above) to call the new method.

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -181,11 +191,27 @@ public ContentWriter getWriter(NodeRef nodeRef, QName propertyQName, boolean upd
...
         return requestContentDirectUrl(nodeRef, attachment, null);
     }
 
+    /**
+     * Gets a presigned URL to directly access the content. It is up to the actual store
+     * implementation if it can fulfil this request with an expiry time or not.
+     *
+     * @param nodeRef Node ref for which to obtain the direct access {@code URL}.
+     * @param attachment {@code true} if an attachment URL is requested, {@code false} for an embedded {@code URL}.
+     * @param propertyQName the name of the property, which must be of type <b>content</b>
```
montgolfiere, line 206: Trivial: re-order in JavaDoc to match order in method sig

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
```diff
@@ -142,6 +145,21 @@ public void testRequestContentDirectUrl_SystemWideIsDisabled()
...
+        try
+        {
+            contentService.requestContentDirectUrl(NODE_REF, QNAME, true, 20L);
+            fail("Expected DirectAccessUrlDisabledException");
+        }
+        catch (DirectAccessUrlDisabledException ex)
+        {
+            verify(mockContentStore, never()).isContentDirectUrlEnabled();
+        }
+    }
```
PunchedChimera, line 161: Not needed. Just update the method signature in the old test. Once deprecated the old signature should eventually be removed from the product.

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
```diff
@@ -153,6 +171,17 @@ public void testRequestContentDirectUrl_SystemWideIsEnabledButStoreIsDisabled()
...
+    public void testRequestContentDirectUrlwithQname_SystemWideIsEnabledButStoreIsDisabled()
+    {
+        setupSystemWideDirectAccessConfig(ENABLED);
+        when(mockContentStore.isContentDirectUrlEnabled()).thenReturn(DISABLED);
+
+        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(NODE_REF, QNAME,true, 20L);
+        assertNull(directAccessUrl);
+        verify(mockContentStore, never()).requestContentDirectUrl(anyString(), eq(true), anyString(), anyString(), anyLong());
+    }
+
```
PunchedChimera, line 184: As above.

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
```diff
@@ -164,6 +193,17 @@ public void testRequestContentDirectUrl_StoreIsEnabledButNotImplemented()
...
+    public void testRequestContentDirectUrlwithQname_StoreIsEnabledButNotImplemented()
+    {
+        setupSystemWideDirectAccessConfig(ENABLED);
+        when(mockContentStore.isContentDirectUrlEnabled()).thenReturn(ENABLED);
+
+        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(NODE_REF, QNAME, true, 20L);
+        assertNull(directAccessUrl);
+        verify(mockContentStore, times(1)).requestContentDirectUrl(anyString(), eq(true), anyString(), anyString(), anyLong());
+    }
+
```
PunchedChimera, line 206: Not needed. Just update the method signature in the old test. Once deprecated the old signature should eventually be removed from the product.

repository/src/test/java/org/alfresco/repo/version/ContentServiceImplTest.java
```diff
@@ -187,4 +188,48 @@ public void testWhenRequestContentDirectUrlIsNotSupported()
         assertNull(contentService.requestContentDirectUrl(nodeRef, true, null));
         assertNull(contentService.requestContentDirectUrl(nodeRef, true, validFor));
     }
+
+    @Test
+    public void testwithQnameWhenRequestContentDirectUrlIsNotSupported()
```
PunchedChimera, line 193: As above.

montgolfiere (changes requested):
@aymanthefirst looks good so far. I've added a trivial comment. I haven't reviewed the test methods yet.

One thing missing is to change usages - ie. any existing calls to use new non-deprecated methods (eg. in remote-api).

PunchedChimera (changes requested):
Good first pass. I don't think the extra tests are necessary, they just need to be updated to the new method.
See comments re ContentService.java & ContentServiceImpl.java

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
mpichura: I wouldn't change interface method in a public project and hardcode that here. 
Very unlikely, but there may be custom implementations for this method which then would be broken. 
Same bit of code resides in `ContentServiceImpl` for this method, which, if you ask me, is safer.

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
mpichura: I would keep the tests - once the deprecated methods are removed - these test of course need to be removed as well.

skopf (dismissed):
Please don't use deprecated methods in the rest api. Instead, the cm:content prop name needs to be hard coded in the rest api, just as we have with the /content endpoint as well for example.
But we should not use these deprecated methods in the rest api.
We want to remove these deprecated methods pretty soon. having code relying on them makes this more difficult.

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
aymanthefirst: This should be resolved now but please double check it is what you meant

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
aymanthefirst: This should be resolved now but please double check it is what you meant

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
aymanthefirst: This should be resolved now but please double check it is what you meant

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
aymanthefirst: Resolved. See https://github.com/Alfresco/alfresco-community-repo/pull/813/commits/ca25e3929b24c9fb3bbb5ff89190930839a09a04

montgolfiere (changes requested):
As per my earlier comment (& Stefan & David )- we need to review & change all usages (tests & remote-api etc) to use new non-deprecated methods.

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
```diff
@@ -78,6 +79,8 @@
     private static final String X_AMZ_HEADER_2 = "x-amz-header-2";
     private static final String VALUE_2 = "value2";
 
+    private static final QName QNAME = ContentModel.PROP_CONTENT;
```
montgolfiere, line 82: Trivial: it's not just any QNAME - maybe rename to PROP_CONTENT_QNAME to be explict.

repository/src/test/java/org/alfresco/repo/version/ContentServiceImplTest.java
```diff
@@ -173,7 +174,7 @@ public void testWhenRequestContentDirectUrlIsNotSupported()
 
         try
         {
-            assertNull(contentService.requestContentDirectUrl(null, true, null));
+            assertNull(contentService.requestContentDirectUrl(null, null, true, null));
```
montgolfiere, line 177: Add -ve test for what happens if qname is passed in as "null" (and all other args are valid including nodeRef).

montgolfiere (changes requested):
As per my earlier comment (& Stefan & David )- we need to review & change all usages (tests & remote-api etc) to use new non-deprecated methods.

Also, added a couple of other minor suggested changes.

repository/src/test/java/org/alfresco/repo/version/ContentServiceImplTest.java
aymanthefirst: This should be resolved now but please double check

PunchedChimera (commented):
Looks good. Thanks for making the changes I previously requested 👍 
I agree with @montgolfiere, might be worth adding the extra test.
Any tests you're changing/updating could also be refactor to use assertThrows() instead of try-catch.

KikoUK (approved):

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -191,13 +206,46 @@ default DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachm
...
+     * implementation if it can fulfil this request with an expiry time or not.
+     *
+     * @param nodeRef Node ref for which to obtain the direct access {@code URL}.
+     * @param propertyQName the name of the property, which must be of type <b>content</b>
      * @param attachment {@code true} if an attachment URL is requested, {@code false} for an embedded {@code URL}.
      * @param validFor The time at which the direct access {@code URL} will expire.
      * @return A direct access {@code URL} object for the content.
      * @throws UnsupportedOperationException if the store is unable to provide the information.
      */
     @Auditable(parameters = {"nodeRef", "validFor"})
```
montgolfiere, line 247: Add "propertQName" to list of @Auditable params ?

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3420,7 +3420,7 @@ public Node unlock(String nodeId, Parameters parameters)
     @Override
     public DirectAccessUrl requestContentDirectUrl(NodeRef nodeRef, boolean attachment, Long validFor)
     {
-        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(nodeRef, attachment, validFor);
+        DirectAccessUrl directAccessUrl = contentService.requestContentDirectUrl(nodeRef, ContentModel.PROP_CONTENT, attachment, validFor);
```
montgolfiere, line 3423: This looks good. What about all other usages of the deprecated methods ?

montgolfiere (changes requested):
Please see comment(s) - I assume there are other usages of the deprecated methods ?

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
aymanthefirst: Do you mean I should find methods in other projects such as rest-api-explorer? Because according to IntelliJ, there are no more usages in this project.
image

montgolfiere: After quick review with @aymanthefirst it looks like all REST APIs go via the NodesImpl after all (including for Deleted Nodes &/or Renditions).

repository/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
aymanthefirst: done

montgolfiere:
@aymanthefirst Thanks for the additional updates LGTM (cc @skopf @PunchedChimera).

montgolfiere (approved):

repository/src/test/java/org/alfresco/repo/content/ContentServiceImplUnitTest.java
```diff
@@ -133,7 +136,7 @@ public void testRequestContentDirectUrl_SystemWideIsDisabled()
         setupSystemWideDirectAccessConfig(DISABLED);
         try
         {
-            contentService.requestContentDirectUrl(NODE_REF, true, 20L);
+            contentService.requestContentDirectUrl(NODE_REF, PROP_CONTENT_QNAME, true, 20L);
```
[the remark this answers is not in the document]
aymanthefirst: Please double-check this is what you meant https://github.com/Alfresco/alfresco-community-repo/pull/813/commits/4df90ca37e9fbafe03eeadcc13376949fd7ef9b5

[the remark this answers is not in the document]
PunchedChimera: Yes it is, but you do not need the fail() method in an assertThrows. It will fail by itself if the error isn't thrown.

PunchedChimera (approved):
Thanks for making those changes

PR #821, opened 2021-11-29
[MNT-21901] Prevent update preferences concurrency errors

tiagosalvado10:


jottley (approved):

repository/src/main/java/org/alfresco/repo/preference/script/ScriptPreferenceService.java
```diff
@@ -76,8 +77,16 @@ public NativeObject getPreferences(String userName)
     public NativeObject getPreferences(String userName, String preferenceFilter)
     {
         // It's a tad unusual to return a NativeObject like this - at least within Alfresco.
-        // But we can't change it to e.g. a ScriptableHashMap as the API is published.
-        Map<String, Serializable> prefs = this.preferenceService.getPreferences(userName, preferenceFilter);        
+        // But we can't change it to e.g. a ScriptableHashMap as the API is published.
+        Map<String, Serializable> prefs = this.transactionService.getRetryingTransactionHelper().doInTransaction(new RetryingTransactionCallback<Map<String, Serializable>>()
```
killerboot, line 81: Why do you need to wrap the calls to preferenceService in a transaction? Isn't it better to modify the spring bean to use the upper case "PreferenceService" with a transaction interceptor instead?

repository/src/main/java/org/alfresco/repo/preference/script/ScriptPreferenceService.java
```diff
@@ -76,8 +77,16 @@ public NativeObject getPreferences(String userName)
...
-        Map<String, Serializable> prefs = this.preferenceService.getPreferences(userName, preferenceFilter);        
+        // But we can't change it to e.g. a ScriptableHashMap as the API is published.
+        Map<String, Serializable> prefs = this.transactionService.getRetryingTransactionHelper().doInTransaction(new RetryingTransactionCallback<Map<String, Serializable>>()
+        {
+            @Override
+            public Map<String, Serializable> execute() throws Throwable
+            {
+                return preferenceService.getPreferences(userName, preferenceFilter);
+            }
+        }, true, true);
```
killerboot, line 88: Why do you need to create a new transaction here? It is usually a very dangerous approach. Especially, if it will start a new transaction within a thread that already has a transaction, this may result in a deadlock inside a DB.

core/src/main/java/org/alfresco/error/ExceptionStackUtil.java
```diff
@@ -38,10 +38,16 @@ public static Throwable getCause(Throwable throwable, Class<?> ... possibleCause
...
+            boolean isJavaScriptException = throwableClass.getName().contains("JavaScriptException");
+            String throwableMsg = throwable.getMessage() != null ? throwable.getMessage() : "";
+
             for (Class<?> possibleCauseClass : possibleCauses)
             {
-                Class<?> throwableClass = throwable.getClass();
-                if (possibleCauseClass.isAssignableFrom(throwableClass))
+                String possibleCauseClassName = possibleCauseClass.getName();
+
+                if (possibleCauseClass.isAssignableFrom(throwableClass) || (isJavaScriptException && throwableMsg.contains(possibleCauseClassName)))
```
tpage-alfresco, line 50: Is there a delimiter we could use?  Sometimes class names contain each other.

tpage-alfresco (approved):

core/src/main/java/org/alfresco/error/ExceptionStackUtil.java
tiagosalvado10: Added a colon as delimiter.

repository/src/test/java/org/alfresco/repo/preference/PreferenceServiceImplTest.java
```diff
@@ -240,7 +247,75 @@ public void testBadUser()
             }
         });
     }
-    
+
+    @Test
+    public void testConcurrentPreferencesUpdate() throws InterruptedException
```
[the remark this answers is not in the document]
tiagosalvado10: Test removed from this class.

PR #822, opened 2021-12-01
ACS-2321 Downloads throws 412 error if content is archived

PunchedChimera:


remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -87,6 +100,8 @@ public Download createDownloadNode(Download download)
         
         checkNodeIdsReadPermission(zipContentNodeRefs);
         
+        checkArchiveStatus(zipContentNodeRefs);
```
montgolfiere, line 103: Add discovery - only makes sense to call if either S3 Connector (or Azure Connector in future) are deployed.

montgolfiere (commented):
I've added a comment about only running if S3 Connector or Azure Connector are deployed.

More generally, this request (to create a "download" operation) could take some time if the folder hierarchies are deep - ie. large number of content nodes to process. 

Maybe we need a configurable cut-off, eg. after first 500 content items.

At that point, defer to background action and then ensure if ArchiveException occurs that this is caught and handled appropriately, ie. marking the download as failed, with reason/error code (if possible).

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: Thanks, I've added the discovery using the `ModuleService` and add a hardcoded limit of 500 to the number of checks.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -172,4 +198,99 @@ private Download getStatus(NodeRef downloadNodeRef)
...
+            // We hit the number of nodes we want to check.
+            if (cache.size() == checkLimit) 
+            {
+                logger.info(
+                        String.format(
+                                "Maximum check of %d reached for archived content. No more checks will be performed and download will still be created.",
+                                checkLimit));
+                return;
+            }
+            // Already checked this node, we can skip.
```
montgolfiere, line 251: How likely (when) will a node appear more than once (given we're only walking primary child assocs) ? 

Could we use two HashSet instead (eg. folderRefs and nonFolderRefs) and then only track a total count ?

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -172,4 +198,99 @@ private Download getStatus(NodeRef downloadNodeRef)
...
+        {
+            NodeRef[] childRefs = nodeService.getChildAssocs(nodeRef).stream()
+                                                                     .map(childAssoc -> childAssoc.getChildRef())
+                                                                     .toArray(NodeRef[]::new);
+            checkArchiveStatus(childRefs, checkLimit, cache); // We'll keep going until we have no more folders in children.
+        }
+    }
+
+    @Experimental
+    protected boolean canCheckArchived()
```
montgolfiere, line 285: Do we need this extra method (could we move up into protected "checkArchiveStatus" ?

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -87,6 +111,8 @@ public Download createDownloadNode(Download download)
         
         checkNodeIdsReadPermission(zipContentNodeRefs);
         
+        checkArchiveStatus(zipContentNodeRefs, 500);
```
montgolfiere, line 114: Magic number - at minimum should be a constant. We should consider make it a configurable property (ie. add to "repository.properties" and inject into this bean.

montgolfiere (commented):
I've added a few minor comments.

We should also log in JIRA separate tech debt to improve the handling of errors that might occur for larger downloads (eg. if beyond our initial limit when accepting the request).

KikoUK (approved):
LGTM once other comments are addressed.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -172,4 +198,99 @@ private Download getStatus(NodeRef downloadNodeRef)
...
+    @Experimental
+    private void checkArchiveStatus(NodeRef[] nodeRefs, int checkLimit, List<NodeRef> cache)
+    {
+        // Create the cache for recursive calls.
+        if (cache == null) 
+        {
+            cache = new ArrayList<NodeRef>();
+        }
+
+        var folders = new ArrayList<NodeRef>();
```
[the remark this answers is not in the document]
PunchedChimera: Yes, I agree. I'm adding to the next commit.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: It's unlikely but possible. Image the following, simple, file heirarchy: folder1/folder2/content. The following download request is legal: 
{  
    "nodeIds": [    
        "node for folder1",
        "node for folder2",
        "node for content
    ]
}
You can see that we could attempt to look at the child refs of folder2 and check content multiple times.
I have attempted to replicate this complexity in the unit tests provided.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: Doesn't look like it, I'll move in the next commit. I think I was keep my options open for unit tests but never used it.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: Thanks, refactoring to global property to be overriden by the connectors

PunchedChimera: Thanks, refactoring to global property to be overriden by the connectors

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
```diff
@@ -172,4 +198,99 @@ private Download getStatus(NodeRef downloadNodeRef)
...
+                                                                     .map(childAssoc -> childAssoc.getChildRef())
+                                                                     .toArray(NodeRef[]::new);
+            checkArchiveStatus(childRefs, checkLimit, cache); // We'll keep going until we have no more folders in children.
+        }
+    }
+
+    @Experimental
+    protected boolean canCheckArchived()
+    {
+        return moduleServiceContains("org_alfresco_integrations_AzureConnector") ||
```
mpichura: I'm happy with these string literals (and `moduleServiceContains` solution) but would it be possible to either move them to constants or even do something like:
```
public static final String [] CLOUD_CONNECTOR_MODULES = {"org_alfresco_integrations_AzureConnector", "org_alfresco_integrations_S3Connector"};
@Experimental
protected boolean canCheckArchived() {
    return Arrays.stream(CLOUD_CONNECTOR_MODULES).anyMatch(m-> moduleService.getModule(m) != null);
}
```

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
montgolfiere: Ah ok - thanks.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
montgolfiere: Thanks. I'm not sure connectors need to override (do they) ? Wouldn't a global (ie. system-wide) be sufficient ?

mpichura (approved):
Some suggestions which not necessarily need to be considered in this iteration.

remote-api/src/test/java/org/alfresco/rest/api/impl/DownloadsImplUnitTest.java
```diff
@@ -0,0 +1,184 @@
...
+        // No archived nodes, each content node should only be checked once
+        downloads.checkArchiveStatus(nodeRefsToTest, -1);
+        verify(contentService, times(1)).getStorageProperties(contentNode1, TYPE_CONTENT);
+        verify(contentService, times(1)).getStorageProperties(contentNode2, TYPE_CONTENT);
+        verify(contentService, times(1)).getStorageProperties(contentNode3, TYPE_CONTENT);
+        verify(contentService, times(1)).getStorageProperties(contentNode4, TYPE_CONTENT);
+        verify(contentService, times(1)).getStorageProperties(contentNode5, TYPE_CONTENT);
+        verify(contentService, times(1)).getStorageProperties(contentNode6, TYPE_CONTENT);
+        
+        reset(contentService);
```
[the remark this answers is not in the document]
PunchedChimera: Refactored and fixed the tests 👍

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: Love this, added 👍 
We could easily, make the CLOUD_CONNECTOR_MODULES a config bean later if needed.

remote-api/src/main/java/org/alfresco/rest/api/impl/DownloadsImpl.java
PunchedChimera: I have since refactor and re-added this method. The new implementation recommended by @mpichura is better sat in it's own method.

montgolfiere (approved):
Thanks for the updates. 

Do we need jira reminder to review cleanup of pending downloads (eg. if error occurs midway) ?

PR #827, opened 2021-12-06
ATS-969 Upgrade tika (not poi)

pzhyland:
### Noticeable changes:
1. Tika 2.x (parsers to be precise) doesn't bring the same transitive dependencies as Tika 1.x
    * Explicitly included needed dependencies for projects depending on transitive ones.
    * Removed excludes which are no longer needed.
2. New configuration option `loadErrorHandler` not possible to set to `IGNORE` the only option which allows us to run in case of misconfigured classpath) is to set it to `WARN`.
### What about POI?
Tika 2.1.0 still depend on the POI 4.1.2 so we shouldn't upgrade it.

### DIFF
--- a/data-model/pom.xml
+++ b/data-model/pom.xml
@@ -230,13 +230,9 @@
         </dependency>
         <dependency>
             <groupId>org.apache.tika</groupId>
-            <artifactId>tika-parsers</artifactId>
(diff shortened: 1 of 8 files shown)


alandavis (approved):
LGTM

kcichonczyk (approved):

PR #834, opened 2021-12-08
ACS-2365 license plugin

tpage-alfresco:


SaraAspery (approved):
LGTM! Do we ever need to add headers to files other than .java and .jsp (e.g. yaml files)?

PR #836, opened 2021-12-09
MNT-22715 - Setting Version Child Assoc Index Configurable

evasques:
* Added configuration to use child association index on version creation - disabled by default
* Added unit test to verify both cenarios
* Included configuraion in [ags]

### DIFF
--- a/amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/version/RecordableVersionServiceImpl.java
+++ b/amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/version/RecordableVersionServiceImpl.java
@@ -434,7 +434,11 @@ public class RecordableVersionServiceImpl extends    Version2ServiceImpl
                         QName.createQName(Version2Model.NAMESPACE_URI, Version2Model.CHILD_VERSIONS + "-" + versionNumber),
                         sourceTypeRef,
                         null);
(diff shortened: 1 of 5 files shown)


evasques:
Summary:

- AGS breaks with the the fix as class RecordableVersionServiceImpl creates versions in methods createNewRecordedVersion and createRecordFromLatestVersion without using versionService but by creating the child assoc themselves. I also added the fix here but this  will only be present when AGS is next released (7.1.N / 7.2)

- To not render this HF on 7.0.1 incompatible with AGS, I switched the fix to be configurable and it’s disabled by default. If it is to be made compatible with AGS, a new HF for AGS must be released with the changes you can find in this PR.

- To enable the index on version child assocs, in alfresco-global.properties set:
version.store.useVersionAssocIndex=true

- Once this is enabled when a new version of a node is created, the child association will have the index set to the current amount of versions of that node instead of the default value of -1. This also makes it compatible with older nodes that already have versions (and index set to -1) as the ordering is by assoc_index and then by id. If you have a problem with the sequence of Id’s (as it’s the case in this issue) the update script needs to be ran to update the assoc_index of old versions

- Once this is enabled you should not disable it, or else new versions will have the new index of -1 (the default) and that is not compatible with having older versions with index value > 0. If it is to be disabled again, the same update script provided above must be ran but instead of setting the assoc_index by the count column of the sub-query, it needs to be set to -1, the default value.

jottley (approved):

PR #870, opened 2022-01-04
MNT-22668: missing encoded characters for some characters like ‘(’ and ‘)’

kcichonczyk:
Replaced org.springframework.extensions.surf.util.URLEncoder with the java.net.URLEncoder for encoding filenames containing characters like ")" and "(" in the content-disposition response header.


alandavis (approved):

antoniorv75:
The problem persists with the version 7.3.0-M1

aborroy:
This patch seems to be working as expected:
https://github.com/aborroy/mnt-22668

PR #872, opened 2022-01-05
ACS-2242 Attempt to fix Initialization of SecureRandom in EncryptionTest

kmagdziarz:


repository/src/test/java/org/alfresco/encryption/EncryptionTests.java
```diff
@@ -101,6 +101,18 @@
 
 	private UserTransaction tx;
 
+	static SecureRandom random;
+	static {
```
pzhyland, line 105: Consider moving initialisation to a static method. It will allow the field to be `final`.

repository/src/test/java/org/alfresco/encryption/EncryptionTests.java
```diff
@@ -101,6 +101,18 @@
...
 
+	static SecureRandom random;
+	static {
+		try
+		{
+			random = SecureRandom.getInstance("SHA1PRNG");
+			random.setSeed(System.nanoTime());
+		} catch (NoSuchAlgorithmException e)
+		{
+			e.printStackTrace();
```
pzhyland, line 112: Throw some kind of `RuntimeException` here. not throwwing exception here will lead to the NPEs later on.

repository/src/test/java/org/alfresco/encryption/EncryptionTests.java
```diff
@@ -101,6 +101,18 @@
 
 	private UserTransaction tx;
 
+	static SecureRandom random;
```
pzhyland, line 104: Follow the convention and use capital letters for `static` field names

pzhyland (approved):

PR #883, opened 2022-01-11
ACS-2180 solr shared secret

skopf:


killerboot (approved):
LGTM.
Additional work will be required in a separate set of tickets, see linked stories to ACS-2180.

PR #885, opened 2022-01-12
MNT-22563: bumped jdom2 version to 2.0.6.1

vrmoreira:


pom.xml
```diff
@@ -851,6 +852,11 @@
                 <version>1.18.22</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
```
pzhyland, line 855: Could you put it closer to the `tika` dependencies with some comment that if tika is upgraded we would need to check if hardcoded jdom2 dependency version is still needed?

pzhyland (approved):

PR #908, opened 2022-01-25
ACS-2104: Changing apache.camel and netty-codec-http dependency versions

mpichura:
Due to dependency conflicts (in Azure Connector AMP) and vulnerabilities apache.camel and netty-codec-http dependency versions have to be changed. 
This PR is mostly dealing with netty.io related version bumps (due to vulnerability found in netty-codec-http library). 
Community-repo is providing these libraries in AMPs and there (combined with AMP specific dependencies) a dependency conflict is very likely to happen.
A follow up PR will be created in Azure Connector AMP project to properly handle provided dependencies and exclude conflicting Azure libraries.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
         <dependency.ooxml-schemas.version>1.4</dependency.ooxml-schemas.version>
         <dependency.keycloak.version>15.0.2</dependency.keycloak.version>
         <dependency.jboss.logging.version>3.4.3.Final</dependency.jboss.logging.version>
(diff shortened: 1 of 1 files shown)


skopf (commented):
No need for me to review. The test cases will tell us if this combination of library versions is working or causing any trouble.

montgolfiere:
Given that we're bumping Camel and, in this case, specifically overridding version of netty-codec-http

Could (should ?) we go further and also attempt to:
- explicitly set the same version for all netty components currently pulled-in by our Camel dependencies
- also add additional netty components pulled-in by other OOTB amps (eg. Azure Identity)

Obviously subject to build / tests.

In future releases, we can then look to converge again when we next upgrade Camel (&/or Netty) components.

alandavis (approved):

PR #917, opened 2022-01-28
ACS-2104: Remove Camel exclusion from DependABot

montgolfiere:
- we're now at Camel 3.7.7 ( hence remove Camel exclusion of > 3.7.1 )
- to enable future suggestions, eg. should prompt with 3.7.7 to 3.14.0 (or maybe 3.14.1)
- also logged ACS-2476 (in case of compile-time/run-time incompatibilities and aligning with Netty)


mpichura (approved):
LGTM

tpage-alfresco (approved):

PR #918, opened 2022-01-28
Solr: Add missing flag mapping for NodesMetaDataGet

AFaust:
In https://github.com/Alfresco/SearchServices/commit/943fafcca383a995db221805fed2cd9036ec66e2 @aborroy explicitly set the flag for SOLR not to load child associations from the ACS repository for regular update / creation indexing, as child associations and child ids are not relevant for this processing step. This PR adds the missing parameter flag mapping in the repository-tier web script to actually make this change take effect. Without this, the repository will always load child associations regardless of what the SOLR request specifies, as the default value for that flag if not mapped / set is `true`. In repositories with folders that contain potentially thousands or (against sane advice) millions of items, this can cause extreme pressure on the system.


jottley:
@AFaust @aborroy Is this PR still relevant or needed?

aborroy:
It's still needed, thanks @jottley

jottley:
This was merged as #2441

PR #926, opened 2022-02-01
Add nodeRef (node) to alfresco-access audit log

Muffexx:
Add nodeRef (node) to the default audit log (alfresco-access).
This way the audit entries can be associated with the actual nodes.

There are several tickets on this topic in the "Alfresco Hub". Some examples:

1. https://hub.alfresco.com/t5/alfresco-content-services-forum/how-to-get-audit-trail-of-existing-documents-with-new-audit/td-p/196663
2. https://hub.alfresco.com/t5/alfresco-content-services-forum/alfresco-access-get-noderef-nodeid-or-nodeuuid/td-p/311175

IMHO this is quite useful, as the currently used "path" alone is not sufficient to establish a unambiguous relation between node and audit record. After all, the "path" can change (move/delete) and thus a relation is no longer feasible without further ado.


joleaf:
@Muffexx, this feature is exactly what we need in our context. The Audit-Log is useless for some of out analytic tasks without the NodeRef.

albie91:
****Push ***** I am also looking forward to this feature.

Muffexx:
Bump.

PR #950, opened 2022-02-09
Fix/mnt 22780 fix rule behavior

damianujma:


MateuszBiel (approved):

skopf (changes requested):
x

skopf (commented):
Is it possible that the node has both aspects applied, the parent aspect and the child aspect? Not sure if this is even possible.
Something you could quickly check. If so, we still have a problem.

damianujma:
> 

Yes, there is a possibility to have both aspects applied. In this case, if we want to remove the child aspect, then the inherited properties will also be removed.

jamalkm (approved):

aonych (approved):

PR #951, opened 2022-02-09
MNT-22807: improve sitesTest test (#925)

vrmoreira:
(cherry picked from commit 1b42d80f940b7bf679a5a836ae4d36b598e92c0a)


jottley (approved):

vrmoreira:
Although this is just a CP, I'm adding @jottley because we're doing this the "wrong" way: first commit was in "release/6.2.2" because of the code-freeze and now we're CP into master.

PR #957, opened 2022-02-10
ACS-2476 Bump Camel version to 3.8.0

gevagel:


montgolfiere (approved):

tpage-alfresco (approved):

montgolfiere:
As per discussion with @mpichura this will likely need to be reverted since it may break Azure Connector downstream.

We will need to take extra upgrade steps to align Camel/Netty/Azure.

PR #1056, opened 2022-04-07
Fix/mnt 22946 spring rce databind jdk9 71 n

vrmoreira:


vrmoreira:
Reverted spring-security upgrade because it's not related to this issue and I couldn't get a feedback for this upgrade.

PR #1084, opened 2022-04-22
REPO-5712 Fix exception logging in EventGeneratorQueue

jlosornogil:
Include the exception stacktrace so that any event framework
wrong configuration can be easily detected.


jottley:
The JIRA issue was closed as Won't Fix

PR #1085, opened 2022-04-22
ACS-2925: Elasticsearch - add support for OpenCMIS and DB Switching

montgolfiere:
- enable OpenCMIS & DB Switching for Elasticsearch connector
- see also PR for alfresco-enterprise-repo
- note: ACS-2860 & ACS-2920

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/search/impl/DbOrIndexSwitchingQueryLanguage.java
+++ b/repository/src/main/java/org/alfresco/repo/search/impl/DbOrIndexSwitchingQueryLanguage.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * Copyright (C) 2005 - 2022 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software. 
  * If the software was purchased under a paid Alfresco license, the terms of 
@@ -23,7 +23,7 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  * #L%
  */
-package org.alfresco.repo.search.impl.solr;
+package org.alfresco.repo.search.impl;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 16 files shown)


montgolfiere:
Note: do not merge until after 23.1.0-M1

repository/src/main/java/org/alfresco/repo/search/impl/DbOrIndexSwitchingQueryLanguage.java
```diff
@@ -110,6 +113,11 @@ public void setHybridEnabled(boolean hybridEnabled)
         this.hybridEnabled = hybridEnabled;
     }
 
+    public void setSubsystemName(String subsystemName)
+    {
+        this.subsystemName = subsystemName;
```
montgolfiere, line 118: Note: only used for logging. I'm sure if there is a way to derive at runtime (instead of config property).

repository/src/main/resources/alfresco/subsystems/Search/noindex/noindex-search-context.xml
```diff
@@ -99,7 +99,13 @@
             <ref bean="search.indexerAndSearcherFactory" />
         </property>
     </bean>
-    
+
+    <bean id="search.fts.alfresco.switching" parent="base.search.fts.alfresco.switching" class="org.alfresco.repo.search.impl.DbOrIndexSwitchingQueryLanguage" >
```
krdabrowski, line 103: Maybe it will be better to define class type `class="org.alfresco.repo.search.impl.DbOrIndexSwitchingQueryLanguage"` within (abstract) bean parent. It won't be required to repeat it so many times.

montgolfiere: Ah, I think I know what you mean. Good idea if it works - I'll try it 🤞

montgolfiere: @krdabrowski I've incorporated your suggestion. Thanks 👍 

Marking this thread as resolved and will ask you for a re-review.

montgolfiere:
Closing without merge. Branch was renamed, hence new PR ( see https://github.com/Alfresco/alfresco-community-repo/pull/1088 )

PR #1102, opened 2022-05-05
Origin/feature/apps 1164 alfresco logs

Sbisht19:
added test group NOT_SUPPORTED_ON_SINGLE_PIPELINE for the test cases trying to access the alfresco logs so these test cases will continue to run in component pipeline and we would exclude this group in single pipeline to skip these test cases to run.


kavitshah-gl (approved):
The changes looks fine. These test cases will run only for local docker-compose.

tpage-alfresco (approved):

PR #1105, opened 2022-05-05
ACS-2745: Adding security mechanism for private actions to rule POST/PUT

mstrankowski:
Implemented checking mechanism for private actions in AbstractRuleWebScript class.
Added RuleServiceTestUtil and extending classes to externalise build methods and make it easier to choose what implementation should be built.
Refactored RuleServiceTest mostly by reordering methods inside to go like setUp-> terminate -> test cases -> checking methods -> build methods. I feel it's cleaner and easier to follow than before.
I've added 2 test cases to RuleServiceTest in regards to PrivateActions.


remote-api/src/main/java/org/alfresco/repo/web/scripts/rule/AbstractRuleWebScript.java
```diff
@@ -432,10 +442,40 @@ private Serializable convertValue(QName typeQName, Object propertyValue) throws
 
     protected void checkRule(Rule rule)
     {
+        List<Action> actions = getAllActions(rule);
+        blockPrivateActions(actions);
+        validateOutboundRuleActions(rule, actions);
+    }
+
+    private List<Action> getAllActions(Rule rule) {
```
mstrankowski, line 450: I really don't like having to use instanceof here, but I just don't have any other idea. I could try to add "getAllActions" to Action interface and implement it accordingly, but it seems to clash between CompositeAction and Action interfaces idea.

remote-api/src/test/java/org/alfresco/repo/web/scripts/rule/util/DefaultRuleServiceTestUtil.java
```diff
@@ -0,0 +1,125 @@
...
+ * #L%
+ */
+package org.alfresco.repo.web.scripts.rule.util;
+
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class DefaultRuleServiceTestUtil implements RuleServiceTestUtil {
```
mstrankowski, line 33: Extracted build methods for JSON to this class from RuleServiceTest, so they could be used interchangeably with ability to choose on test case level.

remote-api/src/test/java/org/alfresco/repo/web/scripts/rule/RuleServiceTest.java
```diff
@@ -1,38 +1,40 @@
-/*
```
mstrankowski, line 1: There is a ton of modifications shown, but it's mostly refactoring, moving methods so that they would go in order setUp -> shutdown -> tests -> check methods -> additional methods. With JSON building methods extracted to RuleJsonUtil.
The only real changes here are addition of 2 tests:
testCreatePrivateActionRule
testUpdatePrivateActionRule

addition of custom createRule method, allowing to pass JSONObject consumer and expected status,
addition of verifyPrivateActionFailureResponse method validating error response.

remote-api/src/main/java/org/alfresco/repo/web/scripts/rule/AbstractRuleWebScript.java
```diff
@@ -432,10 +442,40 @@ private Serializable convertValue(QName typeQName, Object propertyValue) throws
...
+            actionList.add(action);
+        }
+
+        return actionList;
+    }
+
+    private void blockPrivateActions(List<Action> actions) {
+        for (Action action : actions)
+        {
+            ActionExecutionContext aec = ActionExecutionContext.builder(action.getActionDefinitionName())
```
mstrankowski, line 471: Was the intent behind the check about "action.getActionDefinitionName()" as I've assumed? For action.getId() all tests failed and it also seemed counter-intuitive

pzhyland: Yes, the `getActionDefinitionName` returns the bean id which is used to identify the logic behind the action.

pzhyland (approved):

mstrankowski:
Test failure on reject operation even when AGS config marks it as public.

   <bean id="reject" class="org.alfresco.module.org_alfresco_module_rm.action.impl.RejectAction" parent="rmAction">
      <property name="publicAction" value="true"/>
   </bean>
   
My suspicion is that the isExposed method doesn't manage AGS actions properly yet, nothing to do with the code I've implemented.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/repo/action/ExtendedActionServiceImpl.java
```diff
@@ -49,12 +53,17 @@
  */
 public class ExtendedActionServiceImpl extends ActionServiceImpl implements ApplicationContextAware
 {
+    private static final Set<String> ALLOWED_CONTEXTS_FOR_RM_ACTIONS = Set.of("rules");
```
mstrankowski, line 56: Any ideas of how to do it better? I feel kinda bad with comparing values through a string variable and not at least a static variable if not even an enum, but I also don't want to add any unneeded dependencies. I don't know how the repo is structured, is there a place to add such an enum/static?

repository/src/main/java/org/alfresco/repo/action/ActionServiceImpl.java
```diff
@@ -279,7 +279,11 @@ public void init()
         {
             configProperties = applicationContext.getBean("global-properties", Properties.class);
         }
-        actionExecutionValidator = new ActionExecutionValidator(configProperties::getProperty, actionDefinitions::containsKey);
+        actionExecutionValidator = new ActionExecutionValidator(configProperties::getProperty, isActionPublicPredicate());
+    }
+
+    protected Predicate<ActionExecutionContext> isActionPublicPredicate() {
```
mstrankowski, line 285: I've changed the type provided for predicate to allow for ExtendedActionServiceImpl override including overriding only isPublic predicate, so that it could check the context the action is called from and enable Records Management actions only for rules.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/repo/action/ExtendedActionServiceImpl.java
```diff
@@ -49,12 +53,17 @@
...
+
     /** File plan service */
     private FilePlanService filePlanService;
 
     /** Application context */
     private ApplicationContext extendedApplicationContext;
 
+
+    private Set<String> recordManagementActions;
```
pzhyland, line 65: Volatile of AtomicReference could be used to avoid possibility of other thread not seeing changes from other thread.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/repo/action/ExtendedActionServiceImpl.java
```diff
@@ -72,6 +81,29 @@ public void setFilePlanService(FilePlanService filePlanService)
 		this.filePlanService = filePlanService;
 	}
 
+    @Override
+    protected Predicate<ActionExecutionContext> isActionPublicPredicate() {
+        return aec -> {
```
pzhyland, line 86: I think more idiomatic way would be something like:
```java
return super.isActionPublicPredicate().or(this::isAllowedRmAction)
```
Just personal opinion.

mstrankowski:
Change of approach due to many collisions on RM actions, instead of trying to get private actions to be actually, strictly, private, the approach switches to just verifying security of all private/public actions, hence this change is no longer needed.

PR #1141, opened 2022-05-30
ACS-3013: Uppercasing services used to allow security proxy

mstrankowski:
Not much happens in class, check node, create version.
Added Beans including security interceptors, but VersionService has no restrictions and for node service all there is are a few checks, so it likely never was a security issue to begin with.

### DIFF
--- a/repository/src/main/resources/alfresco/action-services-context.xml
+++ b/repository/src/main/resources/alfresco/action-services-context.xml
@@ -375,10 +375,10 @@
          <value>false</value>
       </property>
       <property name="nodeService">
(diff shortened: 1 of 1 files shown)


mstrankowski:
Should be no change at all aside from attempting to access private nodeRef not from the user calling it.
Added services with security interceptors, since they exist, if there will be issues with access then it's a matter of deciding permissions. Though no unsafe operation can be done with a new version anyway. This literally is a description + boolean of major/minor version release, no way to pass some destructive data or honestly mess up anything.

PR #1153, opened 2022-06-10
Feature/acs 3015 analyse error prone action executor action

SaraAspery:
The `ErrorProneActionExecutor` appears to have been written specifically for use by the `FeedNotifierTest` and belongs in the test code folder. Likewise, `MockUserNotifier`, which is not currently referenced at all.


pzhyland (approved):
LGTM

PR #1159, opened 2022-06-14
ACS-2322 Don't compare audit properties for CMIS TCK RootFolderTest

pzhyland:


remote-api/src/test/java/org/alfresco/rest/api/tests/AlfrescoCMISBasicsTestGroup.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.chemistry.opencmis.tck.tests.basics.RepositoryInfoTest;
+import org.apache.chemistry.opencmis.tck.tests.basics.RootFolderTest;
+import org.apache.chemistry.opencmis.tck.tests.basics.SecurityTest;
+
+class AlfrescoCMISBasicsTestGroup extends AbstractSessionTestGroup
+{
+    @Override
+    public void init(Map<String, String> parameters) throws Exception
+    {
+        super.init(parameters);
```
mstrankowski, line 47: How about that instead?
Not forcing, just thought it would be better in case there are any updates in external library in the future adding more tests in their implementation, that aren't RootFolderTest

`    @Override
    public void init(Map<String, String> parameters) throws Exception
    {
        super.init(parameters);

        replaceRootFolderTest();
    }

    private void replaceRootFolderTest() throws Exception {
        getTests().removeIf(t -> t instanceof RootFolderTest);
        addTest(new RootFolderTest()
        {
            @Override
            protected CmisTestResult assertEquals(CmisObject expected, CmisObject actual, CmisTestResult success, CmisTestResult failure, boolean checkAcls, boolean checkPolicies)
            {
                return super.assertEquals(hideAsynchronouslyChangedProperties(expected), hideAsynchronouslyChangedProperties(actual), success, failure, checkAcls, checkPolicies);
            }
        });
    }`

mstrankowski: Also class would extend BasicsTestGroup then

remote-api/src/test/java/org/alfresco/rest/api/tests/AlfrescoCMISBasicsTestGroup.java
```diff
@@ -0,0 +1,76 @@
...
+        {
+            @Override
+            protected CmisTestResult assertEquals(CmisObject expected, CmisObject actual, CmisTestResult success, CmisTestResult failure, boolean checkAcls, boolean checkPolicies)
+            {
+                return super.assertEquals(hideAsynchronouslyChangedProperties(expected), hideAsynchronouslyChangedProperties(actual), success, failure, checkAcls, checkPolicies);
+            }
+        });
+    }
+
+    private CmisObject hideAsynchronouslyChangedProperties(final CmisObject target)
```
mstrankowski, line 64: Haven't tested yet, but wouldn't Mockito.spy be easier to use here?


    private CmisObject hideAsynchronouslyChangedProperties(final CmisObject target)
    {
        CmisObject spiedObject = spy(target);
        when(spiedObject.getProperties()).then(a -> {
            List<Property<?>> properties = (List<Property<?>>) a.callRealMethod();
            return properties.stream()
                    .filter(p -> !p.getId().startsWith("cmis:lastMod"))
                    .collect(Collectors.toUnmodifiableList());
        });
        
        return spiedObject;
    }

remote-api/src/test/java/org/alfresco/rest/api/tests/AlfrescoCMISBasicsTestGroup.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.chemistry.opencmis.client.api.CmisObject;
+import org.apache.chemistry.opencmis.client.api.Property;
+import org.apache.chemistry.opencmis.tck.CmisTestResult;
+import org.apache.chemistry.opencmis.tck.tests.basics.BasicsTestGroup;
+import org.apache.chemistry.opencmis.tck.tests.basics.RootFolderTest;
+
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+
+class AlfrescoCMISBasicsTestGroup extends BasicsTestGroup
```
mstrankowski, line 42: modified

remote-api/src/test/java/org/alfresco/rest/api/tests/AlfrescoCMISBasicsTestGroup.java
```diff
@@ -0,0 +1,76 @@
...
+{
+    @Override
+    public void init(Map<String, String> parameters) throws Exception
```
mstrankowski, line 45: modified

remote-api/src/test/java/org/alfresco/rest/api/tests/AlfrescoCMISBasicsTestGroup.java
```diff
@@ -0,0 +1,76 @@
...
+    {
+        super.init(parameters);
+
+        replaceRootFolderTest();
+    }
+
+    private void replaceRootFolderTest() throws Exception {
```
mstrankowski, line 52: Replacing test by using the fact that getTests() method returns actual modifiable list reference

PR #1189, opened 2022-06-29
Feature/acs 3169 implement security mechanism for mail action

mstrankowski:


repository/src/main/java/org/alfresco/repo/action/ActionServiceImpl.java
```diff
@@ -298,7 +303,7 @@ public ActionDefinition getActionDefinition(String name)
      */
     public List<ActionDefinition> getActionDefinitions()
     {
-        return new ArrayList<ActionDefinition>(this.actionDefinitions.values());
```
mstrankowski, line 301: just refactor of existing code

repository/src/main/java/org/alfresco/repo/action/ActionServiceImpl.java
```diff
@@ -588,11 +593,24 @@ public void executeAction(Action action, NodeRef actionedUponNodeRef, boolean ch
         }
         else
         {
+            verifyActionAccessRestrictions(action);
```
mstrankowski, line 596: Added verification here for async action executions to check restrictions immediatelly. For async actions they will run when they are created and when they are executed, a bit of redundancy, but didn't see a feasible way to avoid it.

remote-api/src/main/java/org/alfresco/repo/web/scripts/rule/ActionQueuePost.java
```diff
@@ -1,35 +1,36 @@
...
- * 
- * You should have received a copy of the GNU Lesser General Public License
- * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
- * #L%
- */
+/*
+ * #%L
+ * Alfresco Remote API
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
SaraAspery, line 5: update year

repository/src/main/java/org/alfresco/repo/action/access/ActionAccessRestrictionAbstractBase.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        innerVerifyAccessRestriction(action);
+    }
+
+    private boolean isActionExposed(Action action) {
+        return !isActionFromControlledContext(action) || isExposedInConfig(action).orElse(Boolean.FALSE);
+        //TODO consider idea of action from not controlled contexts but restricted in config
+        //e.g. new, custom customer action context
+        //I don't think it has that much merit though
```
SaraAspery, line 78: do you still want this TODO?

repository/src/main/java/org/alfresco/repo/action/access/AdminActionAccessRestriction.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Only admin can run action access restriction
+     *
+     * @param action
+     */
+    protected void innerVerifyAccessRestriction(Action action) {
+        boolean isAdminOrSystemUser = authorityService.hasAdminAuthority() || AuthenticationUtil.isRunAsUserTheSystemUser();
+        if (!isAdminOrSystemUser) {
+            throw new ActionAccessException("Only admin or system user is allowed to define uses of/directly execute this action");
```
SaraAspery, line 49: I think we generally try to avoid using '/' in messages, could just use 'or'

SaraAspery (commented):
Is it definitely necessary to distinguish between v0 and v1 action contexts?

repository/src/main/java/org/alfresco/repo/action/access/ActionAccessRestrictionAbstractBase.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public abstract class ActionAccessRestrictionAbstractBase implements ActionAccessRestriction {
+
+    private static final List<String> CONTROLLED_ACTION_ACCESS_CONTEXT =
```
pzhyland, line 44: `Set` seems to be a more suitable collection type. But in case of a small number of elements it doesn't matter here.

repository/src/main/java/org/alfresco/repo/action/access/ActionAccessRestrictionAbstractBase.java
pzhyland: +1

mstrankowski: I need to talk to Piotr about it, there is plenty of possible customisation to be done through config, I just don't know if it's needed

repository/src/main/java/org/alfresco/repo/action/access/AdminActionAccessRestriction.java
mstrankowski: fixed

remote-api/src/main/java/org/alfresco/repo/web/scripts/rule/ActionQueuePost.java
mstrankowski: fixed

repository/src/main/java/org/alfresco/repo/action/executer/ActionExecuterAbstractBase.java
```diff
@@ -79,10 +85,14 @@ public abstract class ActionExecuterAbstractBase extends ParameterizedItemAbstra
      */
     public void init()
     {
-        if (this.publicAction == true)
+        if (this.publicAction)
         {
             this.runtimeActionService.registerActionExecuter(this);
         }
+
+        if (!this.actionAccessRestrictions.isEmpty()) {
```
pzhyland, line 93: Does it make sense to replace this registration by obtaining an executer bean from application context in the `ActionServiceImpl` ?

repository/src/main/resources/alfresco/action-services-context.xml
```diff
@@ -768,5 +769,22 @@
             <ref bean="TemplateService"/>
         </property>
     </bean>
+
+    <!-- Action Restriction Beans -->
+
+    <bean id="actionAccessRestrictionBase" abstract="true" class="org.alfresco.repo.action.access.ActionAccessRestrictionAbstractBase">
+        <property name="nodeService">
+            <ref bean="NodeService"/>
```
pzhyland, line 777: Shouldn't we use a lower case `nodeService` here? Right now the only restriction is for admin users so it will work but if we think of it as a generic framework then I think security interceptors on the `NodeService` might prevent us from the real check.

pzhyland (approved):
Looks really great. It shows you hunger for writing a code :) Just added some comments/ideas to consider.

repository/src/main/java/org/alfresco/repo/action/access/ActionAccessRestrictionAbstractBase.java
pzhyland: What we have is enough. It is possible to disable the restriction check in all contexts or in a specific one.

mstrankowski: Removing TODO then

repository/src/main/java/org/alfresco/repo/action/access/ActionAccessRestrictionAbstractBase.java
mstrankowski: Switched to Set

repository/src/main/java/org/alfresco/service/cmr/action/ActionService.java
```diff
@@ -155,7 +155,14 @@
...
     @Auditable()
    CompositeActionCondition createCompositeActionCondition();
-    
+
+    /**
+     * Verify users access to an action with restrictions
+     *
+     * @param action
+     */
+    void verifyActionAccessRestrictions(Action action);
```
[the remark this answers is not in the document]
mstrankowski: Moved

repository/src/main/java/org/alfresco/repo/action/executer/ActionExecuterAbstractBase.java
mstrankowski: Changed, we'll end up with Map being filled as requests are made to different ActionExecuter instances as opposed to only access restricted ones, but they aren't extensive enough to pose an issue.

repository/src/main/resources/alfresco/action-services-context.xml
mstrankowski: I don't know about it, the upper case NodeService is fine for now and should be fine for the current uses. If we/someone else meets a problem here, there's no issue with it being replaced to lowercase nodeService. The other way around might be an issue though (as we've had experience with during action security analysis).

I could theoretically set it as "nodeService" and make the variable private in ActionAccessRestrictionAbstractBase

pzhyland: I don't know if it makes sense bu the scenario I had in mind is as follows:
1. As a regular user I change some node I have access to.
2. Some rule matches my node and an action is triggered.
3. I don't have access to the Action/Rule node created by admin.
4. It fails on checking `isActionCausedByRule` even without reaching the `innerVerifyAccessRestriction`

mstrankowski: I'll try to execute a scenario locally
Admin user creates a folder in his own space with rule of secured action,
Admin user creates a folder in shared location with rule of move action,
Regular user uploads file in shared location.

Theoretically that, if possible, could trigger the scenario. I don't know ACS that well though.

mstrankowski: It seems it has never been considered as a use case to be honest. I've tried to set up the mentioned scenario and didn't even finish triggering Copy action before getting "Access Denied.  You do not have the appropriate permissions to perform this operation.". Since I didn't get to finish copy action by not having access to the folder specified in a rule by admin, then I can't trigger other actions there regardless of the security interceptor being present or not on a restricted access action.

mstrankowski: I've tested same scenario when user has read access to both folder nodes, seems to work properly in that case.

repository/src/main/java/org/alfresco/repo/action/executer/ActionExecuter.java
```diff
@@ -76,6 +76,13 @@
      * @return  the action definition
      */
     ActionDefinition getActionDefinition();
+
+    /**
+     * Verify action access restrictions
+     *
+     * @param action
+     */
+    void verifyActionAccessRestrictions(Action action);
```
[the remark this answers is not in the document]
mstrankowski: Provided

repository/src/main/resources/alfresco/action-services-context.xml
pzhyland: If it's not a case then just leave it as it is. As you wrote we can always change it in the future.

SaraAspery (approved):
Should probably update the remaining license headers. LGTM!

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
```diff
@@ -1,28 +1,28 @@
-/*
```
mstrankowski, line 1: No changes in this file in the end.

PR #1193, opened 2022-06-30
ACS-3198: Folder rules v1 REST API - List rules for node

krdabrowski:
Get folder rules API: `GET /nodes/{folderNodeId}/rulesets/{ruleSetId}/rules` implementation proposal.


remote-api/src/main/java/org/alfresco/rest/api/Rules.java
```diff
@@ -0,0 +1,48 @@
...
+import org.alfresco.rest.api.model.Rule;
+import org.alfresco.rest.framework.resource.parameters.CollectionWithPagingInfo;
+import org.alfresco.rest.framework.resource.parameters.Paging;
+
+/**
+ * Folder node rules API.
+ *
+ * @author krdabrowski
+ */
+public interface Rules
```
montgolfiere, line 38: Add @Experimental annotation for now. Similar for other new interfaces & classes ?

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPages.java
```diff
@@ -0,0 +1,76 @@
...
+import org.alfresco.query.PagingRequest;
+import org.alfresco.query.PagingResults;
+import org.alfresco.rest.api.impl.Util;
+import org.alfresco.rest.api.search.context.SearchContext;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+public class ListPages<T> extends CollectionWithPagingInfo<T>
```
montgolfiere, line 39: (Why) do we need to add this subclass to the REST framework ?

Is this for in-memory paging (if underlying service does not support paging - with assumption that items are listed in a repeatable order over multiple paged calls) ?

If so, I wonder if there are any other existing implemented areas that do something similar that might benefit ?

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public CollectionWithPagingInfo<Rule> getRules(final String nodeId, final Paging paging)
+    {
+        final NodeRef nodeRef = nodes.validateNode(nodeId);
+
+        final Set<QName> folders = new HashSet<>(List.of(ContentModel.TYPE_FOLDER));
+        if (!nodes.nodeMatches(nodeRef, folders, null)) {
+            throw new InvalidArgumentException("NodeId of a folder is expected!");
+        }
```
montgolfiere, line 60: Does is account for sub-types of cm:folder (it might, I haven't checked yet) ?

remote-api/src/main/java/org/alfresco/rest/api/Rules.java
krdabrowski: Ok, I will add annotations everywhere where it's needed.

krdabrowski: Annotations added

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPages.java
montgolfiere: Thanks. 

Do we have a list of "everyone" (eg. code links) that we could add to your comment (for reference) ?

Maybe also log as tech debt in Jira (ie. option to go back and rationalise others) to use your generic approach ?

montgolfiere (commented):
LGTM

Before we merge, we should decide along with other tech leads (cc @tpage-alfresco):

1. Minor: maybe add a few more key fields, in particular, maybe description and enabled fag. I think these two fields are useful for at least the "linked rules" use-case, ie. UI picker to select a folder with existing rules to link to.

2. Major: maybe wait for API design decision as to whether we will change this endpoint to be one more level below, eg. GET /nodes/{folderNodeId}/rulesets/-default-/rules

Otherwise, we merge as experimental with the knowledge that we will update & maybe refactor soon with further PRs. 

Either, we should also ensure we keep in-sync with proposed API spec (and vice-versa).

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPages.java
krdabrowski: "Everyone" is not a proper wording. A lot of solutions (better sentence) handles the paging in its custom way. I don't have any list, but it's possible to do some research. Some of the solutions are old and exists there for long time.

tpage-alfresco:
> For #2, I see @tpage-alfresco has now added extra level to the [PR](https://github.com/Alfresco/rest-api-explorer/pull/161) for the API spec.
> 
> It probably makes sense to change now so that we don't get too far out-of-sync, especially as we add more endpoints (eg. to get rule, create rule, update rule and delete rule) and start initial client dev work.

I'm still not sure which is best tbh.  I think we need to work on the Word doc a bit more to get all the interactions ironed out (including potential future ones). This should give us a better idea of what the endpoints should be.

montgolfiere (approved):
LGTM - added a minor comment about Auditable (also trivial / subjective one about code).

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Experimental
+public class RulesImpl implements Rules
+{
+
+    private static final String DEFAULT_RULE_SET_ID = "-default-";
```
tpage-alfresco, line 50: At some point I think we'll want a Ruleset object and then this might want to move there.  For now this is great.

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public CollectionWithPagingInfo<Rule> getRules(final String folderNodeId, final String ruleSetId, final Paging paging)
+    {
+        final NodeRef folderNodeRef = validateNode(folderNodeId, ContentModel.TYPE_FOLDER);
+
+        if (isNotDefaultId(ruleSetId)) {
+            final NodeRef ruleSetNodeRef = validateNode(ruleSetId, ContentModel.TYPE_SYSTEM_FOLDER);
+
+            if (!ruleService.isRuleSetAssociatedWithFolder(ruleSetNodeRef, folderNodeRef)) {
+                throw new InvalidArgumentException("Rule set is not associated with folder node!");
```
tpage-alfresco, line 65: Does this end up as a 404 error?  I would assume we want 404 for this error anyway.

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        final List<org.alfresco.service.cmr.rule.Rule> rulesModels = ruleService.getRules(folderNodeRef);
```
tpage-alfresco, line 69: Maybe we should inline this variable too?  We don't use the rulesModels variable except to immediately stream it.

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    {
+        this.nodes = nodes;
+    }
+
+    public void setRuleService(RuleService ruleService)
+    {
+        this.ruleService = ruleService;
+    }
+
+    private NodeRef validateNode(final String nodeId, final QName namespaceType)
```
tpage-alfresco, line 85: Please could we have some javadoc to say what counts as validation for a node?

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelation.java
```diff
@@ -0,0 +1,40 @@
...
+import org.alfresco.rest.framework.resource.RelationshipResource;
+import org.alfresco.service.Experimental;
+
+/**
+ * Folder node rule sets.
+ *
+ */
+@Experimental
+@RelationshipResource(name = "rulesets", entityResource = NodesEntityResource.class, title = "Folder node rule sets")
+public class NodeRuleSetsRelation
```
tpage-alfresco, line 38: Ugh! I thought we used the term "ruleset", but Googling suggests we call it a "rule set."  I think we should change the URL too.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPage.java
```diff
@@ -0,0 +1,79 @@
...
+@Experimental
+public interface ListPage<E> extends List<E>, PagingResults<E>, SerializablePagedCollection<E>
+{
+
+    default List<E> getPage()
+    {
+        return this;
+    }
+
+    default Pair<Integer, Integer> getTotalResultCount()
```
tpage-alfresco, line 52: I don't understand why this returns a pair.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPage.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param list - the list that needs to be paged
+     * @param paging - paging request info
+     * @return list page in {@link CollectionWithPagingInfo}
+     * @param <E> - list element type
+     */
+    static <E> CollectionWithPagingInfo<E> of(final List<? extends E> list, final Paging paging)
+    {
+        if (list == null)
+        {
+            return CollectionWithPagingInfo.asPaged(paging, null);
```
tpage-alfresco, line 74: I don't understand what it means to create a ListPage from no list.

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
```diff
@@ -329,4 +329,13 @@
...
     public List<NodeRef> getLinkedFromRuleNodes(NodeRef nodeRef);
+
+    /**
+     * Check if rule set's associated parent is equal to folder node.
+     *
+     * @param ruleSetNodeRef - node reference of a rule set
+     * @param folderNodeRef - node reference of a folder
+     * @return true if rule set is associated with folder
+     */
+    boolean isRuleSetAssociatedWithFolder(NodeRef ruleSetNodeRef, NodeRef folderNodeRef);
```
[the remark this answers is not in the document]
krdabrowski: Fixed

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
krdabrowski: Good point! I'll fix it.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelation.java
krdabrowski: I'll fix it.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/ListPage.java
krdabrowski: This is a contract method from `interface PagingResults<R>`. Can't really do much about it.

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    {
+        final NodeRef nodeRef = nodes.validateNode(nodeId);
+
+        final Set<QName> expectedTypes = Set.of(namespaceType);
+        if (!nodes.nodeMatches(nodeRef, expectedTypes, null)) {
+            final String expectedType = namespaceType.getLocalName().replace("systemfolder", "rule set");
```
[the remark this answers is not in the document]
krdabrowski: On it.

PR #1194, opened 2022-07-01
Feature/acs 3163 secure import action against non recursive zip bomb

SaraAspery:


pzhyland (changes requested):

repository/src/main/java/org/alfresco/repo/action/executer/ImporterActionExecuter.java
```diff
@@ -405,6 +438,82 @@ public static void extractFile(ZipFile archive, String extractDir)
...
     }
 
+    /**
+     * Extract the file and folder structure of a ZIP file into the specified directory using a progress tracker
+     *
+     * @param archive       The ZIP archive to extract
+     * @param extractDir    The directory to extract into
+     * @param tracker       The extraction progress tracker to check against during the extraction process
+     */
+    private static void extractFile(ZipFile archive, String extractDir, ExtractionProgressTracker tracker)
```
pzhyland, line 448: This code is mostly duplicated from the method above. Could you unify it somehow. Maybe just calling this method from the above one but with a progress tracker which does nothing?

pzhyland (changes requested):

repository/src/main/java/org/alfresco/repo/action/executer/ImporterActionExecuter.java
SaraAspery: done

pzhyland: The more idiomatic way would be to call it in a following way `extractFile(archive, extractDir, ExtractionProgressTracker.NONE);` The `NONE` filed should act as a [Null Object Pattern](https://en.wikipedia.org/wiki/Null_object_pattern)

pzhyland (changes requested):

pzhyland (approved):
Looks great.

PR #1204, opened 2022-07-08
Feature/acs 3160 replace reload4j with log4jv2

aonych:
The current state of replacing reload4j with log4jv2. Do not merge!


damianujma:
I think we need to migrate all log4j.properties files under the `packaging/tests` directory.

damianujma:
Is there any reason to remain the old `log4j.properties` file?

dsibilio (approved):
LGTM. Assuming that internal releases (`*-DEV-LOG4J2` artifacts) will be eventually removed from the repo and are only here to make the build pass and propagate downstream first.

damianujma (approved):

mikolajbrzezinski (approved):

PR #1208, opened 2022-07-12
Fix/mnt 22293 limit on declare version as record

brijmohan1:


repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
```diff
@@ -3780,22 +3780,35 @@ public void done()
...
+                    // Remember exception we want to throw and exit. If we throw within here, it will be wrapped by IBatis
+                    toThrow = new CyclicChildRelationshipException(
+                            "Child Association Cycle detected hitting nodes: " + nodeIds,
+                            childAssociationRef);
+                    return false;
+                }
             }
-            cycleCheck(nodeId);
-            nodeIds.remove(nodeId);
+            // cycleCheck(nodeId);
```
suneet-gupta, line 3810: we shd remove commented code

suneet-gupta (changes requested):
Please do not touch original formatting or add/remove spaces if you are not making any changes at that place.

kavitshah-gl (approved):
This seems fine.

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -953,7 +953,47 @@
...
+               assoc.assoc_index         as assoc_index
+        from alf_child_assoc assoc
+                 join childAssosList on (assoc.parent_node_id = childAssosList.childNodeId)
+                 join alf_node parentNode on (parentNode.id = assoc.parent_node_id)
+                 join alf_store parentStore on (parentStore.id = parentNode.store_id)
+                 join alf_node childNode on (childNode.id = assoc.child_node_id)
+                 left join alf_store childStore on (childStore.id = childNode.store_id)
+            )
+        select *
+        from childAssosList
```
tpage-alfresco, line 994: Typo - "Assos"

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -953,7 +953,47 @@
...
+        <include refid="alfresco.node.select_ChildAssocsOfParent_Snippet"></include>
+        <include refid="alfresco.node.select_ChildAssocsOfParentRecursiveQuery"></include>
+    </select>
+    <sql id="select_ChildAssocsOfParentWithRecursive">
+        WITH RECURSIVE childAssosList AS (
+    </sql>
+    <sql id="select_ChildAssocsOfParentRecursiveQuery">
+        UNION ALL
+
+        select assoc.id                  as id,
```
tpage-alfresco, line 968: I'm concerned about what this returns. If we're adding a node to the root of the repository then does it return the whole repository?  What if we move a folder near the root of the repository to a new location?

tpage-alfresco:
I think the issue in general is not something that can sensibly be solved. I think to check for cycles in the general case then we have to traverse almost every parent-child relationship in the repository.

I noticed this comment:
```
     * TODO: When is it cheaper to go up and when is it cheaper to go down?
     *       Look at using direct queries to pass through layers both up and down.
```

For the problem we're trying to solve then I think we know there's no cycle because the node has no children?  If this is the case then we could solve the problem by changing the logic to do:
```
If the child has no children
  // No cycle
else
  // Perform existing check
```

This would improve performance for the majority of cases (creating a new child) and make it slightly worse when moving a folder or adding a secondary parent association to a folder.

jottley:
The related JIRA was closed as Won't Do.  Closing the PR

PR #1213, opened 2022-07-14
ACS-3170 Implement Tests For Mail Actions

goleksyalfresco:


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/access/AccessRestrictionUtil.java
```diff
@@ -0,0 +1,102 @@
+package org.alfresco.rest.actions.access;
```
mstrankowski, line 1: No idea if copyright notice here are needed, don't seem required and none of other tests have them (only 2 files, perhaps by accident)

PR #1216, opened 2022-07-14
ACS-3225 REST API to create one or more rules.

tpage-alfresco:


remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
```diff
@@ -39,14 +39,16 @@
...
 
 import javax.servlet.http.HttpServletResponse;
+import java.util.List;
 
 /**
  * Folder node's rules.
  *
  */
 @Experimental
 @RelationshipResource(name = "rules", entityResource = NodeRuleSetsRelation.class, title = "Folder node rules")
```
tpage-alfresco, line 49: Missing Javadoc and @WebApiDescription.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
```diff
@@ -103,6 +105,14 @@ public Rule readById(String folderNodeId, String ruleSetId, Parameters parameter
         return rules.getRuleById(folderNodeId, ruleSetId, ruleId);
     }
 
+    @Override
+    public List<Rule> create(String folderNodeId, List<Rule> rules, Parameters parameters)
```
mpichura, line 109: To avoid confusion, maybe `List<Rule> rules` could be renamed to `List<Rule> ruleList`?
That way we could also get rid of `this` keyword in line 113.

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
```diff
@@ -281,6 +284,115 @@ public void testGetRuleByIdForNotAssociatedRuleToRuleSet()
...
+
+        then(ruleServiceMock).should().isRuleSetAssociatedWithFolder(ruleSetNodeRef, folderNodeRef);
+        then(ruleServiceMock).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    public void testSaveRules_emptyRuleList()
+    {
+        given(ruleServiceMock.isRuleSetAssociatedWithFolder(any(), any())).willReturn(true);
+        List<Rule> ruleList = List.of();
```
mpichura, line 354: Again - only opinion - for me `Collections.emptyList()` is more readable here but functionally it doesn't make too much difference anymore.

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
```diff
@@ -235,8 +235,8 @@
      * @param rule Rule
      */
     @Auditable(parameters = {"nodeRef", "rule"})
-    public void saveRule(NodeRef nodeRef, Rule rule);
-    
+    public Rule saveRule(NodeRef nodeRef, Rule rule);
```
mpichura, line 238: I don't think we should change this existing (almost sine beginning of the world) public contract. 
Maybe adding a new method which would do the trick here?

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+ *  GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.rest.api.model.rules;
+
+/** Builder class for {@link Rule}. */
+public class RuleBuilder
```
krdabrowski, line 29: The builder could be also placed as a static inner class in the `Rule`. Then it could be initiated within factory method builder(), eg. `Rule.builder().id().name().create()`. In fact this is a matter of preferences, so I'm also ok with this approach.

remote-api/src/main/java/org/alfresco/rest/api/Rules.java
```diff
@@ -57,4 +59,14 @@
...
+
+    /**
+     * Create new rules (and potentially a rule set if "_default_" is supplied).
+     *
+     * @param folderNodeId The node id of a folder.
+     * @param ruleSetId The id of a rule set (or "_default_" to use/create the default rule set for the folder).
+     * @param rule The definition of the rule.
+     * @return The newly created rules.
+     */
+    List<Rule> saveRules(String folderNodeId, String ruleSetId, List<Rule> rule);
```
montgolfiere, line 71: Minor: rename "saveRules" as "createRules" ?

remote-api/src/main/java/org/alfresco/rest/api/Rules.java
```diff
@@ -57,4 +59,14 @@
...
      */
     Rule getRuleById(String folderNodeId, String ruleSetId, String ruleId);
+
+    /**
+     * Create new rules (and potentially a rule set if "_default_" is supplied).
+     *
+     * @param folderNodeId The node id of a folder.
+     * @param ruleSetId The id of a rule set (or "_default_" to use/create the default rule set for the folder).
+     * @param rule The definition of the rule.
+     * @return The newly created rules.
```
montgolfiere, line 69: Does the returned list also include existing rules (if any) or only the rules being created, even if some others still exist ?

Does this API return an error if the folder already links to another rule-set ? (EDIT: I can see "checkForLinkedRules" lower-down in RuleServiceImpl)

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
montgolfiere: Yes, new method and deprecate old. Given that RuleService is part of "cmr" package. We could also check whether explicitly marked as @PublicApi (embedded Java level) or not.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        this.name = name;
+        return this;
+    }
+
+    public Rule createRule()
+    {
+        Rule rule = new Rule();
+        rule.setId(id);
+        rule.setName(name);
```
montgolfiere, line 50: More to follow, I assume ? 

- enabled / disabled ?
- description
- apply to sub-folders (or not)
- other properties (background / foreground, error script for background) ...
- triggers ("when"), conditions, actions, ...

montgolfiere (commented):
Added some minor initial comments.

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
tpage-alfresco: It's not part of the public API, so I felt it was fair game.  I can add a new method instead if that's preferred.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleBuilder.java
tpage-alfresco: Right - these will follow @krdabrowski 's corresponding work on the GET API.

remote-api/src/main/java/org/alfresco/rest/api/Rules.java
tpage-alfresco: It only returns the newly created rules. I assume this is consistent with other APIs (although I didn't check tbh) - e.g. creating users you wouldn't want all the users in the system returned.

remote-api/src/main/java/org/alfresco/rest/api/Rules.java
tpage-alfresco: Yes - good suggestion. This will now match with the other APIs classes.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
```diff
@@ -45,13 +44,28 @@ public static Rule from(final org.alfresco.service.cmr.rule.Rule ruleModel) {
...
         return rule;
     }
 
+    /**
+     * Convert the REST model object to the equivalent service POJO.
+     *
+     * @param nodes The nodes API.
+     * @return The rule service POJO.
+     */
+    public org.alfresco.service.cmr.rule.Rule toServiceModel(Nodes nodes)
```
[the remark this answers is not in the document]
tpage-alfresco: It's a bit awkward to do this, because then whenever we want to call `rule.toServiceModel` then we have to do:
```
NodeRef ruleNode = nodes.validateOrLookupNode(rule.getId(), null);
rule.toServiceModel(ruleNode);
```
compared with:
```
rule.toServiceModel(nodes);
```
I agree that the current logic is twisted though - I just couldn't find a nicer way than this.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleBuilder.java
tpage-alfresco: I spent a while trying to get IntelliJ to do this for me, but in the end gave up and generated the builder as a new class.  I'll move it to a static inner class, even though I hate inner classes for almost every purpose then I think this is neater (it's basically a constructor).

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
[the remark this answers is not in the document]
mpichura: As said before - I don't mind when it stays as is.

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
tpage-alfresco: Actually I'm not going to bother about this. I'm only changing the return from `void` so:
* Anyone calling the method won't notice a difference
* Anyone extending/implementing the interface will need to update their code regardless of the approach.

tpage-alfresco: > If it stays and a new method is added, you don't need to do anything (however, you may want to).
In fact it's almost certainly true that the only case you need to look at this new method is if you've implemented the existing `saveRules` method.  See as an example the AGS implementation - if I hadn't seen the compilation error then I wouldn't have realised I needed to update the implementation.

I think this is an argument for _not_ keeping the existing method.

PR #1224, opened 2022-07-15
ACS-3229: Rules v1 REST API - Get rule definition - rule mapping

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-3229

Adding POJOs mapping.


remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -63,10 +63,12 @@ public class RulesImpl implements Rules
     public CollectionWithPagingInfo<Rule> getRules(final String folderNodeId, final String ruleSetId, final Paging paging)
     {
         final NodeRef folderNodeRef = validateFolderNode(folderNodeId);
-        validateRuleSetNode(ruleSetId, folderNodeRef);
+        final NodeRef ruleSetNodeRef = validateRuleSetNode(ruleSetId, folderNodeRef);
 
+        final boolean isShared = ruleService.isRuleSetShared(ruleSetNodeRef);
         final List<Rule> rules = ruleService.getRules(folderNodeRef).stream()
             .map(Rule::from)
+            .peek(r -> r.setShared(isShared))
```
tpage-alfresco, line 71: I think we call this every time we call `Rule.from`. Maybe it would make sense to refactor it - we're going to have more methods that return rules as we implement more of the API.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Action.java
```diff
@@ -0,0 +1,131 @@
...
+    private Map<String, Serializable> params;
+
+    /**
+     * Converts service POJO action to REST model action.
+     *
+     * @param actionModel - {@link org.alfresco.service.cmr.action.Action} service POJO
+     * @return {@link Action} REST model
+     */
+    public static Action from(final org.alfresco.service.cmr.action.Action actionModel) {
+        if (actionModel == null) {
```
tpage-alfresco, line 49: [Minor] Formatting of braces

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Action.java
```diff
@@ -0,0 +1,131 @@
...
+            return null;
```
tpage-alfresco, line 50: I guess this isn't done yet?

tpage-alfresco (commented):
This PR is pretty big. Would it be possible to split it into three PRs:
* Actions
* Conditions
* Other fields

We planned to do this (see https://alfresco.atlassian.net/browse/ACS-3207 ) but the subtasks didn't get created like it.

PR #1232, opened 2022-07-20
ACS-3220 Folder rules delete rule v1 API

mpichura:
This PR covers following features: 
- endpoint for DELETE single rule
  - DELETE /nodes/{folderNodeId}/rule-sets/{ruleSetId}/rules/{ruleId}
- business logic to remove the rule
- unit tests covering above

Not covered - bulk delete rules under rule set or folder.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
@@ -46,7 +46,8 @@ import javax.servlet.http.HttpServletResponse;
  */
 @Experimental
 @RelationshipResource(name = "rules", entityResource = NodeRuleSetsRelation.class, title = "Folder node rules")
-public class NodeRulesRelation implements RelationshipResourceAction.Read<Rule>, RelationshipResourceAction.ReadById<Rule>, InitializingBean
+public class NodeRulesRelation implements RelationshipResourceAction.Read<Rule>, RelationshipResourceAction.ReadById<Rule>,
+        RelationshipResourceAction.Delete, InitializingBean
 {
 
     private Rules rules;
@@ -106,5 +107,27 @@ public class NodeRulesRelation implements RelationshipResourceAction.Read<Rule>,
     public void setRules(Rules rules)
     {
         this.rules = rules;
+    }
+
+    /**
+     * Delete single folder rule for given node's, rule set's and rule's IDs.
+     *
+     * - DELETE /nodes/{folderNodeId}/rule-sets/{ruleSetId}/rules/{ruleId}
+     *
+     * @param folderNodeId - entity resource context for this relationship
+     * @param ruleSetId - rule set node ID (associated with folder node)
+     * @param parameters - Should not be null. Should contain at least ruleId (relationship2Id)
+     * @throws RelationshipResourceNotFoundException in case resource was not found
+     */
+    @WebApiDescription(
+            title="Delete folder node rule",
+            description = "Deletes a single rule definition for given node's, rule set's and rule's IDs",
+            successStatus = HttpServletResponse.SC_OK
+    )
+    @Override
+    public void delete(String folderNodeId, String ruleSetId, Parameters parameters)
+    {
+        final String ruleId = parameters.getRelationship2Id();
+        rules.deleteRuleById(folderNodeId, ruleSetId, ruleId);
     }
 }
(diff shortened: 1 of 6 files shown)


remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
```diff
@@ -130,4 +130,26 @@ public void setRules(Rules rules)
...
+     *
+     * @param folderNodeId - entity resource context for this relationship
+     * @param ruleSetId - rule set node ID (associated with folder node)
+     * @param parameters - Should not be null. Should contain at least ruleId (relationship2Id)
+     * @throws RelationshipResourceNotFoundException in case resource was not found
+     */
+    @WebApiDescription(
+            title="Delete folder node rule",
+            description = "Deletes a single rule definition for given node's, rule set's and rule's IDs",
+            successStatus = HttpServletResponse.SC_OK
```
tpage-alfresco, line 147: This should be 204 for a delete.

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
```diff
@@ -394,10 +395,136 @@ public void testSaveRules_createMultipleRules()
         assertThat(actual).isEqualTo(expected);
     }
 
+    @Test
+    public void testDeleteRuleById() {
+        given(nodesMock.validateNode(eq(RULE_SET_ID))).willReturn(ruleSetNodeRef);
+        given(nodesMock.validateNode(eq(RULE_ID))).willReturn(ruleNodeRef);
```
tpage-alfresco, line 401: FWIW we don't need the `eq` in these two lines.

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
```diff
@@ -394,10 +395,136 @@ public void testSaveRules_createMultipleRules()
...
+        rules.deleteRuleById(FOLDER_NODE_ID, RULE_SET_ID, RULE_ID);
+
+        then(nodesMock).should().validateOrLookupNode(eq(FOLDER_NODE_ID), isNull());
+        then(nodesMock).should().validateNode(eq(RULE_SET_ID));
+        then(nodesMock).should().validateNode(eq(RULE_ID));
+        then(nodesMock).should().nodeMatches(eq(folderNodeRef), any(), isNull());
+        then(nodesMock).should().nodeMatches(eq(ruleSetNodeRef), any(), isNull());
+        then(nodesMock).should().nodeMatches(eq(ruleNodeRef), any(), isNull());
+        then(nodesMock).shouldHaveNoMoreInteractions();
+        then(permissionServiceMock).should().hasReadPermission(eq(folderNodeRef));
```
montgolfiere, line 417: I assume you need more than just folder "Read" permission to delete a rule ? 

Maybe  folder"Write" permission or something more specific ? As an aside, I also see references to "ChangePermssion" later in the PR.

montgolfiere (approved):
LGTM

I've added one comment to mock unit test about specific permission required (to delete a rule) otherwise presumably causing 403 (via AccessDeniedException / PermissionDeniedException) ?

tpage-alfresco (commented):
LGTM - Just the 200 -> 204 issue.  (I assume this manifests itself as a documentation issue rather than a code issue)

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
tpage-alfresco: This must be an issue with the POST API too. I've only tested a user who has access so far.

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
mpichura: Thanks for spotting that. I'm cleaning up all `eq(...)` uses in that class as there are way too many not needed.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRulesRelation.java
mpichura: Thanks - changing that.

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
mpichura: I tested DELETE API when authenticated as a user with read-only permissions and it is not allowed to remove the rule in that case. `RuleServiceImpl` remove rule (existing) code is checking the permissions (`permissionService.hasPermission(nodeRef, PermissionService.CHANGE_PERMISSIONS) == AccessStatus.ALLOWED`) as Jan noticed.
It might be better to check that in `RulesImpl` at the very beginning (considering efficiency of double permission check when permissions are okay). WDYT?

tpage-alfresco (approved):
LGTM!

remote-api/src/test/java/org/alfresco/rest/api/impl/RulesImplTest.java
tpage-alfresco: I think it sounds like a good candidate for a separate PR (and maybe a separate ticket).

PR #1240, opened 2022-07-22
ACS-3227 Ensure the rule name is set. [tas]

tpage-alfresco:


repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
montgolfiere, line 5: Trivial: 2022 :-)

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
```diff
@@ -783,7 +783,9 @@ public Rule saveRule(NodeRef nodeRef, Rule rule)
             }
 
             // Update the properties of the rule
-            this.nodeService.setProperty(ruleNodeRef, ContentModel.PROP_TITLE, rule.getTitle());
+            String title = rule.getTitle();
+            ParameterCheck.mandatoryString("Rule name", title);
```
montgolfiere, line 787: Does V0 API check for rule name or allow empty ?  I assume there are no custom integrations (eg. using V0) that don't supply a name. 

We may need to (eventually) remember to list this change in ACS 7.3.0 release notes ?

montgolfiere (approved):
Added trivial/minor comments but overall LGTM.

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
tpage-alfresco: Yes - weirdly there is validation in both the v0 webscript and in Share, but there was nothing in the service layer.

tpage-alfresco:
Tested this against ent-repo, ent-share and acs-packaging and all the errors looked intermittent.

PR #1244, opened 2022-07-25
Rules v1 REST API - Get rule definition - mapping of "other fields"

krdabrowski:
PR contains mapping of rule's so called: "other fields".

In addition to rule id and name, also:

- description
- enabled (whether rule is enabled or disabled)
- cascade (whether rule applies to children / sub-folders or not)
- asynchronous (whether rule is executed in background / async or foreground / sync)
- shared (whether rule / ruleset is being shared, ie. other folders are linking to this rule / rule set, ie. linked rule set)
- errorScript (script name to run in case of an error occurrence)


tpage-alfresco:
The PR build test failures are due to the naive automated merge of this branch into master by GitHub.  The main issue seems to be when creating rules we get a 500 error. This is caused by calling `Rule.from` in which we hit an access issue.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleSet.java
```diff
@@ -37,10 +39,9 @@
 
     public static RuleSet of(String id)
     {
-        final RuleSet ruleSet = new RuleSet();
-        ruleSet.id = id;
-
-        return ruleSet;
+        return builder()
```
mpichura, line 42: Isn't a builder in this case (single field) an overkill? Or maybe other fields are coming in next PRs?

krdabrowski: This is meant for other fields coming in next PRs.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
```diff
@@ -26,19 +26,21 @@
...
 import org.alfresco.repo.action.ActionImpl;
 import org.alfresco.repo.action.executer.SetPropertyValueActionExecuter;
 import org.alfresco.rest.api.Nodes;
+import org.alfresco.repo.action.executer.ScriptActionExecuter;
 import org.alfresco.rest.framework.resource.UniqueId;
 import org.alfresco.service.Experimental;
-import org.alfresco.service.cmr.action.Action;
+import org.alfresco.service.cmr.action.CompositeAction;
+
+import java.io.Serializable;
```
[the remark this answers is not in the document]
krdabrowski: So we might agree now on some approach, I will adjust my code styling. How should it look like, e.g.
import static*
import java*
import all others*
this order is ok?

CC @tpage-alfresco @gevagel @KikoUK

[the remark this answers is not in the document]
krdabrowski: Fixed

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
```diff
@@ -79,23 +80,25 @@ public Rule getRuleById(final String folderNodeId, final String ruleSetId, final
...
         final NodeRef folderNodeRef = validateFolderNode(folderNodeId);
         // Don't validate the ruleset node if -default- is passed since we may need to create it.
+        NodeRef ruleSetNodeRef = null;
         if (RuleSet.isNotDefaultId(ruleSetId))
         {
-            validateRuleSetNode(ruleSetId, folderNodeRef);
+            ruleSetNodeRef = validateRuleSetNode(ruleSetId, folderNodeRef);
         }
 
+        final boolean isShared = isRuleSetNotNullAndShared(ruleSetNodeRef);
```
tpage-alfresco, line 97: Will this work if `-default-` is passed in?  I think we'll end up always saying that the `-default-` rule set is not shared.

remote-api/src/test/java/org/alfresco/rest/api/model/rules/RuleTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Test
+    public void testFrom()
+    {
+        final org.alfresco.service.cmr.rule.Rule ruleModel = createRuleModel();
+
+        // when
+        final Rule rule = Rule.from(ruleModel, RULE_SHARED);
+
+        assertThat(rule).is(havingExpectedConstantValues());
```
tpage-alfresco, line 64: I'm not sure this is easier to read with the assertions in the private method.  I don't think you're reusing the method either.  Does this maybe help by giving more information in the case of multiple errors?

tpage-alfresco (approved):
LGTM. I really like what you've done with those import statements :100:

mpichura (approved):

remote-api/src/test/java/org/alfresco/rest/api/model/rules/RuleTest.java
krdabrowski: Fixed that by comparing actual rule with predefined expected rule.

remote-api/src/main/java/org/alfresco/rest/api/impl/RulesImpl.java
krdabrowski: Fixed it, assuming that missing rule set node will be created within `ruleService.saveRule(folderNodeRef, rule)`, but I'm not sure about that. Further fix might be required in future.

PR #1245, opened 2022-07-25
ACS-3229: Rules v1 REST API - Get rule definition - mapping of actions

krdabrowski:
Adding mapping of "actions"


remote-api/src/main/java/org/alfresco/rest/api/model/rules/Action.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public void setParams(Map<String, Serializable> params)
+    {
+        this.params = params;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "Action{" + "actionDefinitionId='" + actionDefinitionId + '\'' + ", params=" + params + '}';
```
tpage-alfresco, line 87: Is this autogenerated?  It's a bit weird!
```suggestion
        return "Action{actionDefinitionId='" + actionDefinitionId + "', params=" + params + "}";
```

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
```diff
@@ -209,16 +217,21 @@ public void setTriggers(List<RuleTrigger> triggers)
...
+    {
+        this.actions = actions;
     }
 
     @Override
     public String toString()
     {
         return "Rule{" + "id='" + id + '\'' + ", name='" + name + '\'' + ", description='" + description + '\'' + ", enabled=" + enabled + ", cascade=" + cascade
-            + ", asynchronous=" + asynchronous + ", shared=" + shared + ", errorScript='" + errorScript + '\'' + ", triggers=" + triggers + '}';
+            + ", asynchronous=" + asynchronous + ", shared=" + shared + ", errorScript='" + errorScript + '\'' + ", triggers=" + triggers + ", actions=" + actions + '}';
```
tpage-alfresco, line 234: I guess it was autogenerated because this uses the same `'\''` syntax

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Action.java
krdabrowski: Yes, this is autogenerated.

krdabrowski:
> LGTM. I was initially surprised there weren't any unit tests, but after looking through then there wasn't really much to test anyway. Maybe `Action.from` or `Rule.from` could be worth testing. Anyway, I'm happy with this as is.

Exactly that was also my assumption, that for actions there isn't much to test, but I will add same basic unit tests. Maybe in future when the class will be extended it will be useful.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
krdabrowski: The same as above - also autogenerated.

tpage-alfresco (approved):
Still looks good!

PR #1256, opened 2022-07-27
ACS-3345 e2e REST API TAS tests delete rule

mpichura:


gevagel (approved):
LGTM

krdabrowski (approved):

KikoUK (approved):

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,182 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
KikoUK, line 5: Minor: 2022

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,182 @@
...
+    @BeforeClass(alwaysRun = true)
+    public void dataPreparation()
+    {
+        STEP("Create a user, site and folder");
+        user = dataUser.createRandomTestUser();
+        site = dataSite.usingUser(user).createPublicRandomSite();
+        ruleFolder = dataContent.usingUser(user).usingSite(site).createFolder();
+
+        STEP("Create rules in the folder");
+        createdRules = Stream.of("ruleA", "ruleB")
```
KikoUK, line 76: Why are you creating this list of rules and then use only one of the rules in testing?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Delete a rule by its id.
+     */
+    @Test(groups = {TestGroup.REST_API, TestGroup.RULES, TestGroup.SANITY})
+    public void deleteSingleRuleAndGet204()
+    {
+        STEP("Attempt delete the rule ");
+        restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet().deleteRule(createdRuleA.getId());
```
KikoUK, line 94: Is it possible to get the rule set id instead of using the default one?

tpage-alfresco, line 94: I don't think this is a good idea. The test results will depend on the order of the tests.  If we create a second test that tries to delete the same rule then I assume both will pass, even though the second one didn't really delete anything.

KikoUK: I agree with @tpage-alfresco, e2es should be independent of each other and be able to run on its own.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
tpage-alfresco: Maybe we could have a test that only one rule is deleted? (i.e. check that the second rule is still there afterwards)

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: Thanks, fixed.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: Thanks for both remarks. I've refactored the methods into more independent ones and started using other rules from the list to check whether there are still rules left in a folder (as Tom suggested).

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: Thanks - fixed that - it is independent.

tpage-alfresco: I don't see how it is independent.  The rule is created in a BeforeClass?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,182 @@
...
+        final SiteModel privateSite = dataSite.usingUser(privateUser).createPrivateRandomSite();
+        final FolderModel privateFolder = dataContent.usingUser(privateUser).usingSite(privateSite).createFolder();
+        final RestRuleModel ruleModel = new RestRuleModel();
+        ruleModel.setName("Private site rule");
+        final RestRuleModel createdRule = restClient.authenticateUser(privateUser).withCoreAPI().usingNode(privateFolder).usingDefaultRuleSet()
+                .createSingleRule(ruleModel);
+
+        STEP("Create a manager in the private site");
+        final UserModel collaborator = dataUser.createRandomTestUser();
+        collaborator.setUserRole(SiteManager);
```
tpage-alfresco, line 174: I think it's fine to use either a collaborator or a site manager, but you've mixed both terms.

mpichura:
I added one general fix to this PR. 
It handles the situation when API is called for following conditions: 
- a proper `folderNodeId` is passed, however, this folder has no rules created
- a `-default-` rule set is passed
- an existing `ruleId` is passed, however, the rule is not coming from a folder that was passed as `folderNodeId`
In this case a `-default-` rule set should not exist in folder without rules and `RulesImpl::validateRuleSetNode` will throw `RelationshipResourceNotFoundException`.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,231 @@
...
+        final SiteModel privateSite = dataSite.usingUser(privateUser).createPrivateRandomSite();
+        final FolderModel privateFolder = dataContent.usingUser(privateUser).usingSite(privateSite).createFolder();
+        final RestRuleModel ruleModel = new RestRuleModel();
+        ruleModel.setName("Private site rule");
+        final RestRuleModel createdRule =
+                restClient.authenticateUser(privateUser).withCoreAPI().usingNode(privateFolder).usingDefaultRuleSet()
+                        .createSingleRule(ruleModel);
+
+        STEP("Try to delete the rule with another user");
+        restClient.authenticateUser(user).withCoreAPI().usingNode(privateFolder).usingDefaultRuleSet().deleteRule(createdRule.getId());
```
tpage-alfresco, line 192: Let's use a collaborator here and make it a requirement that collaborators can't delete private rules.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
```diff
@@ -0,0 +1,231 @@
...
+        rulesAfterDeletion.assertThat().entriesListCountIs(createdRules.size() - 1);
+        Assert.assertTrue(rulesAfterDeletion.getEntries()
+                .stream()
+                .noneMatch(r -> r.onModel().getId().equals(ruleA.getId()))
+        );
+        rulesAfterDeletion.getEntries().stream()
+                .map(rulesLeft -> rulesLeft.onModel().getId())
+                .forEach(left -> Assert.assertTrue( createdRules.stream()
+                        .map(RestRuleModel::getId)
+                        .anyMatch(left::equals)
```
tpage-alfresco, line 107: This is fine, but it would be more accurate to just check the set of ids match.

tpage-alfresco (approved):
A couple of really minor suggestions, but this looks great.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
KikoUK: Not sure if it's the same across acs, but for versioning and uploads, collaborators usually have the same privileges as contributors/managers

tpage-alfresco: I assume this comes under "Delete content created by another user" - only Managers have that permission:
image

tpage-alfresco: So it would be good to test:
* Consumer cannot create a rule
* Contributor can delete their own rule
* Collaborator cannot delete rule created by another user
* Manager can delete rule created by another user

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: I pushed the changes after the comment.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: Good point - changed that.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
mpichura: I've added a test for collaborator (forbidden to delete) and there is still is the one that covers successful delete by manager.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/DeleteRulesTests.java
KikoUK: According to @tpage-alfresco 
[Tom Page]
[12:41 PM](https://alfresco.slack.com/archives/CTWMLBG6N/p1659094890731439)
Currently you have to use -default-
[12:42](https://alfresco.slack.com/archives/CTWMLBG6N/p1659094923123509)
We're adding an API to GET rule sets at the moment though





[12:42](https://alfresco.slack.com/archives/CTWMLBG6N/p1659094923123509)
We're adding an API to GET rule sets at the moment though

tpage-alfresco: Right. So the current tests are the best we can manage (non-existent rule set id and mocking).

PR #1257, opened 2022-07-27
ACS-3214 Unit tests for update rule API.

tpage-alfresco:


KikoUK (approved):
LGTM - shall we split these tests a bit as this file is getting quite large?

gevagel (approved):

tpage-alfresco:
> LGTM - shall we split these tests a bit as this file is getting quite large?

I don't have any objections to someone investigating this.  We've still got some more few features to add here (conditions and actions) and so there will probably be more tests too.

I had a look at `RulesImpl` and I don't think we can make that much smaller. I think we've just added a relatively large amount of unit test validation for the class.

PR #1260, opened 2022-07-28
ACS-2064 DAU: Fix response content type always PDF (#754)

jottley:
* ACS-2064 ContentStore.java to require mimetype

- Updates the ContentStore interface to also require a mimetype, deprecates old method signatures
- Original interface methods maintained so code should be backwards compatible
- Implementing classes updated to reflect changes


vrmoreira (approved):
LGTM

PR #1262, opened 2022-07-29
ACS-3227: REST API TAS test for create rule with consumer role.

mpichura:
Adding a test for not enough permissions to create a rule with consumer role.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
@@ -28,6 +28,7 @@ package org.alfresco.rest.rules;
 import static java.util.stream.Collectors.toList;
 
 import static org.alfresco.utility.constants.UserRole.SiteCollaborator;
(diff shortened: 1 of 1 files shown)


tpage-alfresco (approved):
LGTM!

KikoUK:
Can a collaborator role create a rule? 
If no, I think we should add a negative test that covers that scenario too. 

I think we should also have a positive test for contributor creating a rule.

mpichura:
> Can a collaborator role create a rule? If no, I think we should add a negative test that covers that scenario too.
> 
> I think we should also have a positive test for contributor creating a rule.

I didn't work on create rule API but assume that both collaborator and contributor can create rules in proper folders. 
Shall I add these cases within this PR or we'll decide to add them in a separate one?

tpage-alfresco:
> > Can a collaborator role create a rule? If no, I think we should add a negative test that covers that scenario too.
> > I think we should also have a positive test for contributor creating a rule.
> 
> I didn't work on create rule API but assume that both collaborator and contributor can create rules in proper folders. Shall I add these cases within this PR or we'll decide to add them in a separate one?

Adding them in a separate one sounds good.  If you don't have time then just raising a task would be great too, since the test case isn't exactly urgent.

PR #1263, opened 2022-07-29
ACS-3360 GET APIs for rule sets.

tpage-alfresco:
Move RulesImpl and RuleSetsImpl into their own package.
Split out node validation into a new class.


mpichura (approved):
Overall LGTM, though I think we could rename one class (as stated in a comment)

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RulesImplTest.java
```diff
@@ -102,6 +102,13 @@ public void setUp() throws Exception
     {
         MockitoAnnotations.openMocks(this);
 
+        // Create the node validator.
+        NodeValidator validator = new NodeValidator();
+        validator.setNodes(nodesMock);
+        validator.setPermissionService(permissionServiceMock);
```
krdabrowski, line 108: Mock `permissionServiceMock` shouldn't be present here. All checks against this mock should be done in `NodeValidator` unit tests.

krdabrowski (approved):

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RulesImplTest.java
krdabrowski: I have fixed the `RulesImplTest` and moved some tests to new `NodeValidatorTest`.

PR #1266, opened 2022-08-01
ACS-3316 Fix HTML sanitisation bypass

mikolajbrzezinski:


remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
```diff
@@ -65,6 +68,16 @@ public class ContentGet extends StreamContent implements ServletContextAware
     private NamespaceService namespaceService;
     private ContentService contentService;
 
+    private Set<String> nonAttachContentTypes = Collections.emptySet();
+
+    public void setNonAttachContentTypes(String nonAttachAllowListStr)
+    {
+        if ((nonAttachAllowListStr != null) && (! nonAttachAllowListStr.isEmpty()))
```
dsibilio, line 75: There is an extra space between `!` and `nonAttachAllowListStr`.
Either way, there is a much cleaner way to inject comma-separated Strings as Lists by leveraging Spring features.
Look at https://github.com/Alfresco/alfresco-enterprise-repo/pull/819/files#diff-be731d09427ccf7e6a2346777d6fdc801b5ef77d5c9459476ded44d396fa9fdbR63 for example.
It would be good to use the latter.

dsibilio (changes requested):

remote-api/src/main/java/org/alfresco/repo/web/scripts/MimeTypeUtil.java
```diff
@@ -0,0 +1,52 @@
...
+package org.alfresco.repo.web.scripts;
+
+import org.alfresco.repo.content.MimetypeMap;
+import org.alfresco.service.cmr.repository.ContentReader;
+import org.alfresco.service.cmr.repository.MimetypeService;
+import org.springframework.extensions.webscripts.WebScriptRequest;
+
+public class MimeTypeUtil
+{
+    public static String determineMimetype(ContentReader reader, WebScriptRequest req, MimetypeService mimetypeService)
```
dsibilio, line 35: Let's add some Java Doc explaining what this method does in a generic way.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
```diff
@@ -140,6 +148,22 @@ public void execute(WebScriptRequest req, WebScriptResponse res)
...
             }
         }
+        // determine attachment and force download for specific mimetypes - see PRODSEC-5862
+        boolean attach = Boolean.valueOf(req.getParameter("a"));
+        ContentReader reader = contentService.getReader(nodeRef, propertyQName);
+        String mimetype = MimeTypeUtil.determineMimetype(reader,req,mimetypeService);
+
+        if (!attach)
+        {
+            if ((nonAttachContentTypes != null) && (!(nonAttachContentTypes.contains(mimetype))))
```
dsibilio, line 158: This conditional has not been inverted properly. If it was `nonAttachContentTypes != null && nonAttachContentTypes.contains(mimetype)`, then the negation should be `nonAttachContentTypes == null || !nonAttachContentTypes.contains(mimetype)`.

Also, there are a lot of unnecessary `(` and `)` characters. Let's get rid of them.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentInfo.java
```diff
@@ -79,18 +79,7 @@ protected void streamContentImpl(WebScriptRequest req, WebScriptResponse res,
...
-        {
-            mimetype = MimetypeMap.MIMETYPE_BINARY;
-            int extIndex = extensionPath.lastIndexOf('.');
-            if (extIndex != -1)
-            {
-                String ext = extensionPath.substring(extIndex + 1);
-                mimetype = mimetypeService.getMimetype(ext);
-            }
-        }
+        String mimetype = MimeTypeUtil.determineMimetype(reader,req,mimetypeService);
```
dsibilio, line 82: Formatting, missing spaces between arguments: `(reader, req, mimetypeService)`.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
```diff
@@ -361,18 +362,7 @@ public void streamContentImpl(WebScriptRequest req,
...
-        {
-            mimetype = MimetypeMap.MIMETYPE_BINARY;
-            int extIndex = extensionPath.lastIndexOf('.');
-            if (extIndex != -1)
-            {
-                String ext = extensionPath.substring(extIndex + 1);
-                mimetype = mimetypeService.getMimetype(ext);
-            }
-        }
+        String mimetype = MimeTypeUtil.determineMimetype(reader,req,mimetypeService);
```
dsibilio, line 365: Same formatting issues as above.

dsibilio (changes requested):

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
```diff
@@ -140,6 +148,22 @@ public void execute(WebScriptRequest req, WebScriptResponse res)
...
+            {
+                attach = true;
+            }
+            else
+            {
+                logger.warn("Ignored a=false for " + nodeRef.getId() + " since " + mimetype + " is not in the whitelist for non-attach content types");
+            }
```
dsibilio, line 165: This else block won't be needed after the if conditional is properly inverted. The code should set override `attach = true` and print the log within the same conditional block, it doesn't make sense for this statement to be here.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
mikolajbrzezinski: nonAttachContentTypes is just a check to see if there are any content types restricted for showing in browser, as it was also done this way in other places (I guess in case we ever change our mind about that). If you think its not necessary I can remove it completely.  It should not be inverted together with the other check which actually is part of the logic of the operation, as opposed to just checking if the list of acceptable mime types is empty

remote-api/src/main/resources/alfresco/web-scripts-application-context.xml
```diff
@@ -249,6 +249,7 @@
       <property name="delegate" ref="webscript.content.streamer" />
       <property name="contentService" ref="contentService" />
       <property name="repository" ref="repositoryHelper" />
+      <property name="nonAttachContentTypes" value="#{T(java.util.Arrays).asList('${content.nonAttach.mimetypes}')}" />
```
dsibilio, line 252: Please double-check the differences with the implementation shown [here](https://github.com/Alfresco/alfresco-enterprise-repo/pull/819/files#diff-be731d09427ccf7e6a2346777d6fdc801b5ef77d5c9459476ded44d396fa9fdbR63). The implementation that I linked defaults to an empty list. We don't want `content.nonAttach.mimetypes` to be unnecessarily mandatory.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
dsibilio: Clarified via quick Zoom meeting that we do need to completely negate the conditional clause.

amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/SlingshotContentGetTest.java
```diff
@@ -212,15 +213,58 @@ public void testRelativePath() throws Exception
...
     }
 
-    public NodeRef createNodeWithTextContent(NodeRef parentNode, String nodeCmName, QName nodeType, String content)
+    public void testForcedAttachment() throws Exception
+    {
+        Repository repositoryHelper = (Repository) getServer().getApplicationContext().getBean("repositoryHelper");
+        NodeRef companyHome = repositoryHelper.getCompanyHome();
+
+        NodeRef rootFolder = createNode(companyHome, "rootFolder", ContentModel.TYPE_FOLDER);
+        NodeRef testhtml = createNodeWithTextContent(rootFolder,"testhtml", ContentModel.TYPE_CONTENT, "testhtml content", MimetypeMap.MIMETYPE_HTML);
```
dsibilio, line 222: Careful about formatting, this method call and many of the following ones don't have proper spacing inbetween arguments. Just select the code that you added (e.g.: the whole unit test method) and run the formatter just on that piece of code.

dsibilio (changes requested):

dsibilio (approved):

damianujma (approved):

aonych (approved):

PR #1269, opened 2022-08-02
ACS-3290: V1 REST API endpoint for linking to ruleset 

gevagel:


remote-api/src/main/java/org/alfresco/rest/api/Rules.java
```diff
@@ -93,4 +95,10 @@
      * @param ruleId - rule ID     *
      */
     void deleteRuleById(String folderNodeId, String ruleSetId, String ruleId);
+
+    /**
+     * Link a rule set to a folder
+     *
+     */
+    RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId);
```
tpage-alfresco, line 103: I guess this would be better on the `RuleSet` interface?

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
```diff
@@ -112,6 +119,39 @@ public void deleteRuleById(String folderNodeId, String ruleSetId, String ruleId)
...
     }
 
+    @Override
+    public RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId) {
+
+        RuleSetLink ruleSetLink = new RuleSetLink();
+        final NodeRef folderNodeRef = validator.validateFolderNode(folderNodeId,false);
+        final NodeRef linkToNodeRef = validator.validateFolderNode(linkToNodeId, false);
+
+        if(nodeService.exists(folderNodeRef))
```
tpage-alfresco, line 129: What should happen if the folderNodeRef doesn't exist?  Is this already validated above?

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
```diff
@@ -112,6 +119,39 @@ public void deleteRuleById(String folderNodeId, String ruleSetId, String ruleId)
...
+            }
+
+            //The folder shouldn't have any pre-existing rules
+            if((!ruleService.hasRules(folderNodeRef)))
+            {
+                // Create the destination folder as a secondary child of the first
+                NodeRef ruleSetNodeRef = runtimeRuleService.getSavedRuleFolderAssoc(linkToNodeRef).getChildRef();
+                // The required aspect will automatically be added to the node
+                nodeService.addChild(folderNodeRef, ruleSetNodeRef, RuleModel.ASSOC_RULE_FOLDER, RuleModel.ASSOC_RULE_FOLDER);
+                ruleSetLink.setId(ruleSetNodeRef.getId());
```
tpage-alfresco, line 144: I think the logic in this method can be inverted to simplify it to:
```
Validation stuff
Stuff that throws exceptions
Create the rule set link
return ruleSetLink;
```

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
```diff
@@ -112,6 +119,39 @@ public void deleteRuleById(String folderNodeId, String ruleSetId, String ruleId)
         ruleService.removeRule(folderNodeRef, rule);
     }
 
+    @Override
+    public RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId) {
+
+        RuleSetLink ruleSetLink = new RuleSetLink();
+        final NodeRef folderNodeRef = validator.validateFolderNode(folderNodeId,false);
+        final NodeRef linkToNodeRef = validator.validateFolderNode(linkToNodeId, false);
```
tpage-alfresco, line 127: The linkToNodeId is allowed to be a folder or a rule set, but I think this code only copes when it's a folder. It's probably easiest to create a new sub task where we'll add support for linking using a rule set noderef.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
```diff
@@ -112,6 +119,39 @@ public void deleteRuleById(String folderNodeId, String ruleSetId, String ruleId)
...
+        if(nodeService.exists(folderNodeRef))
+        {
+            //The target node should have pre-existing rules to link to
+            if(!ruleService.hasRules(linkToNodeRef))
+            {
+                throw new AlfrescoRuntimeException("The target node has no rules to link.");
+            }
+
+            //The folder shouldn't have any pre-existing rules
+            if((!ruleService.hasRules(folderNodeRef)))
```
tpage-alfresco, line 138: ((((((Extra brackets))))))

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleSetLink.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.rest.api.model.rules;
+
+public class RuleSetLink {
+    private String id;
```
tpage-alfresco, line 30: It would be useful to have some information about what this id is.  (I think in this PR it's always the node ref of the linked-to-folder)

remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -888,6 +888,8 @@
         <property name="nodes" ref="Nodes" />
         <property name="validator" ref="nodeValidator"/>
         <property name="ruleService" ref="RuleService" />
+        <property name="nodeService" ref="NodeService"/>
+        <property name="runtimeRuleService" ref="ruleService"/>
```
tpage-alfresco, line 892: Is it deliberate to use the internal (lowercase) `ruleService` rather than the `RuleService`?

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RulesImplTest.java
```diff
@@ -611,4 +689,11 @@ private static List<Exception> ruleValidationExceptions()
             new RelationshipResourceNotFoundException(RULE_ID, "fake-relationship-id")
         );
     }
+
+    private void setNodeRefsForLinking()
+    {
+        given(nodeValidatorMock.validateFolderNode(eq(LINK_TO_NODE_ID), anyBoolean())).willReturn(linkToNodeRef);
+        given(nodeValidatorMock.validateFolderNode(eq(FOLDER_NODE_ID), anyBoolean())).willReturn(folderNodeRef);
```
tpage-alfresco, line 696: Could these go in the setUp method instead?  We could replace the existing `given(nodValidatorMock.validateFolderNode...` line.

tpage-alfresco: Oh - except that I suggested moving the method to `RuleSetImpl`, so all of this test code should go to `RuleSetImplTest` instead.

remote-api/src/test/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelationTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    public void shouldProperlyCreateLink()
+    {
+        RuleSetLink requestBody = new RuleSetLink();
+        RuleSetLink expectedRuleSetLink = new RuleSetLink();
+        List<RuleSetLink> expectedRuleResult = List.of(expectedRuleSetLink);
+
+
+        when(rules.linkToRuleSet(FOLDER_NODE_ID, LINK_TO_NODE_ID)).thenReturn(expectedRuleSetLink);
```
tpage-alfresco, line 67: I don't think you should be returning the expected value from a mock.

remote-api/src/test/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelationTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        requestBody.setId(LINK_TO_NODE_ID);
+        expectedRuleResult.get(0).setId("dummy-ruleset-id");
+
+        List<RuleSetLink> actual = nodeRuleSetLinksRelation.create(FOLDER_NODE_ID,List.of(requestBody), parameters);
+        Assert.assertEquals(expectedRuleResult.get(0).getId(), actual.get(0).getId());
```
tpage-alfresco, line 73: Can we assert equality of the whole list instead?  It might require writing an `equals` method (and `hashCode`), but I think it would be neater.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetLinksRelation.java
```diff
@@ -0,0 +1,74 @@
...
+import org.alfresco.rest.framework.resource.parameters.Parameters;
+import org.alfresco.util.PropertyCheck;
+import org.springframework.beans.factory.InitializingBean;
+
+
+@RelationshipResource(name = "rule-set-links", entityResource = NodesEntityResource.class, title = "Linking to a rule set")
+public class NodeRuleSetLinksRelation implements InitializingBean, RelationshipResourceAction.Create<RuleSetLink>
+{
+
+    private final Rules rules;
```
krdabrowski, line 49: Logic around rule sets linking could be placed in `RuleSetsImpl` class instead of `RulesImpl`

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
gevagel: Do you think it looks simpler now, Tom?

tpage-alfresco: Not simple enough :)  I don't think you need the statement `if(nodeService.exists(folderNodeRef))`. Removing this allows you to refactor where `ruleSetLink` is created.

remote-api/src/test/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelationTest.java
gevagel: I don't see how else it could be done in this case since that method has to be mocked

remote-api/src/test/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelationTest.java
gevagel: I'm now asserting the equality of the whole list, didn't seem like writing an extra method was required

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -69,6 +76,34 @@ public RuleSet getRuleSetById(String folderNodeId, String ruleSetId, List<String
...
+        final NodeRef linkToNodeRef = validator.validateFolderNode(linkToNodeId, false);
+
+        //The target node should have pre-existing rules to link to
+        if (!ruleService.hasRules(linkToNodeRef)) {
+            throw new AlfrescoRuntimeException("The target node has no rules to link.");
+        }
+
+        //The folder shouldn't have any pre-existing rules
+        if (ruleService.hasRules(folderNodeRef)) {
+            throw new AlfrescoRuntimeException("Unable to link to a ruleset because the folder has pre-existing rules.");
```
tpage-alfresco, line 93: Do you know what status code this returns?  I would assume 500, but we want 40x (maybe 400?)

Same comment for the other AlfrescoRuntimeException above.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -69,6 +76,34 @@ public RuleSet getRuleSetById(String folderNodeId, String ruleSetId, List<String
         return ruleSetLoader.loadRuleSet(ruleSetNode, folderNode, includes);
     }
 
+    @Override
+    public RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId)
+    {
+
+        final NodeRef folderNodeRef = validator.validateFolderNode(folderNodeId,false);
```
tpage-alfresco, line 83: I think this should be true, not false.  I'm not sure, but I'd assume the user needs write access on the folder in order to create a child of it.

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RuleSetsImplTest.java
```diff
@@ -58,10 +63,12 @@
...
 {
-    private static final String FOLDER_ID = "dummy-folder-id";
+    private static final String FOLDER_NODE_ID = "dummy-folder-node-id";
+    private static final String LINK_TO_NODE_ID = "dummy-link-to-node-id";
     private static final String RULE_SET_ID = "dummy-rule-set-id";
-    private static final NodeRef FOLDER_NODE = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, FOLDER_ID);
-    private static final NodeRef RULE_SET_NODE = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, RULE_SET_ID);
+    private static final NodeRef folderNodeRef = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, FOLDER_NODE_ID);
+    private static final NodeRef linkToNodeRef = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, LINK_TO_NODE_ID);
+    private static final NodeRef ruleSetNodeRef = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, RULE_SET_ID);
```
tpage-alfresco, line 71: These constants should be ALL_CAPS, as they were before.

remote-api/src/test/java/org/alfresco/rest/api/nodes/NodeRuleSetsRelationTest.java
```diff
@@ -0,0 +1,75 @@
...
+    private Parameters parameters;
+
+    @InjectMocks
+    private NodeRuleSetLinksRelation nodeRuleSetLinksRelation;
+
+    @Test
+    public void shouldProperlyCreateLink()
+    {
+        RuleSetLink expectedRuleSetLink = new RuleSetLink();
+        List<RuleSetLink> expectedRuleResult = List.of(expectedRuleSetLink);
```
tpage-alfresco, line 63: I would rename these to `ruleSetLink` and `ruleResult` because they're not "expected" at this point.

tpage-alfresco (approved):
Minor comments, some of which should become obvious when we add the e2e tests (e.g. is AlfrescoRuntimeException correct or not).

Generally looks really good :+1:

remote-api/src/main/java/org/alfresco/rest/api/model/rules/RuleSetLink.java
gevagel: So initially I changed this `id` to `folderNodeId` but after a discussion with @mpichura ` we concluded that maybe we should re-think this. One other possible solution would be to add 2 different fields in this class, one for the folder node id and another for the ruleset node id. Or we could simply rename the existing one to `nodeId` and add a comment to provide further information. I wanted to know what others thing because this also impacts the api-explorer

PR #1280, opened 2022-08-08
ACS_3362 GET rule set can include optional owningFolder field.

tpage-alfresco:


tpage-alfresco:
This requires https://github.com/Alfresco/alfresco-tas-restapi/pull/218 to be merged and the SNAPSHOT dependency to be replaced.

krdabrowski (approved):

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -65,7 +66,32 @@ public RuleSet getRuleSetById(String folderNodeId, String ruleSetId, List<String
...
+    }
+
+    /**
+     * Load a rule set for the given node ref.
+     *
+     * @param ruleSetNodeRef The rule set node.
+     * @param includes A list of fields to include.
+     * @return The rule set object.
+     */
+    protected RuleSet loadRuleSet(NodeRef ruleSetNodeRef, List<String> includes)
```
mpichura, line 79: Why is this method `protected`? I assume (but may be wrong) that it should be `private` and additional unit test should be hitting `getRuleSetById` and `getRuleSets`  methods (just with different set of input parameters).

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RuleSetsImplTest.java
```diff
@@ -130,4 +139,14 @@ public void testGetRuleSetById()
 
         assertEquals(RuleSet.of(RULE_SET_ID), actual);
     }
+
+    @Test
+    public void testLoadRuleSet()
```
mpichura, line 144: I think (as stated in previous comment) this test should be split into at least 2 hitting against `getRuleSetById` and `getRuleSets` methods with regards to a different `includes` list parameters (assuming that this list will also grow in the future).

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
tpage-alfresco: The method is protected to allow unit testing it. I expect this method to grow as we add support for more fields, and so I think it makes sense to test it in isolation.  I could move it to a new class instead?

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RuleSetsImplTest.java
tpage-alfresco: I agree that the method will grow, and so we'll want more tests in the future.  I don't think it makes sense to test it twice for each new includes field though.  I'll move the method into a new class and we can unit test it separately.

mpichura (approved):

PR #1286, opened 2022-08-09
ACS-3361 Support for inclusionType.

tpage-alfresco:


tpage-alfresco:
This requires https://github.com/Alfresco/alfresco-tas-restapi/pull/221 to be merged/released/included so that we don't have a SNAPSHOT dependency.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetLoader.java
```diff
@@ -45,15 +53,32 @@
...
+                if (parentRef.equals(folderNodeRef))
+                {
+                    ruleSet.setInclusionType(OWNED);
+                }
+                else
+                {
+                    boolean linked = nodeService.getParentAssocs(ruleSetNodeRef)
+                                           .stream().map(ChildAssociationRef::getParentRef)
+                                           .anyMatch(folderNodeRef::equals);
+                    ruleSet.setInclusionType(linked ? LINKED : INHERITED);
```
krdabrowski, line 79: I'm wondering if there exists a case when rule set could be linked and inherited in the same time. What should be the inclusion value for this case?

tpage-alfresco: The comment and extra unit tests are now added.

krdabrowski: Thanks for clarification and comment

krdabrowski (approved):

PR #1305, opened 2022-08-12
ACS-3429 post support for actions

mpichura:
This is a first and basic installment of the functionality.
It contains the business logic to persist rules with actions so that it is backwards compatible with V0 API ways (Share). This part may still need some polishing as some code is almost copy-pasted from existing solution.
It has basic unit test coverage and small fixes to existing TAS tests so that it wouldn't break the build but still needs a bit more test coverage (only several action types are covered in unit tests). 
It also contains a part of action rule validation which is also used in V0 API (this part may need merging V0 and V1 code into common class)
Next steps:
- cover all action types in `ActionParameterConverter` unit tests 
- enhance TAS tests (better action coverage)
- enhance action rule validation (and merge into common code with V0 parts) when other rule fields mappings are ready


remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -0,0 +1,130 @@
...
+        this.namespaceService = namespaceService;
+    }
+
+    void convertParameters(Map<String, Serializable> params, String name)
+    {
+        final ParameterizedItemDefinition definition = actionService.getActionDefinition(name);
+
+        if (definition == null)
+        {
+            throw new AlfrescoRuntimeException("Could not find definition for action/condition " + name);
```
tpage-alfresco, line 69: I asked this on Georgios' PR. Do you know what status code the AlfrescoRuntimeException returns?

gevagel: It returns a 500 status code

mpichura: I think it returns 500 (not 100% sure)

tpage-alfresco, line 69: [Continuing conversation from other thread] I guess this should return a 404.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        for (Map.Entry<String, Serializable> param : params.entrySet())
+        {
+            final ParameterDefinition paramDef = definition.getParameterDefintion(param.getKey());
+            if (paramDef == null && !definition.getAdhocPropertiesAllowed())
+            {
+                throw new AlfrescoRuntimeException("Invalid parameter " + param.getKey() + " for action/condition " + name);
```
tpage-alfresco, line 77: and this should return some sort of 400 error

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionPermissionValidator.java
```diff
@@ -0,0 +1,83 @@
...
+        this.runtimeActionService = runtimeActionService;
+    }
+
+    Rule validateRulePermissions(Rule rule)
+    {
+        final List<Action> actions = ((CompositeActionImpl) rule.getAction()).getActions();
+
+        checkRestrictedAccessActions(actions);
+        //FIXME must be called when rule type is filled
+        //checkRuleOutboundHasNoCheckOutAction(rule, actions);
```
tpage-alfresco, line 59: Can we delete this and add something to the ACs of the ticket to populate the rule type?

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+@Experimental
+@RunWith(MockitoJUnitRunner.class)
+public class ActionParameterConverterTest
+{
+    private static final String NAMESPACE_URI = "http://www.alfresco.org/model/dictionary/1.0";
+    private static final String VERSIONABLE = "versionable";
+    private static final String CM_PREFIX = "cm";
+    private static final String COLON = ":";
```
tpage-alfresco, line 64: This constant seems a bit excessive :yum:

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final String VERSIONABLE_PREFIX = CM_PREFIX + COLON + VERSIONABLE;
```
tpage-alfresco, line 65: This is the whole aspect name - not a prefix?

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @InjectMocks
+    private ActionParameterConverter objectUnderTest;
+
+    @Test
+    public void testAddAspectConversion()
+    {
+        final String name = AddFeaturesActionExecuter.NAME;
+        final String aspectNameKey = AddFeaturesActionExecuter.PARAM_ASPECT_NAME;
+        final Map<String, Serializable> params = new HashMap<>(1);
+        params.put(aspectNameKey, VERSIONABLE_PREFIX);
```
tpage-alfresco, line 94: Assuming we don't need this to be modifiable then we could use:
```
Map.of(aspectNameKey, VERSIONABLE);
```

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
mpichura: Changing this one as suggested.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
mpichura: Changing this one as suggested.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionPermissionValidator.java
mpichura: I've uncommented that and changed the method to be null safe (at least for the time being).

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
mpichura: I've made some cleanup with these constants :)

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
mpichura: Thanks - renaming that.

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/ActionParameterConverterTest.java
mpichura: Action parameters are actually modified (each entry is replaced with converted parameter value).
I am thinking about refactoring that somehow.

mpichura: That part is refactored - immutable map can be and is used here.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Action.java
```diff
@@ -27,17 +27,28 @@
...
+import org.alfresco.service.cmr.action.ParameterizedItemDefinition;
+import org.alfresco.service.cmr.dictionary.DataTypeDefinition;
+import org.alfresco.service.cmr.dictionary.DictionaryException;
 import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.service.cmr.repository.datatype.DefaultTypeConverter;
+import org.alfresco.service.namespace.QName;
 import org.alfresco.util.GUID;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
```
tpage-alfresco, line 51: There seem to be a lot of new imports here, but you only changed one line of code.

mpichura: Thanks for catching that - I removed unused imports.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -0,0 +1,132 @@
...
+            } catch (ClassNotFoundException e)
+            {
+                throw new DictionaryException("Java class " + javaClassName + " of property type " + typeDef.getName() + " is invalid", e);
+            }
+
+            final int length = ((JSONArray) propertyValue).length();
+            final List<Serializable> list = new ArrayList<>(length);
+            for (int i = 0; i < length; i++)
+            {
+                list.add(convertValue(typeQName, ((JSONArray) propertyValue).get(i)));
```
tpage-alfresco, line 117: JSONArrays are a pain to iterate through.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionPermissionValidator.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private void checkRestrictedAccessActions(List<Action> actions) {
+        actions.forEach(action -> {
+            ActionAccessRestriction.setActionContext(action, ActionAccessRestriction.RULE_ACTION_CONTEXT);
+            runtimeActionService.verifyActionAccessRestrictions(action);
+        });
+    }
+
+    private void checkRuleOutboundHasNoCheckOutAction(Rule rule, List<Action> actions) {
+        //TODO: rule types should never be empty in final implementation
```
tpage-alfresco, line 72: I'm not sure what needs to be "TODONE" here. Are you planning to create a second PR to address it, or is there another Jira ticket maybe?

tpage-alfresco (approved):
LGTM. One comment about a TODO that I wasn't sure whether you meant to leave in the code or not.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
mpichura: Yes, they are, but at least for now I couldn't find a faster/simpler way to make all types of action mappings sustaining backwards compatibility.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionPermissionValidator.java
mpichura: I guess we should throw an exception when `rule.ruleTypes` is null or empty.

mpichura:
I will merge this PR since Source Clear Scan failure is not introduced in the PR.

PR #1322, opened 2022-08-18
ACS-3402: Camel upgrade

mstrankowski:
Moving all declarations of netty libraries from dependent projects up to community-repo, to avoid problems when updating camel libraries


SaraAspery (approved):
LGTM!

mpichura:
Are there any corresponding PRs in other projects that use these `netty` libraries? For instance, Azure Connector will be affected by this change (maybe more projects too - I only had issues with Azure Connector).
If there are such PRs, it would be good to list or mention them here.

mstrankowski:
@mpichura 
Not from what I've seen, I've looked for all mentions of netty in pom.xml's of all projects across Alfresco.

What I've found is that those projects are either excluding netty from their dependencies or are not based on community-repo (azure-connector is).

This change though is honestly equivalent of adding below dependencies to community-repo:

            <dependency>
                <groupId>io.netty</groupId>
                <artifactId>netty-handler-proxy</artifactId>
                <version>${dependency.netty.version}</version>
            </dependency>
            <dependency>
                <groupId>io.netty</groupId>
                <artifactId>netty-tcnative-classes</artifactId>
                <version>${dependency.netty-tcnative.version}</version>
            </dependency>
All other mentioned dependencies are provided anyway by Camel as transitive libraries.

I've just added them to the pom.xml explicitly to keep their versions consistent and easier to track (if they are used outside of community-repo). Now you still have to check what versions are provided by Camel after update (as comments attached indicate) and modify the netty versions in pom.xml to reflect that, you just won't have to do this across 2 different projects (community-repo + azure-connector).

mstrankowski:
Gytheio is using camel, ACS uses Gytheio, so an update in Gytheio, release (0.17 released in this case) and version update in ACS is warranted.

Non-ACS teams projects that use recent versions:
dsync-service seems to declare camel version

Questionable projects (will add info when/if I'll get someone knowledgeable to answer me, ignored for now):
- alfresco-repo-extensions-parent
    has library version defined, references to an old version of alfresco-community-repo - 10.11, so updating one without the other seems to be a bit of a miss.
- alfresco-events-extensions
    extends alfresco-repo-extensions-parent pom, declares camel dependency version in older version than inherited, supposedly should be the same that is used in 10.11 alfresco-repository
- alfresco-messaging-extensions
    just uses versions provided by alfresco-repo-extensions-parent
- alfresco-health-extensions
    just uses versions provided by alfresco-repo-extensions-parent

PR #1323, opened 2022-08-18
ACS-3280 Get inherited rule sets. [tas]

tpage-alfresco:
This needs to work the exact same way as get inherited rules.


krdabrowski (approved):
Nice work!

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -59,10 +59,12 @@ public CollectionWithPagingInfo<RuleSet> getRuleSets(String folderNodeId, List<S
...
-        NodeRef ruleSetNode = ruleService.getRuleSetNode(folderNode);
-        List<RuleSet> ruleSets = Optional.ofNullable(ruleSetNode)
-                                         .map(nodeRef -> ruleSetLoader.loadRuleSet(nodeRef, folderNode, includes))
-                                                                      .stream().collect(toList());
+        List<RuleSet> ruleSets = ruleService.getNodesSupplyingRuleSets(folderNode)
+                                            .stream()
+                                            .map(ruleService::getRuleSetNode)
+                                            .filter(Objects::nonNull)
+                                            .map(nodeRef -> ruleSetLoader.loadRuleSet(nodeRef, folderNode, includes))
+                                            .collect(toList());
```
montgolfiere, line 67: Is there a well-defined order to the rule sets (given that in theory they can be paged) ?

Actually, maybe my question is answered further below, ie. getNodesSupplyingRuleSets(folderNode:

> The order of nodes returned by this methods has to match the order used by {@link #getInheritedRules}.

montgolfiere (approved):
I've added one minor question.

More generally, I'd be interested to know which calls the UI would need to make in order to display "Manage Rules" screen to list rules in order, along with links when clicking on specific rules (eg. to navigate to parent folder/ruleset &/or inherited folder/ruleset).

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
tpage-alfresco: I have reproduced the existing rule order from the v0 API (which is the order in which the rules will be executed).

The order is top-down for a simple hierarchy. When we start adding secondary parents in then the order is not clear to me - I assume it appears consistent because the db decides to return things in a fixed order (it's not based on creation date or anything like that).

I have an additional test that uses secondary parents to demonstrate this issue, which we discussed here: https://alfresco.slack.com/archives/CTWMLBG6N/p1660832592553109

mpichura (approved):
LGTM, just a minor remark.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
tpage-alfresco: Having spent yesterday investigating the test I mentioned then I now feel more confident that it is working as expected. I will push it as a separate PR.

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
```diff
@@ -597,31 +600,74 @@ private boolean checkNodeType(NodeRef nodeRef)
...
+     * Traverse the folder hierarchy find all the folder nodes that could supply rules by inheritance.
+     * <p>
+     * The order of nodes returned by this methods has to match the order used by {@link #getInheritedRules}.
+     *
+     * @param nodeRef The starting node ref.
+     * @param visitedNodeRefs All the visited node refs (will be modified).
+     * @return A list of node refs, starting with the first parent of the first parent of ... and ending with the object generated by the
+     * given node ref.
+     */
+    private List<NodeRef> getNodesSupplyingRuleSets(NodeRef nodeRef, LinkedList<NodeRef> visitedNodeRefs)
```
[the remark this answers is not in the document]
tpage-alfresco: Actually I think we'd be better without the `LinkedList` at all.  I've refactored to have an interface of `List` and to pass an `ArrayList`.

PR #1324, opened 2022-08-18
Support for linking to a single rule set

gevagel:
This PR introduces changes to support linking to a single ruleset by using it's nodeId, expanding upon the existing functionality of linking to a folder that contains rules.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/rules/NodeValidator.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/rules/NodeValidator.java
@@ -39,7 +39,9 @@ import org.alfresco.rest.framework.core.exceptions.InvalidArgumentException;
 import org.alfresco.rest.framework.core.exceptions.PermissionDeniedException;
 import org.alfresco.rest.framework.core.exceptions.RelationshipResourceNotFoundException;
 import org.alfresco.service.Experimental;
+import org.alfresco.service.cmr.repository.ChildAssociationRef;
 import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.service.cmr.repository.NodeService;
 import org.alfresco.service.cmr.rule.RuleService;
 import org.alfresco.service.cmr.security.PermissionService;
 import org.alfresco.service.namespace.QName;
@@ -53,6 +55,7 @@ public class NodeValidator
     private Nodes nodes;
     private RuleService ruleService;
     private PermissionService permissionService;
+    private NodeService nodeService;
 
     /**
      * Validates if folder node exists and the user has permission to use it.
@@ -105,7 +108,8 @@ public class NodeValidator
     public NodeRef validateRuleSetNode(String linkToNodeId, boolean requireChangePermission)
     {
         final Node ruleSetNode = nodes.getNode(linkToNodeId);
-        final NodeRef parentNode = ruleSetNode.getParentId();
+        final ChildAssociationRef primaryParent = nodeService.getPrimaryParent(ruleSetNode.getNodeRef());
+        final NodeRef parentNode = primaryParent.getParentRef();
         validateChangePermission(requireChangePermission, parentNode);
         return parentNode;
     }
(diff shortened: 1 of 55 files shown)


KikoUK:
Are we going to add e2es as part of this PR or Is there a separate jira for adding e2es for this functionality?

gevagel:
I think the E2Es for this functionality are covered by this issue https://alfresco.atlassian.net/browse/ACS-3431

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/NodeValidator.java
```diff
@@ -142,6 +142,24 @@ private NodeRef validateNode(final String nodeId, final QName expectedType, fina
         return nodeRef;
     }
 
+    private void validateChangePermission(boolean requireChangePermission, NodeRef nodeRef)
```
krdabrowski, line 145: Method name is confusing for me, because you could try to validate change permission without requirement for change permission: `validateChangePermission(false, nodeRef);` - in this case you are validating read permission. You could simply change method name to e.g. `validatePermission()`.

krdabrowski: Or you could solve it more generally and pass expected permission as an input param instead of boolean, e.g.:
`
validatePermission(NodeRef nodeRef, String expectedPermission) {
if (READ_PERMISSIONS.equals(expectedPermission) {
  if (permissionService.hasReadPermission(nodeRef) != ALLOWED)...
else
  if (permissionService.hasPermission(nodeRef, expectedPermission) != ALLOWED)...
`

krdabrowski: Solution is on you :)

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RuleSetsImplTest.java
```diff
@@ -172,6 +173,32 @@ public void testLinkingToRuleSet()
...
+    {
+        NodeRef childNodeRef = new NodeRef(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, "dummy-child-id");
+
+        given(nodeValidatorMock.isRuleSetNode(any())).willReturn(true);
+        given(ruleServiceMock.hasRules(any(NodeRef.class))).willReturn(true, false);
+        given(runtimeRuleServiceMock.getSavedRuleFolderAssoc(any(NodeRef.class))).willReturn(assocRef);
+        given(assocRef.getChildRef()).willReturn(childNodeRef);
+
+        //when
+        assertEquals(ruleSets.linkToRuleSet(FOLDER_ID,LINK_TO_NODE_ID).getId(), childNodeRef.getId());
```
krdabrowski, line 187: Assuming that test methods should be divided into 3 blocks: given, when and then, assertions should be placed in the last block. At first glance I didn't noticed that here is an assertion.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/NodeValidator.java
gevagel: Thanks for the feedback Krystian. This method got created to avoid duplicate code, as you can see from the removed lines. I think that changing the boolean in the method would require a lot more work in a bigger scope as the argument gets passed from the methods that are calling this one. Do you think we could rename the method to something like checkPermissions to make it easier to understand and avoid changing the code?

krdabrowski: Of course, I'm ok with the method name update fix.

krdabrowski (approved):

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/NodeValidatorTest.java
```diff
@@ -329,6 +359,12 @@ public void testValidateRuleNode_notAssociatedRuleToRuleSet()
         then(ruleServiceMock).shouldHaveNoMoreInteractions();
     }
 
+    @Test
+    public void testIsRuleSetNode()
+    {
+        Assert.assertTrue(nodeValidator.isRuleSetNode(RULE_SET_ID));
```
mpichura: You could also cover `nodeValidator.isRuleSetNode()` returning false.

KikoUK (approved):

mpichura (approved):

PR #1327, opened 2022-08-19
MNT-23103 - Downloading a document replaces filename having space with + character

kmagdziarz:
MNT-23103 Change encoder for content disposition filename to be consistent with RFC5987.

Add test covering filename encoding.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
+++ b/remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
@@ -30,6 +30,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.SocketException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -59,6 +60,7 @@ import org.alfresco.service.namespace.QName;
 import org.alfresco.util.TempFileProvider;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.cxf.attachment.Rfc5987Util;
 import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.extensions.webscripts.Cache;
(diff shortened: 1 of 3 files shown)


kcichonczyk (approved):

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
```diff
@@ -496,6 +498,20 @@ public void setAttachment(WebScriptRequest req, WebScriptResponse res, boolean a
...
+
+    private String encodeFilename(String attachFileName)
+    {
+        try
+        {
+            return Rfc5987Util.encode(attachFileName);
+        } catch (UnsupportedEncodingException e)
+        {
+            if (logger.isInfoEnabled())
+                logger.info(e.getMessage() + " Changing encoder.");
```
SaraAspery, line 510: Do you think it's worth including the names of the previous and replaced encoders in this log message?

Personally, I prefer even a single `if` statement to be in its own block to make it easier to avoid mistakes later, but I think it's a preference issue, and I'm not going to argue with Robert Martin :)

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
```diff
@@ -496,6 +498,20 @@ public void setAttachment(WebScriptRequest req, WebScriptResponse res, boolean a
             res.setHeader("Content-Disposition", headerValue);
         }
     }
+
+    private String encodeFilename(String attachFileName)
+    {
+        try
+        {
+            return Rfc5987Util.encode(attachFileName);
+        } catch (UnsupportedEncodingException e)
```
SaraAspery, line 507: The catch statement should be on its own line. This Alfresco style link is old but still applies [https://hub.alfresco.com/t5/alfresco-content-services-hub/coding-standards-for-alfresco-content-services/ba-p/290457](https://hub.alfresco.com/t5/alfresco-content-services-hub/coding-standards-for-alfresco-content-services/ba-p/290457)

SaraAspery (changes requested):
Looks good Kacper, just a couple of very minor points re the catch statement style and maybe enhance a log message

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
kmagdziarz: Ah, didn't notice, thanks.

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentStreamer.java
kmagdziarz: Added info about replaced encoders.
I will stay with current `if` for logger, to be compliant with rest of the class.

SaraAspery (approved):
LGTM!

PR #1330, opened 2022-08-22
MNT-23043 - Property Tables Cleaner Job V3

evasques:
Implementation of a v3 algorithm for Property Tables Cleaner Job:

- Single implementation for all DBMS
- Queries were changed in order to filter by unique values (some tables are mostly duplicate values) and all of the selects now limit the result set to the batch size
- Storing in memory the result sets of the secondary tables (a bit slower than using the cursors but when we have aborted connections (with no error communicated to the repo), these result sets are not properly closed and we sometimes see errors on other threads that are accessing the prop_values tables. Also the previous approach causes more memory consumption on the DB side. 
- As we now store in memory the results from the secondary tables, we were able to optimize crossing the data: we now have a single set of values from the secondary tables that is being populated as we gather the data. In the end, it’s that single list that we cross with the results from the primary table and see what values we can delete (all values that are on the primary that aren’t in any of the secondary tables)
- Pause and Recovery: every {pauseAndRecoverBatchSize} rows deleted we close all prepared statements and close the connection and sleep for {pauseAndRecoverTime} milliseconds. This is necessary to allow the DBMS to perform the background tasks without load from ACS. When we do not do this and if we are performing millions of deletes, the connection eventually gets aborted.


tiagosalvado10 (approved):
Surely the final decision won't be mine but as a person who has followed this approach since the beginning and who is aware of the dozen (most likely hundreds) hours Eva has ran it, I can assure the implementation works - it removes some issues from V1 and V2 (e.g., aborted connections) but also seems more optimized from a global perspective (when comparing with the other versions)

jottley (approved):

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+        this.dataSource = dataSource;
+    }
+
+    @Override
+    public void execute() throws Exception
+    {
+        checkProperties();
+
+        String pauseAndRecoverBatchSizeString = globalProperties.getProperty(PROPERTY_PAUSE_AND_RECOVER_BATCHSIZE);
+        pauseAndRecoverBatchSize = pauseAndRecoverBatchSizeString == null ? 500000
```
tpage-alfresco, line 97: 500000 could be a constant - something like `DEFAULT_MAX_BATCH_SIZE`.  (I'm not quite sure at this point in the review how this works)

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+                : Long.parseLong(pauseAndRecoverBatchSizeString);
+
+        String pauseAndRecoverTimeString = globalProperties.getProperty(PROPERTY_PAUSE_AND_RECOVER_TIME);
+        pauseAndRecoverTime = pauseAndRecoverTimeString == null ? 300000 : Long.parseLong(pauseAndRecoverTimeString);
```
tpage-alfresco, line 101: Similarly 300000 could be a constant - something like `DEFAULT_RECOVERY_PAUSE_MS`.

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+
+        return sqlBuilder.toString();
+    }
+
+    @Override
+    protected Long processPrimaryTableResultSet(PreparedStatement primaryPrepStmt, PreparedStatement[] secondaryPrepStmts,
+            PreparedStatement deletePrepStmt, Set<Long> deleteIds, String primaryTableName, String primaryColumnName,
+            Pair<String, String>[] tableColumn) throws SQLException
+    {
+        int rowsProcessed = 0;
```
tpage-alfresco, line 244: I don't think we need to initialise this here - it always gets overridden below.

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+        }
+
+        if (potentialIdsToDelete.size() == 0)
+        {
+            // Nothing more to do
+            return primaryId;
+        }
+
+        Long minPotentialId = Collections.min(potentialIdsToDelete);
+        Long maxPotentialId = Collections.max(potentialIdsToDelete);
```
tpage-alfresco, line 267: I don't know how large this set of potential ids is, but we could compute the min and max during creation of the set above.  Maybe this optimisation is completely negligible though

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+        rowsProcessed = potentialIdsToDelete.size();
```
tpage-alfresco, line 268: The tense of this is wrong. It should be `rowsToProcess`?

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,437 @@
...
+
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("RowsProcessed " + rowsProcessed + " from primary table " + primaryTableName + ". Primary: ["
+                    + minPotentialId + "," + maxPotentialId + "] Secondary: [" + minSecId + "," + maxSecId + "] Total Deleted: "
+                    + deletedCount);
+        }
+
+        // Do we need to pause?
+        if (processedCounter >= pauseAndRecoverBatchSize)
```
tpage-alfresco, line 310: I couldn't see how the batching worked. I'm sure it does, because I know you've tested it, but this is really a comment about the clarity of the code/documentation.  From what I saw then it looked like all the rows have been processed at this point and so it doesn't matter if we pause or not.

tpage-alfresco (commented):
Partial review

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
evasques: The counter is only reset after each pause. Every row in each batch is processed, but when we reach the end of each batch we verify if we need to mark the job to pause before starting the next batch due to the pauseAndRecoverBatchSize being reached. If the flag is set to true, we have a break (L176), we close everything, pause and then resume where we left off (L205). 

When doing massive deletes on the DB after a period of time we start seeing severe performance degradation on the DB that leads to aborted connections. The pause was implemented to prevent that. After X rows processed, we stop for Y amount of time to allow the DB to recover before continuing.

tpage-alfresco: I don't see what's controlling the batches though. This is what I think the flow looks like:

```
process
{
    processBatch
    {
        pauseAndRecover = false;
        if (hasResults)
        {
            while (hasResults && !isTimeoutExceeded())
            {
                processPrimaryTableResultSet
                {
                    if (processedCounter >= pauseAndRecoverBatchSize)
                    {
                        pauseAndRecover = true;
                    }
                }
            }
        }
        if (pauseAndRecover)
        {
            pauseAndRecoverJob
            processBatch
        }
    }
}
```

So `process` always calls `processBatch` once and the only pause can happen after the whole lot is processed, or `isTimeoutExceeded()` returns true.  Maybe the code calling `process` handles the batching?  Or maybe I just missed how the batch is created.

evasques: Its in the processBatch. Completing your flow:

`
process
{
    processBatch
    {
        pauseAndRecover = false;
        if (hasResults)
        {
            while (hasResults && !isTimeoutExceeded())
            {
                primaryId = processPrimaryTableResultSet
                {
                    if (processedCounter >= pauseAndRecoverBatchSize)
                    {
                        pauseAndRecover = true;
                    }
                }
                
               // If the next primaryId is null or flag to pause is true, exit
               if (primaryId == null || pauseAndRecover)
               {
                        break;
               }
                
               // Prepare for next batch doing the queries starting on primaryId + 1 and execute the query for the next batch
               primaryPrepStmt.setLong(1, primaryId + 1);
               primaryPrepStmt.setLong(2, tableUpperLimits[0]);
               hasResults = primaryPrepStmt.execute();
            }
        }
        if (pauseAndRecover)
        {
            pauseAndRecoverJob
            processBatch from last primaryId
        }
    }
}`

repository/src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsV3Executor.java
```diff
@@ -0,0 +1,501 @@
...
+                if (id > maxSecId || secondaryResults.contains(id))
+                {
+                    it.remove();
+                }
+            }
+
+            // The next starting primary ID for the next batch will either be the next last id evaluated from the
+            // primary table or, in case the secondary queries did not get that far, the last secondary table id
+            // evaluated (maxSecId)
+            primaryId = primaryId < maxSecId ? primaryId : maxSecId;
```
tpage-alfresco, line 305: This is `min(primaryId, maxSecId)` and above we had something equivalent to `max(primaryId, maxPotentialId)`.  I think using these tertiary operators is fine, but this was just something that went through my mind as I read the code.

tpage-alfresco (approved):
LGTM

PR #1332, opened 2022-08-23
ACS-3429, ACS-3336: Rule action mappings (create), validations (update) - part 2

mpichura:
This PR is a second installment for https://github.com/Alfresco/alfresco-community-repo/pull/1305.
This PR roughly consists of:
- initial TAS tests for create rule with multiple actions
- action parameters mappings for rule retrieval + unit tests
- rule/action validations for rule update + unit tests

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
@@ -397,7 +397,6 @@ public class CreateRulesTests extends RestTest
 
         restClient.assertStatusCodeIs(CREATED);
         rule.assertThat().isEqualTo(expectedRuleModel, IGNORE_ID, IGNORE_IS_SHARED)
-                .assertThat().field("name").is(RULE_NAME_DEFAULT)
                 .assertThat().field("isShared").isNull();
     }
 }
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
@@ -374,13 +374,13 @@ public class CreateRulesTests extends RestTest
     public void createRuleWithActions()
     {
         final Map<String, Serializable> copyParams =
-                Map.of("destination-folder", "workspace://SpacesStore/dummy-folder-node", "deep-copy", true);
+                Map.of("destination-folder", "dummy-folder-node", "deep-copy", true);
         final RestActionBodyExecTemplateModel copyAction = createCustomActionModel("copy", copyParams);
         final Map<String, Serializable> checkOutParams =
-                Map.of("destination-folder", "workspace://SpacesStore/dummy-folder-node", "assoc-name", "cm:checkout", "assoc-type",
+                Map.of("destination-folder", "dummy-folder-node", "assoc-name", "cm:checkout", "assoc-type",
                         "cm:contains");
         final RestActionBodyExecTemplateModel checkOutAction = createCustomActionModel("check-out", checkOutParams);
-        final Map<String, Serializable> scriptParams = Map.of("script-ref", "workspace://SpacesStore/dummy-script-node-id");
+        final Map<String, Serializable> scriptParams = Map.of("script-ref", "dummy-script-node-id");
         final RestActionBodyExecTemplateModel scriptAction = createCustomActionModel("script", scriptParams);
         final RestRuleModel ruleModel = createRuleModelWithDefaultName();
         ruleModel.setActions(Arrays.asList(copyAction, checkOutAction, scriptAction));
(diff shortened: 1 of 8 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RulesTestsUtils.java
```diff
@@ -100,6 +102,24 @@ public static RestActionBodyExecTemplateModel createDefaultActionModel()
         return restActionModel;
     }
 
+    public static List<RestActionBodyExecTemplateModel> addActionContextParams(List<RestActionBodyExecTemplateModel> inputActions)
+    {
+        inputActions.forEach(inputAction -> {
+            final Map<String, Serializable> params = new HashMap<>((Map<String, Serializable>) inputAction.getParams());
+            params.put("actionContext", "rule");
```
krdabrowski, line 109: If this param is missing in default action, you could simply fix the `createDefaultActionModel()`

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
```diff
@@ -130,6 +131,20 @@ private org.alfresco.service.cmr.rule.Rule mapToServiceModelAndValidateActions(R
         return actionPermissionValidator.validateRulePermissions(serviceModelRule);
     }
 
+    private Rule loadRuleAndConvertActionParams(org.alfresco.service.cmr.rule.Rule ruleModel, List<String> includes)
+    {
+
+        final Rule rule = ruleLoader.loadRule(ruleModel, includes);
+        rule.getActions()
```
krdabrowski, line 138: Minor, this conversion of params could be done within `Action.from()` so it would be made in one place, but I'm also ok with this approach.

krdabrowski (approved):

krdabrowski:
Added minor comments, but overall LGTM.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RulesTestsUtils.java
mpichura: This parameter is not supposed to be passed as input but is added during rule action validation (on create/update).
That is why I didn't add it in `createDefaultActionModel()` and created this method as it is and most likely will be used in multiple tests when we need to create expected rule with actions.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
```diff
@@ -362,4 +366,38 @@ public RestRuleModel testRolePermissionsWith(UserRole userRole)
...
+                .createSingleRule(ruleModel);
+
+        final RestRuleModel expectedRuleModel = createRuleModelWithDefaultName();
+        expectedRuleModel.setActions(addActionContextParams(Arrays.asList(copyAction, checkOutAction, scriptAction)));
+        expectedRuleModel.setConditions(createEmptyConditionModel());
+        expectedRuleModel.setTriggers(List.of("inbound"));
+
+        restClient.assertStatusCodeIs(CREATED);
+        rule.assertThat().isEqualTo(expectedRuleModel, IGNORE_ID, IGNORE_IS_SHARED)
+                .assertThat().field("name").is(RULE_NAME_DEFAULT)
```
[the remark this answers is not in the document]
mpichura: Thanks for catching that - I'll remove that duplicated check.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
mpichura: I am not happy with approach either but IMO, all those mappings (`Rule`, `Action`, `Condition` etc.) should be done in mapper classes responsible only for mapping to/from layers. 
That way we could inject mappers and additional dependencies inside mappers via DI framework mechanisms instead of passing them as method parameters (as for instance in `Rule::toServiceModel`).
If I were to map the parameters via `Action.from()` I'd have to pass dependencies via method parameters through 2 or 3 methods (`Rule` -> `CompositeAction` -> `Action`) which I think is a worse solution.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -89,6 +91,21 @@ Map<String, Serializable> getConvertedParams(Map<String, Serializable> params, S
...
     }
 
+    public Serializable convertParamFromServiceModel(Serializable param)
+    {
+        if (param instanceof QName)
+        {
+            return ((QName) param).toPrefixString(namespaceService);
+        }
+        else if (param instanceof NodeRef) {
+            return param.toString();
```
montgolfiere, line 101: Is this just returned as a node id (for V1 style API) ? (in other words, the store ref is assumed to be workspace://SpacesStore for V1 Nodes API).

mpichura: It is returned as `workspace://SpacesStore/node-id` - same way as it is expected as an input param.
Shall I change that to accept/return `nodeId` only?

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RulesImpl.java
krdabrowski: I agree with that, intermediate mapping layer would be much better approach.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
mpichura: Changed to accept/return `nodeId` only.

mpichura:
I will merge this PR since Source Clear Scan failure is not introduced in the PR.

PR #1343, opened 2022-08-30
ACS-3465 Migrating to GitHub actions

mikolajbrzezinski:


amps/ags/rm-automation/rm-automation-community-rest-api/src/test/resources/testng.xml
```diff
@@ -1,9 +1,16 @@
 <!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
 
 <suite name="TestNG AllTestSuite" configfailurepolicy="continue" verbose="1" time-out="300000">
+   <test name="restapi - test priorities" >
```
dsibilio, line 4: This fix for the AGS tests not working due to re-ordering after the switch from Travis to GHA is mostly fine. However I'd still prefer for the tests to be idempotent and not depend on the ordering if possible. It would be nice to figure out exactly what breaks these tests and if it's possible to make them independent from ordering.

dsibilio (changes requested):
Left some review comments which don't really apply to the PoC, but are mostly a reminder of things that should still be done before a PR like this could actually be merged _(along with fully completing the migration, rather than sampling each Travis job of course)_.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,435 @@
+name: Alfresco Community Repo CI
+
+on:
+  pull_request:
```
gionn, line 4: it's usually good to add the filter for the default branch if there aren't reasons to avoid that

.github/workflows/ci.yml
```diff
@@ -0,0 +1,435 @@
...
+  push:
+    branches:
+      - feature/**
+      - fix/**
```
gionn, line 8: I don't know the branching model adopted in this repo, but unless those branches are never merged into master using a PR, you are going to have duplicated builds (triggered both by pull_request and push on the same branch) - exclude those branches from pull_request or avoid at all having the push event here

.github/workflows/ci.yml
```diff
@@ -0,0 +1,435 @@
...
+    runs-on: ubuntu-latest
+    if: >
+      ! contains(github.event.head_commit.message, '[skip tests]') &&
+      ! contains(github.event.head_commit.message, '[force]')
+    steps:
+      - uses: actions/checkout@v3
+      - name: "Clean cache"
+        run: bash ./scripts/travis/cleanup_cache.sh
+      - name: "Prepare environment"
+        run: bash ./scripts/travis/prepare.sh
```
gionn, line 36: these scripts will always work on an empty .m2 folder if cache is not restored before

If I understood correctly what they are for, you probably need to author a custom composite action to reuse this steps (along with cache restore) in any job that would invoke maven

this job as it is can be removed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,435 @@
...
+          cache: "maven"
+      - name: "Set Transformers tag"
+        run: |
+          echo "TRANSFORMERS_TAG=$(mvn help:evaluate -Dexpression=dependency.alfresco-transform-core.version -q -DforceStdout)" >> $GITHUB_ENV
+      - name: "docker build 1"
+        run: docker run -d -p 5433:5432 -e POSTGRES_PASSWORD=alfresco -e POSTGRES_USER=alfresco -e POSTGRES_DB=alfresco postgres:14.4 postgres -c 'max_connections=300'
+      - name: "docker build 2"
+        run: docker run -d -p 61616:61616 -p 5672:5672 alfresco/alfresco-activemq:5.17.1-jre11-rockylinux8
+      - name: "docker build 3"
+        run: docker run -d -p 8090:8090 -e JAVA_OPTS=" -Xms256m -Xmx256m" alfresco/alfresco-transform-core-aio:${{ env.TRANSFORMERS_TAG }}
```
gionn, line 115: it's far more maintainable to define a docker-compose.yml and launch required infra services with that

PR #1344, opened 2022-08-30
Upgraded Keycloak-Client to match the Identity-Service version.

jamalkm:
Upgraded Keycloak-Client to 18.0.0. This is the same Keycloak server version that Identity Service (IDS) is shipping. 
We should always be in sync with the server. IDS was released back in June, but somehow we missed to upgrade the clients.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
         <dependency.truezip.version>7.7.10</dependency.truezip.version>
         <dependency.poi.version>5.2.2</dependency.poi.version>
         <dependency.ooxml-schemas.version>1.4</dependency.ooxml-schemas.version>
-        <dependency.keycloak.version>15.0.2</dependency.keycloak.version>
+        <dependency.keycloak.version>18.0.0</dependency.keycloak.version>
         <dependency.jboss.logging.version>3.5.0.Final</dependency.jboss.logging.version>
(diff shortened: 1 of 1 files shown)


SaraAspery (approved):

damianujma:
LGTM. Shouldn't we also upgrade Keycloak-Client to 18.0.0. for [alfresco-tas-utility](https://github.com/Alfresco/alfresco-tas-utility/blob/master/pom.xml#L41) and [alfresco-enterprise-repo](https://github.com/Alfresco/alfresco-enterprise-share/blob/924e878c9401fb904f9f9621ad1ca948e22e0d19/pom.xml#L66) ?

damianujma (approved):

jamalkm:
> LGTM. Shouldn't we also upgrade Keycloak-Client to 18.0.0. for [alfresco-tas-utility](https://github.com/Alfresco/alfresco-tas-utility/blob/master/pom.xml#L41) and [alfresco-enterprise-repo](https://github.com/Alfresco/alfresco-enterprise-share/blob/924e878c9401fb904f9f9621ad1ca948e22e0d19/pom.xml#L66) ?

Good catch. For enterprise-share this is for GL. I'll create a PR for them!

PR #1348, opened 2022-08-31
ACS-3463: Updating AIS and ATC versions

mstrankowski:


kmagdziarz (approved):

mstrankowski:
Only SCA fails with same vulnerability already ignored on master

PR #1356, opened 2022-09-02
ACS-3430: POST support for conditions

krdabrowski:


remote-api/src/main/java/org/alfresco/rest/api/model/rules/CompositeCondition.java
```diff
@@ -50,7 +52,7 @@
      * @param actionConditions - list of {@link ActionCondition} service POJOs
      * @return {@link CompositeCondition} REST model
      */
-    public static CompositeCondition from(final List<ActionCondition> actionConditions)
+    public static CompositeCondition from(final List<ActionCondition> actionConditions, final NamespaceService namespaceService)
```
mpichura, line 55: I think we should introduce a mapping "layer" for mapping from/to service and API models instead of mapping inside the model classes.

remote-api/src/main/java/org/alfresco/rest/api/model/rules/Rule.java
```diff
@@ -60,7 +61,7 @@
      * @param ruleModel - {@link org.alfresco.service.cmr.rule.Rule} service POJO
      * @return {@link Rule} REST model
      */
-    public static Rule from(final org.alfresco.service.cmr.rule.Rule ruleModel)
+    public static Rule from(final org.alfresco.service.cmr.rule.Rule ruleModel, final NamespaceService namespaceService)
```
mpichura, line 64: This namespace service is passed through 2 methods down to it's actual use which seems not to clean.
Thus, I think we should introduce a mapping "layer" for mapping from/to service and API models instead of mapping inside the model classes.

mpichura (approved):
LGTM. Only some remarks for future refactor.

tpage-alfresco (approved):
LGTM

PR #1373, opened 2022-09-12
ACS-3489 Use specified rule set.

tpage-alfresco:


repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
```diff
@@ -311,7 +311,17 @@
...
     public NodeRef getOwningNodeRef(Action action);
-    
+
+    /**
+     * Returns the owning node reference for a rule.
+     *
+     * @param ruleSet The rule set node.
+     * @return the owning node reference
+     */
+    @Auditable (parameters = { "rule" })
```
tpage-alfresco, line 321: This should say `ruleSet`.

mpichura (approved):
LGTM and I'm impressed (again) with the explanatory javadoc in test class.

PR #1383, opened 2022-09-14
ACS-3536 Omit actionContext for rules.

tpage-alfresco:
It will always be set to 'Rule' anyway.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
@@ -218,7 +218,9 @@ public class UpdateRulesTests extends RestTest
                                               .updateRule(rule.getId(), rule);
 
         restClient.assertStatusCodeIs(OK);
-        updatedRule.assertThat().field("name").is("Updated rule name");
+        updatedRule.assertThat().field("name").is("Updated rule name")
+                   .assertThat().field("actions.actionDefinitionId").is(List.of("copy"))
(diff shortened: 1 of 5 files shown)


tpage-alfresco:
This needs some more work to investigate why we have a method `RulesTestsUtils.addActionContextParams`.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -198,6 +200,27 @@ public void updateRuleWithIncludedFields()
...
+        RestRuleModel rule = createAndSaveRule("Rule name", List.of(copyAction));
+
+        STEP("Try to update the rule.");
+        rule.setName("Updated rule name");
+        RestRuleModel updatedRule = restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet()
+                                              .include("isShared")
+                                              .updateRule(rule.getId(), rule);
+
+        restClient.assertStatusCodeIs(OK);
+        updatedRule.assertThat().field("name").is("Updated rule name");
```
mpichura, line 221: Should we also check that `updatedRule` also has an action parameter `actionContext->rule`?
Or even we could do a similar check to the one we do in `CreateRulesTests`:
`updatedRule.assertThat().isEqualTo(rule, <list of ignored fields if needed>);`

mpichura (approved):
LGTM - just had a small remark on the E2E test (not a blocker for merge)

KikoUK (approved):

nikita-web-ua (approved):

PR #1386, opened 2022-09-15
ACS-3363 Rule mappers layer pt1

mpichura:
This PR introduces a mapper for `SimpleCondition`, so there will be more mappers coming in next PRs.


mpichura:
Closing because branch name was wrong. I will reopen from a valid branch.

PR #1387, opened 2022-09-15
ACS-3363 Support inheritedBy in GET rule sets.

tpage-alfresco:


mpichura (approved):
LGTM

tpage-alfresco:
I ran the new TAS test 100 times locally alongside the other tests in the class, and it passed every time. I also tried running all the tests in the "part1" suite locally, and they also passed (twice).  I'll try rerunning on Travis to see if this is consistently failing, and then I'll have to think about next steps.

PR #1388, opened 2022-09-15
ACS-3354 Rule mappers pt1 (simple condition mapper)

mpichura:
This PR introduces a mapper for `SimpleCondition`, so there will be more mappers coming in next PRs.


remote-api/src/main/java/org/alfresco/rest/api/model/rules/SimpleCondition.java
```diff
@@ -91,92 +90,20 @@ public static List<SimpleCondition> listOf(final List<ActionCondition> actionCon
      * @param actionCondition - {@link ActionCondition} service POJO
      * @return {@link SimpleCondition} REST model
      */
-    public static SimpleCondition from(final ActionCondition actionCondition, final NamespaceService namespaceService)
+    public static SimpleCondition from(final ActionCondition actionCondition,
+                                       final RestModelMapper<SimpleCondition, ActionCondition> simpleConditionMapper)
     {
-        if (actionCondition == null || actionCondition.getActionConditionDefinitionName() == null || actionCondition.getParameterValues() == null)
+        if (actionCondition == null || actionCondition.getActionConditionDefinitionName() == null ||
+                actionCondition.getParameterValues() == null || actionCondition.getParameterValues().isEmpty())
```
tpage-alfresco, line 97: Why did you add `actionCondition.getParameterValues().isEmpty()` to this condition?

remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -946,6 +946,26 @@
         <property name="ruleSettings" ref="RuleSettings" />
     </bean>
 
+    <bean id="simpleConditionMapper" class="org.alfresco.rest.api.impl.mapper.rules.RestRuleSimpleConditionModelMapper">
+        <constructor-arg name="namespaceService" ref="NamespaceService"/>
+        <constructor-arg name="nodes" ref="Nodes"/>
+    </bean>
+
+
+    <bean id="RestModelMapper" class="org.springframework.aop.framework.ProxyFactoryBean">
```
tpage-alfresco, line 955: Do we need this bean?  Doesn't the exception handling already happen elsewhere?

remote-api/src/main/java/org/alfresco/rest/api/model/rules/SimpleCondition.java
mpichura: Thanks for spotting that - actually I think I meant to replace `actionCondition.getParameterValues() == null` with `actionCondition.getParameterValues().isEmpty()` as it appears that at the moment `actionCondition.getParameterValues()` can never be null:
```
public Map<String, Serializable> getParameterValues()
    {
        Map<String, Serializable> result = this.parameterValues;
        if (result == null)
        {
            result = new HashMap<String, Serializable>();
        }
        return result;
    }
```
Would that make sense?

tpage-alfresco: It's not clear to me what this should say. I just checked the `toRestModel` method - and it starts with:
```
if (actionCondition == null || actionCondition.getActionConditionDefinitionName() == null || actionCondition.getParameterValues() == null)
        {
            return null;
        }
```

So maybe we can remove this validation completely and use the check in `toRestModel` instead?

mpichura: I meant to get rid of `SimpleCondition` duplicate part when removing `SimpleCondition::from` but you're right - this can be done within this PR. 
I'll push changes soon.

remote-api/src/main/resources/alfresco/public-rest-context.xml
mpichura: I thought we needed that one but it seems that not, so removed it.

tpage-alfresco (approved):
LGTM

krdabrowski (approved):
LGTM 👍 I like the approach with `interface RestModelMapper<R, S>` 👍

PR #1411, opened 2022-09-19
ACS-3509: Adding E2E TAS REST tests for update rule conditions.

mpichura:


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -295,6 +301,107 @@ public void updateRuleFields()
...
+    @Test (groups = { TestGroup.REST_API, TestGroup.RULES, TestGroup.SANITY })
+    public void updateRuleAddConditions()
+    {
+        final RestRuleModel rule = createAndSaveRule(createRuleModelWithModifiedValues());
+
+        STEP("Try to update the rule and add conditions.");
+        rule.setConditions(createVariousConditions());
+
+        final RestRuleModel updatedRule = restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet()
+                .include(IS_SHARED)
```
krdabrowski, line 314: Is the "shared" flag relevant? Maybe it will be better to skip including it in this tests as it's not being asserted in any way. It will make the tests a bit more clear.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -295,6 +301,107 @@ public void updateRuleFields()
...
+                .updateRule(rule.getId(), rule);
+
+        //set expected object
+        rule.setConditions(createCompositeCondition(null));
+        restClient.assertStatusCodeIs(OK);
+        updatedRule.assertThat().isEqualTo(rule, ID, IS_SHARED)
+                .assertThat().field(ID).isNotNull();
+    }
+
+    /** Check we can use the POST response and update rule by removing all conditions. */
```
krdabrowski, line 384: This comment is a bit confusing for me because it duplicates text from above test method: `updateRuleRemoveAllConditions()` comment.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
mpichura: You're right - I just removed `shared` usage from all added tests.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -287,14 +293,110 @@ public void updateRuleFields()
...
+                .updateRule(rule.getId(), rule);
+
+        restClient.assertStatusCodeIs(OK);
+        updatedRule.assertThat().isEqualTo(rule, ID)
                 .assertThat().field(ID).isNotNull();
     }
 
+    /** Check we can use the POST response and update rule by removing all conditions. */
+    @Test (groups = { TestGroup.REST_API, TestGroup.RULES, TestGroup.SANITY })
+    public void updateRuleRemoveAllConditions()
```
tpage-alfresco, line 360: What's the difference between this test and updateRuleAddNullConditions?

tpage-alfresco: Ah - I think the point is that if we omit the conditions section completely then we still expect the response to contain a composite condition.  Maybe we could add something like that to the javadoc?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -287,14 +293,110 @@ public void updateRuleFields()
...
+                .updateRule(rule.getId(), rule);
+
+        //set expected object
+        rule.setConditions(createCompositeCondition(null));
+        restClient.assertStatusCodeIs(OK);
+        updatedRule.assertThat().isEqualTo(rule, ID)
+                .assertThat().field(ID).isNotNull();
+    }
+
+    /** Check we get a 404 error when using the POST response and update rule by adding invalid condition. */
```
tpage-alfresco, line 379: I think you should say that we're referencing a non-existent category in the condition.  Presumably other invalid conditions will give 400 errors instead?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -287,14 +293,110 @@ public void updateRuleFields()
...
+        STEP("Try to update the rule with invalid condition.");
+        final RestCompositeConditionDefinitionModel conditions = createCompositeCondition(
+                List.of(createCompositeCondition(!INVERTED, List.of(createSimpleCondition("category", "equals", "fake-category-id")))));
+        rule.setConditions(conditions);
+
+        restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet()
+                .include(IS_SHARED)
+                .updateRule(rule.getId(), rule);
+
+        restClient.assertStatusCodeIs(BAD_REQUEST);
```
tpage-alfresco, line 396: The javadoc said 404, but this looks like 400...

mpichura: Yep - my mistake - fixed that.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
mpichura: It is more like create a rule with some conditions and update it so that there aren't any conditions. 
Test method `updateRuleAddNullConditions` was meant to create a rule without conditions and "update" it by adding null condition which results in a rule with empty/dummy condition.
After mapper refactoring that kind of test case should (and would) return null condition.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
mpichura: I've changed the name and javadoc of the method.

mpichura:
I added several tests which will slightly change (to 4xx error status instead of 500 now) as soon as mapper refactoring pt. 2 will come.

tpage-alfresco (approved):
LGTM

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -287,14 +293,173 @@ public void updateRuleFields()
...
+
+        final RestRuleModel updatedRule = restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet()
+                .updateRule(rule.getId(), rule);
+
+        restClient.assertStatusCodeIs(OK);
+        updatedRule.assertThat().isEqualTo(rule, ID)
+                .assertThat().field(ID).isNotNull();
+    }
+
+    /** Check we can use the POST response and update a rule rule without any conditions by adding null conditions. */
```
KikoUK, line 320: 2x `rule `

KikoUK (approved):

PR #1421, opened 2022-09-21
ACS-3291: Support for unlinking rule sets

gevagel:


remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -110,6 +111,22 @@ public RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId)
...
+        final NodeRef folderNodeRef = validator.validateFolderNode(folderNodeId,true);
+        final NodeRef ruleSetNodeRef = validator.validateRuleSetNode(ruleSetId, folderNodeRef);
+
+        //The folder should be linked to a rule set
+        if (!ruleService.isLinkedToRuleNode(folderNodeRef))
+        {
+            throw new InvalidArgumentException("The folder is not linked to a rule set.");
+        }
+
+        //The following line also handles the deletion of the child folder that gets created during linking
```
tpage-alfresco, line 126: Do you mean that it gets rid of the parent-child association?  Presumably it doesn't get rid of the rule folder.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetLinksRelation.java
```diff
@@ -67,6 +69,24 @@ public List<RuleSetLink> create(String nodeId, List<RuleSetLink> ruleSetLinksBod
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Remove link between a rule set and a folder rule for given rule set's and folder's node IDs.
+     * <p>
+     * - DELETE /nodes/{folderNodeId}/rule-set-links/{rulesetId}
```
tpage-alfresco, line 75:
```suggestion
     * - DELETE /nodes/{folderNodeId}/rule-set-links/{ruleSetId}
```

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
```diff
@@ -110,6 +111,22 @@ public RuleSetLink linkToRuleSet(String folderNodeId, String linkToNodeId)
...
+    @Override
+    public void unlinkRuleSet(String folderNodeId, String ruleSetId)
+    {
+        final NodeRef folderNodeRef = validator.validateFolderNode(folderNodeId,true);
+        final NodeRef ruleSetNodeRef = validator.validateRuleSetNode(ruleSetId, folderNodeRef);
+
+        //The folder should be linked to a rule set
+        if (!ruleService.isLinkedToRuleNode(folderNodeRef))
+        {
+            throw new InvalidArgumentException("The folder is not linked to a rule set.");
```
tpage-alfresco, line 123: This is interesting. I was expecting us to just return successfully if the link doesn't exist.  Would it be possible to check whether other DELETE endpoints can throw a 404?

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetLinksRelation.java
gevagel: Not sure what the suggestion is here

tpage-alfresco:
Looks like there are some test issues in RuleSetsImplTest and I had a question about the 404 error, but otherwise LGTM.

remote-api/src/main/java/org/alfresco/rest/api/nodes/NodeRuleSetLinksRelation.java
tpage-alfresco: :joy: You can just accept it :yum: 

I've suggested changing a lowercase `s` to an uppercase `S`.

mpichura (approved):
LGTM. Are you adding any E2E TAS REST to cover unlinking?

gevagel:
@mpichura I will raise another Jira to add E2Es after I merge this one, as E2Es weren't part of the scope of this ticket

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/RuleSetsImpl.java
gevagel: Looks like we're also throwing a 404 in the endpoint for deleting rules, when the rule doesn't exist.

tpage-alfresco: Ok - that's brilliant :+1:

tpage-alfresco (approved):
LGTM!

KikoUK:
Kind of late to the party but don't we need some e2es for this new functionality? 

If not part of this PR, we should raise a jira for adding e2es later. 

I'll have a look if we already have a jira for that and raise one if we don't 👍

PR #1425, opened 2022-09-21
ACS-3364 Add permission handling to linkedToBy.

tpage-alfresco:


tpage-alfresco:
This PR contains three significant changes:
1. Use admin in E2E test (previously the link wasn't being created in the test)
2. Check permissions in the service
3. Move the service method into RuleService.

PR #1426, opened 2022-09-21
[MNT-23118] Allow import with long paths

tiagosalvado10:


jottley (approved):

tpage-alfresco (approved):
I can't see any reason why the path needs to be limited to MAX_LENGTH either.  LGTM

PR #1431, opened 2022-09-22
ACS-3510: Rule Action updates E2E tests.

mpichura:
Adding simple update rule action E2E test. 
Most of rule action updates were already covered before (negative scenarios).
Rule action mapper refactoring will come in a separate PR soon.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
@@ -44,6 +44,7 @@ import static org.alfresco.utility.constants.UserRole.SiteCollaborator;
 import static org.alfresco.utility.report.log.Step.STEP;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.HttpStatus.FORBIDDEN;
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
 import static org.springframework.http.HttpStatus.OK;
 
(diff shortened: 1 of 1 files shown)


tpage-alfresco (approved):
LGTM

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/UpdateRulesTests.java
```diff
@@ -451,13 +451,40 @@ public void updateRuleWithConditionWithoutParamValueAndFail()
...
+        final RestActionBodyExecTemplateModel copyAction = createCustomActionModel("copy", copyParams);
+        final Map<String, Serializable> checkOutParams =
+                Map.of("destination-folder", "dummy-folder-node", "assoc-name", "cm:checkout", "assoc-type",
+                        "cm:contains");
+        final RestActionBodyExecTemplateModel checkOutAction = createCustomActionModel("check-out", checkOutParams);
+        final Map<String, Serializable> scriptParams = Map.of("script-ref", "dummy-script-node-id");
+        final RestActionBodyExecTemplateModel scriptAction = createCustomActionModel("script", scriptParams);
+        final RestRuleModel ruleModel = createRuleModelWithDefaultValues();
+        rule.setActions(Arrays.asList(copyAction, checkOutAction, scriptAction));
+        final RestRuleModel updatedRule = restClient.authenticateUser(user).withCoreAPI().usingNode(ruleFolder).usingDefaultRuleSet()
```
KikoUK, line 480: Since we have one big "Happy path" test for actions, I guess we can also add some of the "other options" to mix up with actions in the same test, if that makes sense.

mpichura: After our recent meeting and feedback from Stefan I added 2 new negative paths we found.
Anyway, I think that we should create a separate class for combined and thorough E2E REST tests of rules. 
I'd rather create a separate ticket for that as that could be a bigger effort which needs ideation and design work.

PR #1435, opened 2022-09-26
ACS-3589 Merge TAS CMIS into community repo.

tpage-alfresco:


packaging/tests/tas-cmis/pom.xml
```diff
@@ -1,27 +1,28 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.alfresco.tas</groupId>
-    <artifactId>alfresco-community-repo-cmis-test</artifactId>
-    <name>cmis test</name>
-    <packaging>jar</packaging>
-
+    <artifactId>cmis</artifactId>
```
tpage-alfresco, line 5: Currently I've used the existing group id and artifact id from the existing library project, but have updated the version to match community repo (i.e. alfresco-tas-cmis 1.34 will be followed by alfresco-tas-cmis 17.129). I felt this was the right decision because it will mean we don't need to change other places that use the library - Dependabot should automatically update them.

packaging/tests/tas-cmis/pom.xml
```diff
@@ -58,10 +59,30 @@
...
         <dependency>
             <groupId>org.alfresco.tas</groupId>
-            <artifactId>cmis</artifactId>
-            <scope>test</scope>
+            <artifactId>utility</artifactId>
+            <version>${tas.utility.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>mysql</groupId>
+                    <artifactId>mysql-connector-java</artifactId>
```
tpage-alfresco, line 76: I excluded this library since I don't think it's used and because the GPL license wasn't approved for release.  I think we use postgres almost everywhere unless we're explicitly testing another DB, so I didn't think this would be a problem.

dsibilio (approved):
The documentation of this project is dishonest at best, still mentioning Bamboo, Java 8, GitLab-specific emojis and what not. But I assume that could be tracked as a separate tech debt in case.

LGTM.

packaging/tests/tas-cmis/src/main/java/org/alfresco/cmis/AuthParameterProviderFactory.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+        return new AisAuthParameterProvider();
+    }
+
+    public Function<UserModel, Map<String, String>> getBasicProvider()
+    {
+        return new BasicAuthParameterProvider();
+    }
+
+    private class BasicAuthParameterProvider implements Function<UserModel, Map<String, String>>
```
pzhyland, line 61: Implementing `Function` seems _weird_. More idiomatic way would be to use a method reference when returning a function.

packaging/tests/tas-cmis/src/main/java/org/alfresco/cmis/AuthParameterProviderFactory.java
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @return Function which takes a {@link UserModel} and returns a map of
+     * authentication parameters to be used with {@link CmisWrapper#authenticateUser(UserModel, Function)}
+     */
+    public Function<UserModel, Map<String, String>> getDefaultProvider()
+    {
+        if (dataAIS.isEnabled())
+        {
+            STEP(String.format("%s Retrieved default AIS auth parameter provider.", STEP_PREFIX));
+            return new AisAuthParameterProvider();
```
pzhyland, line 42: You could use the `getAISProvider` to have a single source for instantiating this class.

pzhyland (approved):

PR #1438, opened 2022-09-26
ACS-3510 Rule mappers refactor pt3

mpichura:
This is presumably final part of refactoring/introducing mapper layer for Rules.
It roughly comprises to:
- getting rid of mapping logic in all model classes (now they are POJOs and have no corresponding test classes)
- adding last 2 mapper classes (for `Action` and `Rule`)
- covering all the above changes with unit tests
- small refactorings/fixes in E2E tests
- mapping "no-condition" service model rule condition to null for API model


remote-api/src/main/java/org/alfresco/rest/api/impl/mapper/rules/RestRuleModelMapper.java
```diff
@@ -0,0 +1,147 @@
...
+        if (serviceRule.getAction() != null)
+        {
+            builder.conditions(compositeConditionMapper.toRestModel(serviceRule.getAction().getActionConditions()));
+            if (serviceRule.getAction().getCompensatingAction() != null &&
+                    serviceRule.getAction().getCompensatingAction().getParameterValue(ScriptActionExecuter.PARAM_SCRIPTREF) != null)
+            {
+                builder.errorScript(
+                        serviceRule.getAction().getCompensatingAction().getParameterValue(ScriptActionExecuter.PARAM_SCRIPTREF).toString());
+            }
+            if (serviceRule.getAction() instanceof CompositeAction && ((CompositeAction) serviceRule.getAction()).getActions() != null)
```
tpage-alfresco, line 101: Is it possible that this isn't an instance of CompositeAction?  Maybe we should be logging something?

tpage-alfresco (approved):
LGTM

krdabrowski (approved):
LGTM

remote-api/src/main/java/org/alfresco/rest/api/impl/mapper/rules/RestRuleModelMapper.java
mpichura: In our current setup I don't think we can expect anything else than `CompositeAction`
I'll add logging and think we should log a ticket to do better logging.

PR #1457, opened 2022-09-29
ACS-3616: fix exception type for missing node id

gevagel:


remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -409,8 +409,7 @@ public NodeRef validateNode(StoreRef storeRef, String nodeId)
             nodeId = nodeId.substring(0, idx);
             if (versionLabel.equals("pwc"))
             {
-                // TODO correct exception?
-                throw new EntityNotFoundException(nodeId);
+                throw new InvalidArgumentException("Node with id: " + nodeId + "not found.");
```
tpage-alfresco, line 412: I'm usually in favour of fixing legacy code, but this exception (and TODO) has been here for over ten years!  Can we fix the issue with a slightly less scary change?  e.g. Catch the EntityNotFound and rethrow wrapped in an InvalidArgumentException?

I'd be interested to hear what others think about this too.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -400,7 +400,7 @@ public NodeRef validateNode(String nodeId)
     @Override
     public NodeRef validateNode(StoreRef storeRef, String nodeId)
     {
-        String versionLabel = null;
+        String versionLabel;
```
tpage-alfresco, line 403: Yeah, slightly better - or we could even define the variable where it's used.

krdabrowski (approved):

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
```diff
@@ -113,7 +113,7 @@ public void createRuleInNonExistentFolder()
         restClient.authenticateUser(user).withCoreAPI().usingNode(nonExistentFolder).usingDefaultRuleSet().createSingleRule(ruleModel);
 
         restClient.assertStatusCodeIs(NOT_FOUND);
-        restClient.assertLastError().containsSummary("fake-id was not found");
+        restClient.assertLastError().containsSummary("The entity with id: fake-id which is a folder was not found");
```
tpage-alfresco, line 116: These messages are really misleading like this. The entity with id "fake-id" is definitely _not_ a folder.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
krdabrowski: I totally agree with you Tom! Changing ten years old exceptions which some logic may be dependent on is very risky.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
gevagel: I've updated the messages after adding a new constructor to the Exception class. Also took the suggestion to also throw the original exception.

tpage-alfresco (approved):
LGTM

krdabrowski (approved):
Still LGTM.

PR #1466, opened 2022-10-05
ACS-3630 E2E tests for creating and updating rules with private actions.

tpage-alfresco:
Ensure that 403 Forbidden is returned rather than 500 Internal Server Error.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/access/V1AdminAccessRestrictionTest.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/access/V1AdminAccessRestrictionTest.java
@@ -1,20 +1,19 @@
 package org.alfresco.rest.actions.access;
 
-import org.alfresco.rest.RestTest;
-import org.alfresco.utility.model.FolderModel;
-import org.alfresco.utility.model.UserModel;
-import org.springframework.http.HttpStatus;
-import org.testng.annotations.BeforeClass;
-import org.testng.annotations.Test;
-
-import org.alfresco.rest.core.RestWrapper;
-import org.springframework.beans.factory.annotation.Autowired;
-
 import static org.alfresco.rest.actions.access.AccessRestrictionUtil.ERROR_MESSAGE_ACCESS_RESTRICTED;
 import static org.alfresco.rest.actions.access.AccessRestrictionUtil.MAIL_ACTION;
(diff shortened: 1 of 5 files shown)


krdabrowski (approved):

mpichura (approved):

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/access/V1AdminAccessRestrictionTest.java
```diff
@@ -45,7 +44,7 @@ public void userShouldNotExecuteMailAction() throws Exception {
                           .executeAction(MAIL_ACTION, testFolder, createMailParameters(adminUser, testUser));
 
         restClient.onResponse()
-                .assertThat().statusCode(HttpStatus.INTERNAL_SERVER_ERROR.value())
+                .assertThat().statusCode(HttpStatus.FORBIDDEN.value())
```
mstrankowski, line 47: Great :)

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RulesTestsUtils.java
```diff
@@ -162,6 +162,16 @@ public static RestRuleModel createVariousActions()
         return ruleModel;
     }
 
+    public static RestRuleModel createRuleWithPrivateAction()
+    {
+        RestActionBodyExecTemplateModel mailAction = new RestActionBodyExecTemplateModel();
+        mailAction.setActionDefinitionId("mail");
```
mstrankowski, line 168: I would consider either using values from AccessRestrictionUtil constants, or even refactoring AccessRestrictionUtil for your needs here so that action restriction logic could be kept in one place. Like calling createMailParameters. With your changes there are 2 Rule services coexisting and there seems to be logic in AccessRestrictionUtil for one service and for RulesTestsUtils in another, I don't know when/if the first one will be removed but might be worth to keep it in mind for cleanup.

tpage-alfresco: I think it makes sense to have the two helpers - one is for action restriction, the other is for v1 rule creation. They have some overlap, and in my latest change I've removed some of the duplication.

tpage-alfresco:
Merging this. The branch build isn't testing the thing that failed (TAS REST tests), and the PR build isn't being triggered presumably since I merged master to my branch and Travis thinks it must be the same build.

PR #1469, opened 2022-10-06
ACS-3572 V1 API for action constraints

mpichura:


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/ActionsTests.java
```diff
@@ -148,4 +151,71 @@ public void testGetActionDefinitionById() throws Exception
...
+    @Test(groups = {TestGroup.REST_API, TestGroup.ACTIONS, TestGroup.SANITY})
+    public void testGetSingleActionConstraint()
+    {
+        final UserModel testUser = dataUser.createRandomTestUser();
+        restClient.authenticateUser(testUser);
+
+        final String compareOperationsName = "ac-compare-operations";
+        final RestActionConstraintModel actionConstraintCompareOperations = restClient.
+                withCoreAPI().
+                usingActions().
```
tpage-alfresco, line 166: It looks weird to me to have the dots at the end of the lines - not at the start. I'm not sure we have any convention about it though.

mpichura: Thanks for catching that - fixing it.

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
```diff
@@ -0,0 +1,106 @@
...
+
+package org.alfresco.rest.api.model;
+
+import java.util.List;
+import java.util.Objects;
+
+import org.alfresco.service.Experimental;
+
+/**
+ * Representation of action parameter constraint.
```
tpage-alfresco, line 35: If I've understood correctly then this is just a specific type of constraint. It's basically where there is a list of possible values - like an enum.  If this is right then I think it would be useful to explain a bit more about it here.

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
```diff
@@ -0,0 +1,106 @@
...
+        public ConstraintData(final String value, final String label, final boolean isNode)
+        {
+            this.value = value;
+            this.label = label;
+            this.isNode = isNode;
+        }
+
+        private String value;
+        private String label;
+        private Boolean isNode;
```
tpage-alfresco, line 89: I don't know (and can't guess) how `isNode` is meant to work.  Please can we have javadoc for all three of these fields and the class itself?

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
mpichura: I've extended the javadoc to be (hopefully) more descriptive.

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.util.List;
+
+import org.alfresco.service.Experimental;
+
+/**
+ * Representation of action parameter constraint.
+ * Helps to constraint the list of allowable values for an action parameter.
+ * When action parameter has constraints defined (@see ActionDefinition.ParameterDefinition#getParameterConstraintName())
+ * they will be listed here.
```
tpage-alfresco, line 37: Will they definitely be listed? What if an action requires a parameter set to the node id of a folder with a particular aspect applied - will that constraint be listed here?  What about if it requires an even number?

I think this only handles a subset of constraints.

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
tpage-alfresco: I still don't understand this. How can `value` be the "Actual constraint value"?  Do you mean the parameter value?

mpichura: This class represents constraints for parameter values. So with sample action definition of:
```
"entry": {
                    "applicableTypes": [],
                    "parameterDefinitions": [
                        {
                            "name": "aspect-name",
                            "type": "d:qname",
                            "multiValued": false,
                            "mandatory": true,
                            "displayLabel": "Aspect",
                            "parameterConstraintName": "ac-aspects"
                        }
                    ],
                    "name": "add-features",
                    "trackStatus": false,
                    "description": "This will add an aspect to the matched item.",
                    "adhocPropertiesAllowed": true,
                    "id": "add-features",
                    "title": "Add aspect"
                }
```
We need to create an action with name `"add-features"` and exactly one parameter entry where the key of that entry is `"aspect-name"` and possible values are coming from the `value` field in list of `ActionParameterConstraint.ConstraintData` 
So this `value` can be treated as "possible parameter value" or parameter constraint. The `value` filed is there because we also provide display `label` and `isNode` indicator (for some of the constraints).
This class could possibly be avoided and the list of constraints could be List<String> which would be constraints / possible parameter values.
I am  happy to take some advice in here as I also see that this is a bit misleading.

remote-api/src/main/java/org/alfresco/rest/api/model/ActionParameterConstraint.java
tpage-alfresco: Ah - I get it. You're saying that when the constraint is available via the v1 action-definitions API then it will be listed here :+1:

repository/src/main/java/org/alfresco/repo/action/constraint/AspectParameterConstraint.java
```diff
@@ -1,34 +1,34 @@
...
- * 
- * You should have received a copy of the GNU Lesser General Public License
- * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
- * #L%
- */
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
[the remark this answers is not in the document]
mpichura: Thx, will update in all files.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/ActionsTests.java
```diff
@@ -148,4 +151,66 @@ public void testGetActionDefinitionById() throws Exception
...
+        final RestActionConstraintModel actionConstraintCompareOperations =
+                restClient.withCoreAPI().usingActions().getActionConstraintByName(compareOperationsName);
+
+        restClient.assertStatusCodeIs(HttpStatus.OK);
+
+        final RestActionConstraintModel expectedComparatorConstraints = new RestActionConstraintModel();
+        expectedComparatorConstraints.setConstraintName(compareOperationsName);
+        expectedComparatorConstraints.setConstraintValues(getComparatorConstraints());
+        actionConstraintCompareOperations.assertThat().isEqualTo(expectedComparatorConstraints);
+    }
```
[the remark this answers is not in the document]
mpichura: There were no `STEP()`s in the class at all previously - I can add them but that way to be consistent across the whole test class I'd have to add them to all methods...
Not sure if that's the best approach. WDYT?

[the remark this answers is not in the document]
krdabrowski: Hmm... I agree that you shouldn't be fixing all test methods in class. I'm happy when you are happy 😄

PR #1481, opened 2022-10-10
ACS-3649, ACS-3650 Action validation mechanism + validations against Action Definitions

mpichura:
Adding validations to be executed while mapping Rule Actions.
For now it covers:
- check if passed action definition for action exists
- check if all mandatory parameters are passed
- check if no extra parameters are passed
- check if parameter values fall under the constraints
Possibly need to add:
- check if parameter values are of expected type (but that should be somehow covered by `ActionParameterConverter` class)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
```diff
@@ -441,8 +441,74 @@ public void createRuleWithInvalidActionsShouldFail()
...
+        restClient.assertLastError().containsSummary(
+                String.format("Action parameter: %s has invalid value (%s). Look up possible values for constraint name %s",
+                        scriptRef, scriptNodeId, acScriptsConstraint));
+    }
+
+    /**
+     * Check we get error when attempt to create a rule without mandatory action parameters.
+     */
+    @Test(groups = {TestGroup.REST_API, TestGroup.RULES})
+    public void createRuleWithoutMandatoryActionParameterShouldFail()
```
tpage-alfresco, line 496: Test name and javadoc don't describe this test.

I think that they pretty much describe what's happening in `createRuleWithMissingActionParametersShouldFail`, so we don't need another test.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/ExecuteRulesTests.java
```diff
@@ -304,6 +305,7 @@ public void executeRules_privateFolderAsCollaborator()
     /**
      * Try to execute rule with broken action and receive 500.
      */
+    @Ignore("It shouldn't be possible to create a rule with broken action any more.")
```
tpage-alfresco, line 308: It's bound to be possible still. E.g. We can do loads of stuff by executing a script.  I can't think of a simple way to do this though, so maybe delete this test and create another ticket for us to work on in parallel to the other validation work.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
tpage-alfresco: Oh maybe we should have another test. I saw below that you have validation with a message `MISSING_PARAMETER` for this.

tpage-alfresco (approved):
Some pretty minor things, but LGTM.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/ExecuteRulesTests.java
mpichura: Will do, thx.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/CreateRulesTests.java
mpichura: I'll add a test to cover `MISSING_PARAMETER`. This one I renamed to be more clear about what the test is about.

tpage-alfresco (approved):
Looks better still :)

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/ExecuteRulesTests.java
mpichura: Added ACS-3699

krdabrowski: If this issue is fixed then this test case can be completely removed.

krdabrowski (approved):
Overall LGTM.

PR #1486, opened 2022-10-11
Handle XSS

rrajoria:


jottley (approved):
While this does appear to work and should do the job.  A better approach for the future would be to validate the parameters are of the expected type and format. This reduces the need for defensive coding.

MohinishSah (approved):

PR #1496, opened 2022-10-13
ACS-3651 Convert RulesTestsUtils to a service bean.

tpage-alfresco:
Update a few dummy folder nodes to use real folders.


mpichura (approved):
Subject to green travis build.

PR #1497, opened 2022-10-13
ACS-3651 Validate node pameters.

tpage-alfresco:


tpage-alfresco:
This checks the node exists and the user has read access.

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -56,13 +62,19 @@
     private final DictionaryService dictionaryService;
     private final ActionService actionService;
     private final NamespaceService namespaceService;
+    private final NodeService nodeService;
```
mpichura, line 65: I don't think this one is ever used.

remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -910,6 +910,9 @@
         <constructor-arg name="actionService" ref="ActionService"/>
         <constructor-arg name="dictionaryService" ref="DictionaryService"/>
         <constructor-arg name="namespaceService" ref="NamespaceService"/>
+        <constructor-arg name="nodeService" ref="NodeService"/>
```
mpichura, line 913: This one may be obsolete (nod used in the Java class).

mpichura (approved):
LGTM, some small cleanup needed.
Though I am surprised that E2E tests don't fail since this [PR ](https://github.com/Alfresco/alfresco-community-repo/pull/1496) is not yet merged which means that folders in rule actions are set as dummy node ids.  
Subject to green  build. Same (completely unrelated) annoying E2E tests failed in travis again :(

remote-api/src/main/resources/alfresco/public-rest-context.xml
tpage-alfresco: Thanks - good spot.

PR #1505, opened 2022-10-18
ACS-3788: Handling null/empty rule action parameter values.

mpichura:
This PR fixes 500 error returned when null/empty rule action parameter values are passed in POST rule request. 
This doesn't mean other (possibly 4xx) errors are not returned when empty/null values are passed. 
Generally empty (`""`) or null parameter value is treated as any other value passed.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
@@ -34,6 +34,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.alfresco.rest.api.Nodes;
 import org.alfresco.rest.framework.core.exceptions.EntityNotFoundException;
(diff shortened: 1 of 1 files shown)


remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
```diff
@@ -159,16 +160,17 @@ private Serializable convertValue(QName typeQName, Object propertyValue) throws
...
+            final String stringValue = Objects.toString(propertyValue, null);
+            if (typeQName.isMatch(DataTypeDefinition.QNAME) && typeQName.toString().contains(":"))
             {
-                value = QName.createQName(propertyValue.toString(), namespaceService);
+                value = QName.createQName(stringValue, namespaceService);
             }
             else if (typeQName.isMatch(DataTypeDefinition.NODE_REF))
             {
-                NodeRef nodeRef = nodes.validateOrLookupNode(propertyValue.toString(), null);
+                NodeRef nodeRef = nodes.validateOrLookupNode(stringValue, null);
```
tpage-alfresco, line 170: This seems a bit weird. For QNames then the method returns `null` but for NodeRefs then it throws an InvalidArgumentException.

tpage-alfresco: For the typeQName we have some generic validation that throws a NotFoundException if null is supplied. Maybe we could do something similar (with a different exception) for the propertyValue being null?

KikoUK (approved):

remote-api/src/main/java/org/alfresco/rest/api/impl/rules/ActionParameterConverter.java
mpichura: TBH, I am not sure what to do in case of null property and how to handle it. 
The case here is that even when we get empty String (`""`) as parameter value it is deserialized to `null` so actually we don't know if the client sent a request with null value or empty String value. I'll try to investigate if we can retrieve exactly what was sent.

mpichura: We seem to enforce empty `String`s to be deserialized to `null` by defining `RestApiStringDeserializer` and I understand it is a wider decision that we should put up with.

mpichura: I think we have 2 solutions here that could work in more generic way:
1. Even before we start mapping action parameters - we should completely remove those that have null values (which also means an empty String)
2. Always throw an error on null/empty parameter value (example message: "Action parameter <key> must not be null or empty").
I'd prefer second solution as that way we could "educate" API clients not to send unwanted stuff in requests.
WDYT?

mpichura:
I decided to throw exception every time when empty or null parameter value is passed (this will result in status 400 error)

tpage-alfresco (approved):
LGTM

PR #1519, opened 2022-10-24
[MNT-23158] Scripts limits configuration and optimization

tiagosalvado10:


jottley (approved):

repository/src/main/java/org/alfresco/repo/jscript/AlfrescoScriptThreadMxBeanWrapper.java
```diff
@@ -0,0 +1,46 @@
...
+package org.alfresco.repo.jscript;
+
+import java.lang.management.ManagementFactory;
+
+/**
+ * Allows to monitor memory usage
+ */
+public class AlfrescoScriptThreadMxBeanWrapper
+{
+    protected com.sun.management.ThreadMXBean threadMXBean; // Only Oracle / OpenJDK
```
jamalkm, line 35: Why can't we use `ManagementFactory` from the `java.lang` package? I mean `java.lang.management.ManagementFactory.getThreadMXBean();` We should not really write codes that call 'sun' packages.

repository/src/main/resources/alfresco/repository.properties
```diff
@@ -1351,3 +1351,18 @@ import.zip.compressionRatioThreshold=100
...
+scripts.execution.optimizationLevel=0
+
+# Max seconds a script is allowed to run
+scripts.execution.maxScriptExecutionSeconds=-1
+
+# Max call stack depth
+scripts.execution.maxStackDepth=-1
+
+# Max memory (bytes) a script can use
+scripts.execution.maxMemoryUsed=-1
```
jamalkm, line 1365: Same here, changing the name to `maxMemoryUsedInBytes` is more convenient and less error prone without reading the comment!

repository/src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
```diff
@@ -108,7 +109,24 @@ public class RhinoScriptProcessor extends BaseProcessor implements ScriptProcess
...
+    private int optimizationLevel = -1;
+
+    /** Maximum seconds a script is allowed to run */
+    private int maxScriptExecutionSeconds = -1;
+
+    /** Maximum of call stack depth (in terms of number of call frames) */
+    private int maxStackDepth = -1;
+
+    /** Maximum memory (bytes) a script can use */
+    private long maxMemoryUsed = -1L;
```
jamalkm, line 122: Changing the name to `maxMemoryUsedInBytes` is more convenient and less error prone without reading the comment!

repository/src/main/java/org/alfresco/repo/jscript/AlfrescoContextFactory.java
```diff
@@ -0,0 +1,206 @@
...
+    private int optimizationLevel = -1;
+    private int maxScriptExecutionSeconds = -1;
+    private int maxStackDepth = -1;
+    private long maxMemoryUsed = -1L;
+    private int observeInstructionCount = -1;
+
+    private AlfrescoScriptThreadMxBeanWrapper threadMxBeanWrapper;
+
+    private final int INTERPRETIVE_MODE = -1;
+
```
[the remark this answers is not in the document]
tiagosalvado10: Added `@Override` annotation to all methods that have been overridden.

repository/src/main/resources/alfresco/repository.properties
tiagosalvado10: Added `InBytes` suffix to every occurrence of `maxMemoryUsed` property.

repository/src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
tiagosalvado10: Added `InBytes` suffix to every occurrence of `maxMemoryUsed` property.

repository/src/main/java/org/alfresco/repo/jscript/AlfrescoScriptThreadMxBeanWrapper.java
jamalkm: I had a quick look and I don't think we can use `ThreadMXBean` from `java.lang` to get the allocated memory. However, the current implementation needs to be hardened as not all the JVM implementations support this which causes an exceptions to be thrown.
So first, we need to check whether the JVM supports thread memory allocation or not, then if it is enabled.

jottley: I am not sure on my reading of the interface but I think @tiagosalvado10's approach maybe the right now.  I do see there are two methods that could be used to test if 1/it is supported ([isThreadAllocatedMemorySupported](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.management/com/sun/management/ThreadMXBean.html#isThreadAllocatedMemorySupported())) and then 2/ if it is enabled ([isThreadAllocatedMemoryEnabled()](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.management/com/sun/management/ThreadMXBean.html#isThreadAllocatedMemoryEnabled())). However the class documentation states:

> Platform-specific management interface for the thread system of the Java virtual machine.
This platform extension is only available to a thread implementation that supports this extension.

So if it is not supported on the platform will the interface not be available? I think this might be something we need to verify if we can do the tests or will be get the ClassNotFoundException that is already being handled.

tiagosalvado10: I have delegated the checks to `AlfrescoScriptThreadMxBeanWrapper` class, handling the exceptions in case of interface not being available and also using the methods mentioned by Jared to determine if memory check is supported. 

I have performed tests against OpenJDK and Corretto and it has worked on both.

jamalkm (approved):
LGTM

PR #1535, opened 2022-11-03
MNT-22485: audit query with createdAt criteria returns correct totalI…

vrmoreira:
…tems

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/AuditImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/AuditImpl.java
@@ -295,7 +295,20 @@ public class AuditImpl implements Audit
         }
         else
         {
-            totalItems = hasMoreItems ? getAuditEntriesCountByApp(auditApplication) : totalRetrievedItems;
+            if (hasMoreItems) {
+                if (q != null) {
+                    // filtering via "where" clause
+                    AuditEntryQueryWalker propertyWalker = new AuditEntryQueryWalker();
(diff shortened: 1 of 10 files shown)


dsibilio (commented):
Compared to the previous implementation, we're now fetching data for each audit entry, temporarily storing it into a list then getting the total amount of entries by relying on the list size.

I feel like this approach may not be ideal as it could be dangerous in case this query returns a big number of audit entries, for the sole purpose of getting the count instead.

Wouldn't it be possible to do something more similar to the original implementation (e.g.: COUNT query via MyBatis as it was originally done - but with the appropriate filters instead, of course):
- https://github.com/Alfresco/alfresco-community-repo/blob/30892c28d12ea5602435dc235d966bea66a3e906/repository/src/main/java/org/alfresco/repo/domain/audit/ibatis/AuditDAOImpl.java#L227-L236
- https://github.com/Alfresco/alfresco-community-repo/blob/30892c28d12ea5602435dc235d966bea66a3e906/repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/audit-common-SqlMap.xml#L285-L292

**NOTE:** I also noticed that `entriesAudit` is being reassigned to the full list of audit entries, whereas before it was limited by the results of the paged query, is this a desired change?

vrmoreira:
@dsibilio , you're right on both situations, as usual:
1. I've changed the code to use myBatis only, without the ORM
2. that would change the behaviour, which is not the intended goal.

dsibilio (approved):
LGTM. Great work! 🎉

PR #1540, opened 2022-11-08
MNT-23276 - The null facet name should be considered

antoniojfelix:
This is the same issue of the previous PR https://github.com/Alfresco/alfresco-community-repo/pull/1534. I was disregarding the null value, but it turns out that the null value needs to be considered after all. The facet name should be the string "Null" in case of being null instead of being ignored.


pzhyland (approved):

tpage-alfresco (approved):
LGTM

PR #1565, opened 2022-11-18
ACS-3964: fix for non-inheritable rules still appearing in inherited rule sets

gevagel:
This PR introduces changes to avoid non-inheritable rules appearing in inherited rule sets. This is the fix to the following JIRA issue [https://alfresco.atlassian.net/browse/ACS-3964](https://alfresco.atlassian.net/browse/ACS-3964)

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/GetInheritedRulesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/GetInheritedRulesTests.java
@@ -25,6 +25,7 @@
  */
 package org.alfresco.rest.rules;
 
+import static org.alfresco.rest.requests.RuleSettings.IS_INHERITANCE_ENABLED;
 import static org.alfresco.utility.report.log.Step.STEP;
 import static org.testng.Assert.assertEquals;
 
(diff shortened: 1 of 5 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/GetInheritedRulesTests.java
```diff
@@ -91,6 +96,37 @@ public void getInheritedRules()
                       .and().entriesListCountIs(1);
     }
 
+    /**
+     * Check we only get the owned rules and no inherited rules in the child folder when parent rule isn't inheritable.
+     */
+    @Test (groups = { TestGroup.REST_API, TestGroup.RULES, TestGroup.SANITY })
```
KikoUK, line 102: Is this e2e correct for what's implemented? I think you should create a rule in the parent folder which has inherited set as false and then ensure that the rule is not inherited by the child. 

Also, why are you creating a rule in the child folder?

remote-api/src/test/java/org/alfresco/rest/api/impl/rules/RulesImplTest.java
```diff
@@ -141,6 +155,68 @@ public void testGetRules()
...
+        assertThat(rulesPage.getCollection().stream().findFirst().get()).isEqualTo(ruleMock);
+    }
+
+    @Test
+    public void testGetRules_rule_not_applied_to_children()
+    {
+        given(ruleSetMock.getInclusionType()).willReturn(InclusionType.INHERITED);
+        ruleModelInherited.applyToChildren(false);
+
+        //given(ruleSetMock.getInclusionType()).willReturn(InclusionType.OWNED);
```
krdabrowski, line 179: If this check is not necessary here, maybe it should be entirely removed?

krdabrowski (approved):

mpichura (commented):
Most of my remarks are minor and all of them consider test classes.

gevagel:
Thanks for the reviews. Comments have been addressed and PR should look better now

mpichura (approved):
LGTM now

KikoUK (approved):

PR #1569, opened 2022-11-21
Fix/mnt 22686 allow complex rest action params

kmagdziarz:


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/actions/email/EmailTemplateTest.java
```diff
@@ -0,0 +1,95 @@
...
+        testUser = dataUser.createRandomTestUser();
+        testSite = dataSite.usingUser(testUser)
+                           .createPublicRandomSite();
+        testFolder = dataContent.usingUser(testUser)
+                                .usingSite(testSite)
+                                .createFolder();
+    }
+
+    @Test
+    public void adminCanSendEmailUsingTemplateWithModelAsString() throws Exception
```
mstrankowski, line 48: I can see a test verifying if an admin can send this message, my assumption is that a user shouldn't be able to. Is that a wrong assumption, or is another test case needed?

Btw. I don't know if it's possible for the user to create the .ftl template, in this tests setup the admin seems to create it, but if the user can do it, then possibly creating it with lower priviledges user and testing vs admin or user would be better

PR #1577, opened 2022-11-24
ACS-4064: fix for linking from a folder with inherited rules

gevagel:
This PR introduces changes aiming to fix the bug described in this ticket [https://alfresco.atlassian.net/browse/ACS-4064](https://alfresco.atlassian.net/browse/ACS-4064).

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RuleSetLinksTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RuleSetLinksTests.java
@@ -46,6 +46,7 @@ import org.alfresco.utility.model.FolderModel;
 import org.alfresco.utility.model.SiteModel;
 import org.alfresco.utility.model.TestGroup;
 import org.alfresco.utility.model.UserModel;
+import org.springframework.http.HttpStatus;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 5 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/rules/RuleSetLinksTests.java
```diff
@@ -172,6 +173,56 @@ public void linkToRuleSet()
...
+    /**
+     * Check we can link to a rule set when linking from a folder which has inherited rules.
+     */
+    @Test(groups = {TestGroup.REST_API, TestGroup.RULES})
+    public void linkFromFolderWithInheritedRules()
+    {
+        STEP("Create folders");
+        final FolderModel parentFolder = dataContent.usingUser(user).usingSite(site).createFolder();
+        final FolderModel childFolder = dataContent.usingUser(user).usingResource(parentFolder).createFolder();
+        final FolderModel linkingFolder = dataContent.usingUser(user).usingSite(site).createFolder();
```
tpage-alfresco, line 185:
```suggestion
        final FolderModel linkedToFolder = dataContent.usingUser(user).usingSite(site).createFolder();
```

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
```diff
@@ -459,7 +459,13 @@ public boolean isRuleTypeEnabled(String ruleType)
     public boolean hasRules(NodeRef nodeRef)
     {
         return getRules(nodeRef).size() != 0;
-    } 
+    }
+
+    @Override
+    public boolean hasNonInheritedRules(NodeRef nodeRef)
```
KikoUK, line 465: Silly question: How does adding this method fix the issue?

mpichura (commented):
Can you please verify if this issue does not happen in Share? Jamal is not certain this is an unwanted behavior. We should be like for like with Share.

repository/src/main/java/org/alfresco/service/cmr/rule/RuleService.java
```diff
@@ -165,6 +165,15 @@
     @Auditable(parameters = {"nodeRef"})
     public boolean hasRules(NodeRef nodeRef);
 
+    /**
+     * Indicates whether the node in question has any non-inherited rules associated with it.
+     *
+     * @param nodeRef   the node reference
+     * @return          true if the node has rules associated, false otherwise
+     */
+    @Auditable(parameters = {"nodeRef"})
```
tpage-alfresco, line 174: `@Experimental`?

tpage-alfresco:
> Can you please verify if this issue does not happen in Share? Jamal is not certain this is an unwanted behavior. We should be like for like with Share.

This is unwanted behaviour. Share certainly allows nodes that link to a rule set and inherit other rule sets.  I don't know for sure if it allows this sequence of events, but I'm pretty sure that it does.

repository/src/main/java/org/alfresco/repo/rule/RuleServiceImpl.java
gevagel: Inside the method that does the actual linking `linkToRuleSet`, we are checking whether the folder that is trying to link has any pre-existing rules. With this method, we will ignore inherited rules in that check and perform the linking

TomikaArome:
> Can you please verify if this issue does not happen in Share? Jamal is not certain this is an unwanted behavior. We should be like for like with Share.

I can confirm Share allows you to link a rule set to a folder with inherited rules already. I checked that before raising the bug

tpage-alfresco (approved):
LGTM

krdabrowski (approved):

PR #1592, opened 2022-12-02
[MNT-23396] [MNT-23391] Change call stack depth limit to be only applied to custom scripts / Clean scope only for custom scripts

tiagosalvado10:


repository/src/main/java/org/alfresco/repo/jscript/AlfrescoContextFactory.java
```diff
@@ -137,9 +128,40 @@ protected Object doTopCall(Callable callable, Context cx, Scriptable scope, Scri
...
+                acx.setOptimizationLevel(INTERPRETIVE_MODE);
+                acx.setMaximumInterpreterStackDepth(maxStackDepth);
+            }
+        }
+        else
+        {
+            if (acx.getMaximumInterpreterStackDepth() != Integer.MAX_VALUE)
+            {
+                LOGGER.debug("Max call stack depth limit will be set to default value: " + Integer.MAX_VALUE);
+                acx.setOptimizationLevel(INTERPRETIVE_MODE);
```
nbarithel, line 158: setting OptimizationLevel twice, here to INTERPRETIVE_MODE and two lines below to optimizationLevel

tiagosalvado10: @nbarithel the reason behind setting the optimization level twice is to prevent an exception to be thrown when changing the max stack depth to the default value. Please, see [here](https://github.com/mozilla/rhino/blob/0131b5c7e4b3924241e0a6a7eb2d52ed7e888f89/src/org/mozilla/javascript/Context.java#L1986).

nbarithel: Indeed, thanks for the link, it wasn't obvious from reading the code! Although, it seems that the maximum stack depth setting will still not work if the optimization level is set to something different from -1 afterwards, according to the [javadoc](https://github.com/mozilla/rhino/blob/0131b5c7e4b3924241e0a6a7eb2d52ed7e888f89/src/org/mozilla/javascript/Context.java#L1974).

nbarithel: Ok, now I get it, you just wanted to reset the default value of the maximumInterpreterStackDepth 👍

jottley (approved):

jamalkm (approved):

PR #1609, opened 2022-12-08
Feature/aca 4619 self service user password reset adw

MohinishSah:


kavitshah-gl (approved):
Changes are fine.

repository/src/main/java/org/alfresco/repo/security/authentication/ResetPasswordServiceImpl.java
```diff
@@ -373,7 +373,7 @@ public void sendResetPasswordEmail(DelegateExecution execution, String fallbackE
 
         final ClientApp clientApp = getClientAppConfig(clientName);
         Map<String, Serializable> emailTemplateModel = Collections.singletonMap(FTL_RESET_PASSWORD_URL,
-                    createResetPasswordUrl(clientApp, id, key));
+                    createResetPasswordUrl(clientApp, id, key, userName));
```
jamalkm, line 376: Why you need to pass the userName? The username is defined in the workflow properties. 
Probably I'm missing something here, but how this method `sendResetPasswordEmail` supposed to be called? this method will not work if you don't start the workflow first (via `requestReset` method), and there you'll see that the username is added to the workflow properties.

So, based on the ACA-4619 ticket use cases:
- As a user i would like to change my password
  - I don't think this case can be covered by this service. As I mentioned you need to start the workflow first...
- As a user i would like to request to change my password at login  'forgot my password' 
  - In this case, the username is already provided and there is no need to add the username explicitly.

repository/src/main/java/org/alfresco/util/UrlUtil.java
```diff
@@ -146,4 +146,38 @@ else if ("https".equals(protocol) && port == 443)
...
+    /**
+     * Replaces the adw url placeholder, namely {@literal ${alfrescoUrl}}, with <b>adw</b> url.
+     *
+     * @param value          the string value which contains the share url placeholder
+     * @param sysAdminParams the {@code SysAdminParams} object
+     * @return if the given {@code value} contains share url placeholder,
+     * the placeholder is replaced with share url; otherwise, the given {@code value} is simply returned
+     */
+
+    public static String replaceAdwUrlPlaceholder(String value, SysAdminParams sysAdminParams)
```
jamalkm, line 174: We need to make this generic. So let's rename it to `replaceRepoBaseUrlPlaceholder`. See my comment above.

repository/src/main/java/org/alfresco/repo/security/authentication/ResetPasswordServiceImpl.java
```diff
@@ -517,27 +532,33 @@ protected String getConfirmResetPasswordEmailTemplate(ClientApp clientApp)
...
-            sb.append(UrlUtil.getShareUrl(sysAdminParams));
 
-            LOGGER.warn("'resetPasswordPageUrl' property is not set for the client [" + clientApp.getName()
-                        + "]. The default base url of Share will be used [" + sb.toString() + "]");
-        }
-        else
-        {
-            // We pass an empty string as we know that the pageUrl is not null
-            sb.append(getUrl(pageUrl, ""));
+        if(clientApp.getName().equals("adw")) {
```
jamalkm, line 541: Apart from missing Alfresco code standard format, we shouldn't check for the specific registered-client. The client name is provided by the REST API caller, the API makes sure that the client exists, therefore, we should get all the required info from `ClientApp` object.

repository/src/main/java/org/alfresco/repo/security/authentication/ResetPasswordServiceImpl.java
```diff
@@ -504,6 +504,21 @@ private String getUrl(String url, String propName)
...
+        {
+            LOGGER.warn("The url for the property [" + propName + "] is not configured.");
+            return "";
+        }
+
+        if (url.endsWith("/"))
+        {
+            url = url.substring(0, url.length() - 1);
+        }
+        return UrlUtil.replaceAdwUrlPlaceholder(url, sysAdminParams);
```
jamalkm, line 519: gets replaced with the generic method from the UrlUtil

repository/src/main/java/org/alfresco/repo/security/authentication/ResetPasswordServiceImpl.java
```diff
@@ -517,27 +532,33 @@ protected String getConfirmResetPasswordEmailTemplate(ClientApp clientApp)
...
-        }
-        else
-        {
-            // We pass an empty string as we know that the pageUrl is not null
-            sb.append(getUrl(pageUrl, ""));
+        if(clientApp.getName().equals("adw")) {
+            sb.append(getAdwUrl(pageUrl, ""));
+            LOGGER.warn("Client Name is " + clientApp.getName() + " The url used is     " + sb.toString());
+            sb.append("?key=").append(key)
+                    .append("&id=").append(BPMEngineRegistry.createGlobalId(ActivitiConstants.ENGINE_ID, id)).append("&userName=").append(userName);
```
jamalkm, line 545: As I mentioned earlier, I don't think the userName is required.

repository/src/main/java/org/alfresco/repo/security/authentication/ResetPasswordServiceImpl.java
```diff
@@ -517,27 +532,33 @@ protected String getConfirmResetPasswordEmailTemplate(ClientApp clientApp)
...
+
         }
+        else {
+            if (StringUtils.isEmpty(pageUrl)) {
```
jamalkm, line 549: `else {
  if {` should be replace with `else if ()`

jamalkm (changes requested):
From the code change, it seems we need to have a meeting to go over the requirements and resolve any misunderstanding.

jamalkm (changes requested):

jamalkm (changes requested):
Cloud you please add tests. I don't see any tests with this PR.

kavitshah-gl (approved):
This PR can be merged since the FE is dependent on these changes.

jamalkm:
This should be reverted as it is in no shape to be part of the product. FE can work with a mock service.

PR #1617, opened 2022-12-12
ACS-4031 list category children

mpichura:
This PR introduces a `GET: /categories/{categoryId}/subcategories` endpoint to list direct child categories of a given parent category.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/GetCategoriesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/GetCategoriesTests.java
@@ -28,11 +28,20 @@ package org.alfresco.rest.categories;
 
 import static org.alfresco.utility.report.log.Step.STEP;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
+import static org.springframework.http.HttpStatus.CREATED;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
 import static org.springframework.http.HttpStatus.OK;
+import static org.testng.Assert.assertTrue;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.alfresco.rest.RestTest;
+import org.alfresco.rest.core.RestResponse;
 import org.alfresco.rest.model.RestCategoryModel;
+import org.alfresco.rest.model.RestCategoryModelsCollection;
+import org.alfresco.utility.data.RandomData;
 import org.alfresco.utility.model.FolderModel;
 import org.alfresco.utility.model.SiteModel;
 import org.alfresco.utility.model.TestGroup;
@@ -42,6 +51,8 @@ import org.testng.annotations.Test;
 
 public class GetCategoriesTests extends RestTest
 {
+    private static final List<String> DEFAULT_ROOT_CATEGORIES = List.of("Software Document Classification", "Languages", "Regions", "Tags");
+
     private UserModel user;
 
     @BeforeClass(alwaysRun = true)
@@ -57,11 +68,29 @@ public class GetCategoriesTests extends RestTest
     @Test(groups = {TestGroup.REST_API})
     public void testGetCategoryById()
     {
-        STEP("Get category with -root- as id (which does not exist)");
+        STEP("Create a category under root category (as admin)");
         final RestCategoryModel rootCategory = new RestCategoryModel();
         rootCategory.setId("-root-");
-        restClient.authenticateUser(user).withCoreAPI().usingCategory(rootCategory).getCategory();
-        restClient.assertStatusCodeIs(NOT_FOUND);
+        final RestCategoryModel aCategory = new RestCategoryModel();
+        aCategory.setName(RandomData.getRandomName("Category"));
+        final RestCategoryModel createdCategory = restClient.authenticateUser(dataUser.getAdminUser())
+                .withCoreAPI()
+                .usingCategory(rootCategory)
+                .createSingleCategory(aCategory);
+        restClient.assertStatusCodeIs(CREATED);
+
+        createdCategory.assertThat()
+                .field("name").is(aCategory.getName());
+        createdCategory.assertThat()
+                .field("parentId").is(rootCategory.getId());
+        createdCategory.assertThat()
+                .field("hasChildren").is(false);
+
+        STEP("Get the created category (as regular user)");
+        final RestCategoryModel categoryFromGet =
+                restClient.authenticateUser(user).withCoreAPI().usingCategory(createdCategory).getCategory();
+        restClient.assertStatusCodeIs(OK);
+        categoryFromGet.assertThat().isEqualTo(createdCategory);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/GetCategoriesTests.java
```diff
@@ -91,7 +121,87 @@ public void testGetCategoryByIdProvidingFolderAsId()
...
+
+        STEP("Get category children with folder id passed as parent id");
+        final RestCategoryModel parentCategory = new RestCategoryModel();
+        parentCategory.setId(folder.getNodeRef());
+        restClient.authenticateUser(user).withCoreAPI().usingCategory(parentCategory).getCategoryChildren();
+        restClient.assertStatusCodeIs(BAD_REQUEST).assertLastError().containsSummary("Node id does not refer to a valid category");
+    }
+
+    /**
+     * Check we get an error when passing folder node id as parent category id when getting children.
```
tpage-alfresco, line 194:
```suggestion
     * Check we get an error when passing a non-existent node id as parent category id when getting children.
```

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/GetCategoriesTests.java
```diff
@@ -91,7 +121,87 @@ public void testGetCategoryByIdProvidingFolderAsId()
...
+
+        childCategoriesList = restClient.authenticateUser(user).withCoreAPI().usingCategory(rootCategory).getCategoryChildren();
+        restClient.assertStatusCodeIs(OK);
+        assertTrue(childCategoriesList.getEntries().stream()
+                .map(RestCategoryModel::onModel)
+                .map(RestCategoryModel::getId)
+                .collect(Collectors.toList())
+                .contains(createdCategory.getId()));
+
+        STEP("Create 2 more categories under newCategoryUnderRoot and make sure it is returned as children");
```
tpage-alfresco, line 163:
```suggestion
        STEP("Create 2 more categories under newCategoryUnderRoot and make sure they are returned as children");
```

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -106,6 +102,37 @@ public List<Category> createSubcategories(String parentCategoryId, List<Category
...
+    @Override
+    public CollectionWithPagingInfo<Category> getCategoryChildren(String parentCategoryId, Parameters params)
+    {
+        final NodeRef parentNodeRef = getParentNodeRef(parentCategoryId);
+        final List<ChildAssociationRef> childCategoriesAssocs = nodeService.getChildAssocs(parentNodeRef).stream()
+                        .filter(ca -> ContentModel.ASSOC_SUBCATEGORIES.equals(ca.getTypeQName()))
+                        .collect(Collectors.toList());
+        final List<Category> categories = childCategoriesAssocs.stream()
+                .map(c -> mapToCategory(c.getChildRef()))
+                .collect(Collectors.toList());
```
tpage-alfresco, line 114: I think most of this logic should go (or already be present) in the category service. I definitely think that using the node service is the right approach though.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -106,6 +102,37 @@ public List<Category> createSubcategories(String parentCategoryId, List<Category
...
+    {
+        final NodeRef nodeRef = nodes.validateNode(nodeId);
+        if (isNotACategory(nodeRef))
+        {
+            throw new InvalidArgumentException(NOT_A_VALID_CATEGORY, new String[]{nodeId});
+        }
+        return nodeRef;
+    }
+
+    private NodeRef getParentNodeRef(String parentCategoryId)
```
tpage-alfresco, line 128: I assume you didn't call this `getCategoryNode` because the root isn't a category? Perhaps it would be clearer being called `getCategoryHierarchyNode`? Or alternatively if you just added some javadoc to explain the purpose (in particular I'd expect `getParentNodeRef(id)` to give me the parent of the supplied node.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
```diff
@@ -366,12 +371,149 @@ public void testCreateCategories_nonExistingParentNode()
...
+        then(nodesMock).should().validateNode(PARENT_ID);
+        then(nodesMock).should().isSubClass(parentCategoryNodeRef, ContentModel.TYPE_CATEGORY, false);
+        then(nodesMock).shouldHaveNoMoreInteractions();
+        then(nodeServiceMock).should().getChildAssocs(parentCategoryNodeRef);
+        then(nodeServiceMock).shouldHaveNoMoreInteractions();
+
+        then(authorityServiceMock).shouldHaveNoInteractions();
+        then(categoryServiceMock).shouldHaveNoInteractions();
+
+        assertEquals(0, categoryChildren.getTotalItems().intValue());
```
tpage-alfresco, line 467: Is it worth also checking the actual list is empty? It looks like the `totalItems` is pretty independent of the `collection`.

tpage-alfresco (commented):
I had a few questions, but I think the most interesting discussion is whether the logic should be in the category service or CategoriesImpl.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
mpichura: I had trouble to name this method so that it is not confusing.
The general rule here is:
- `getParentNodeRef` is called when we start off and pass parent category id in the call (if that id is not `-root` then it simply returns `getCategoryNodeRef`)
- `getCategoryNodeRef` is called when we start off and pass category id in the call (for now it is for get single category)
Maybe I could merge these methods into one, WDYT?

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
```diff
@@ -387,4 +529,37 @@ private List<Category> prepareCategories()
...
+        given(nodeServiceMock.getPrimaryParent(childRef)).willReturn(parentAssoc);
+        given(nodeServiceMock.getParentAssocs(parentRef)).willReturn(List.of(parentAssoc));
+    }
+
+    private void assertCategoryFields(final Category category, final int index, final String parentId)
+    {
+        assertEquals(CATEGORY_ID + "-" + index, category.getId());
+        assertEquals(CATEGORY_NAME + "-" + index, category.getName());
+        assertEquals(parentId, category.getParentId());
+        assertFalse(category.getHasChildren());
```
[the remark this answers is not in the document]
mpichura: Good point - refactoring that.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
mpichura: Good point - adding that.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
mpichura: Actually that `getParentNodeRef(id)` was meant to handle a case when `-root-` is explicitly passed as a parent category id.
I am pushing refactored code along with some javadocs - hopefully this will clarify this.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
mpichura: I have mixed feelings about usage of `CategoryService` interface here. 
To me it looks it is mentioned for search purposes (it's even placed in `org.alfresco.service.cmr.search` package).

I'd be tempted to create a middle-tier category related service (don't have a good name for it, maybe `CategoryNodeService`) which could take most of the logic which is now put in `CategoriesImpl` but would restrain itself to DB node operations. 
If we'd need 'old' methods or search related queries, we could always call the 'old' `CategoryService` (although it already is a mix of search and DB nodes stuff).
We definitely need to discuss that soon.

krdabrowski (approved):

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
tpage-alfresco: Yes - that sounds like a good solution.

PR #1621, opened 2022-12-13
ACS-4029: Update a category name - PUT /categories/{categoryId}

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-4029

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
@@ -1,0 +1,87 @@
+/*
+ * #%L
+ * Alfresco Remote API
+ * %%
+ * Copyright (C) 2005 - 2022 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+
+package org.alfresco.rest.categories;
+
+import static org.alfresco.utility.data.RandomData.getRandomName;
+import static org.springframework.http.HttpStatus.CREATED;
+
+import org.alfresco.rest.RestTest;
+import org.alfresco.rest.model.RestCategoryModel;
+
+abstract class CategoriesRestTest extends RestTest
+{
+
+    public static final String ROOT_CATEGORY_ID = "-root-";
+    public static final String FIELD_NAME = "name";
+
+    protected RestCategoryModel prepareCategoryUnderRoot()
+    {
+        return prepareCategoryUnder(ROOT_CATEGORY_ID);
+    }
+
+    protected RestCategoryModel prepareCategoryUnder(final String parentId)
+    {
+        final RestCategoryModel parentCategory = createCategoryModelWithId(parentId);
+        final RestCategoryModel categoryModel = createCategoryModelWithName(getRandomName("CategoryName"));
+        final RestCategoryModel createdCategory = restClient.authenticateUser(dataUser.getAdminUser())
+            .withCoreAPI()
+            .usingCategory(parentCategory)
+            .createSingleCategory(categoryModel);
+        restClient.assertStatusCodeIs(CREATED);
+
+        return createdCategory;
+    }
+
+    protected RestCategoryModel createCategoryModelWithId(final String id)
+    {
+        return createCategoryModelWithIdAndName(id, null);
+    }
+
+    protected RestCategoryModel createCategoryModelWithName(final String name)
+    {
+        return createCategoryModelWithIdAndName(null, name);
+    }
+
+    protected RestCategoryModel createCategoryModelWithIdAndName(final String id, final String name)
+    {
+        final RestCategoryModel categoryModel = new RestCategoryModel();
+        categoryModel.setName(name);
+        categoryModel.setId(id);
+
+        return categoryModel;
+    }
+
+    protected RestCategoryModel createCategoryModelOf(final String name, final RestCategoryModel originalCategory)
+    {
+        final RestCategoryModel categoryModel = createCategoryModelWithName(name);
+        categoryModel.setId(originalCategory.getId());
+        categoryModel.setParentId(originalCategory.getParentId());
+        categoryModel.setHasChildren(originalCategory.getHasChildren());
+
+        return categoryModel;
+    }
+}
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
@@ -27,16 +27,31 @@
 package org.alfresco.rest.categories;
 
 import static org.alfresco.utility.data.RandomData.getRandomName;
+import static org.alfresco.utility.report.log.Step.STEP;
 import static org.springframework.http.HttpStatus.CREATED;
 
 import org.alfresco.rest.RestTest;
 import org.alfresco.rest.model.RestCategoryModel;
+import org.alfresco.utility.model.UserModel;
+import org.testng.annotations.BeforeClass;
 
 abstract class CategoriesRestTest extends RestTest
 {
+    protected static final String ROOT_CATEGORY_ID = "-root-";
+    protected static final String CATEGORY_NAME_PREFIX = "CategoryName";
+    protected static final String FIELD_NAME = "name";
+    protected static final String FIELD_ID = "id";
+    protected static final String FIELD_PARENT_ID = "parentId";
+    protected static final String FIELD_HAS_CHILDREN = "hasChildren";
 
-    public static final String ROOT_CATEGORY_ID = "-root-";
-    public static final String FIELD_NAME = "name";
+    protected UserModel user;
+
+    @BeforeClass(alwaysRun = true)
+    public void dataPreparation() throws Exception
+    {
+        STEP("Create a user");
+        user = dataUser.createRandomTestUser();
+    }
 
     protected RestCategoryModel prepareCategoryUnderRoot()
     {
@@ -46,7 +61,7 @@ abstract class CategoriesRestTest extends RestTest
     protected RestCategoryModel prepareCategoryUnder(final String parentId)
     {
         final RestCategoryModel parentCategory = createCategoryModelWithId(parentId);
-        final RestCategoryModel categoryModel = createCategoryModelWithName(getRandomName("CategoryName"));
+        final RestCategoryModel categoryModel = createCategoryModelWithName(getRandomName(CATEGORY_NAME_PREFIX));
         final RestCategoryModel createdCategory = restClient.authenticateUser(dataUser.getAdminUser())
             .withCoreAPI()
             .usingCategory(parentCategory)
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
@@ -68,20 +68,9 @@ abstract class CategoriesRestTest extends RestTest
 
     protected RestCategoryModel createCategoryModelWithIdAndName(final String id, final String name)
     {
-        final RestCategoryModel categoryModel = new RestCategoryModel();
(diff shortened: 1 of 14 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+     * Try to update a non-existing category and receive 404 (Not Found)
+     */
+    @Test(groups = { TestGroup.REST_API})
+    public void testUpdateCategory_usingNonExistingCategoryAndExpect404()
+    {
+        STEP("Create a fake parent category");
+        final RestCategoryModel nonExistingCategory = createCategoryModelWithIdAndName(randomUUID().toString(), getRandomName("CategoryName"));
```
mpichura, line 124: Is it possible (I suppose it is but still extremely unlikely) that `randomUUID()` will return an id of a node existing in ACS instance and this way we can have a different return status (possibly 400 instead of 404)?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        return categoryModel;
+    }
+
+    protected RestCategoryModel createCategoryModelOf(final String name, final RestCategoryModel originalCategory)
+    {
+        final RestCategoryModel categoryModel = createCategoryModelWithName(name);
+        categoryModel.setId(originalCategory.getId());
+        categoryModel.setParentId(originalCategory.getParentId());
+        categoryModel.setHasChildren(originalCategory.getHasChildren());
```
tpage-alfresco, line 83: It might make sense to start introducing a builder pattern for the TAS POJOs similar to what we have for the REST API POJOs.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -106,17 +105,36 @@ public List<Category> createSubcategories(String parentCategoryId, List<Category
...
+        verifyAdminAuthority();
+        final NodeRef categoryNodeRef = nodes.validateOrLookupNode(id, null);
+        if (isNotACategory(categoryNodeRef) || isRootCategory(categoryNodeRef))
+        {
+            throw new InvalidArgumentException(NOT_A_VALID_CATEGORY, new String[]{id});
         }
-        return categoryService.createCategory(parentNodeRef, c.getName());
+
+        final Category originalCategory = mapToCategory(categoryNodeRef);
+        compareAndVerifyCategoryFields(fixedCategoryModel, originalCategory);
```
mpichura, line 119: As I commented in the `Categories` interface, I believe we should either say that we ignore all fields other than `name` or that we will return error (400) when any fields other than `name` are passed in the request so I think this check should either not be triggered at all or just verify if there are any fields other than `name` (if so - we should call the same check in create category).

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CreateCategoriesTests.java
```diff
@@ -100,7 +100,7 @@ public void testCreateCategoryWithoutName_andFail()
                 .withCoreAPI()
                 .usingCategory(rootCategory)
                 .createSingleCategory(aCategory);
-        restClient.assertStatusCodeIs(BAD_REQUEST).assertLastError().containsSummary("Category name must not be null or empty");
+        restClient.assertStatusCodeIs(BAD_REQUEST).assertLastError().containsSummary("Category name must not be null nor empty");
```
tpage-alfresco, line 103: I would say "Category name must not be null or empty" but I don't know which is correct grammar. If I was using "nor" I would say something like "Category name must not be null, nor must it be empty."

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
```diff
@@ -0,0 +1,235 @@
...
+        final RestCategoryModel fixedCategoryModel = createCategoryModelOf(newName, createdCategory);
+        final RestCategoryModel updatedCategory = restClient.authenticateUser(dataUser.getAdminUser())
+            .withCoreAPI()
+            .usingCategory(createdCategory)
+            .updateCategory(fixedCategoryModel);
+
+        restClient.assertStatusCodeIs(OK);
+        updatedCategory.assertThat().isEqualTo(createdCategory, IGNORE_FIELD_NAME);
+        updatedCategory.assertThat().field(FIELD_NAME).isNot(createdCategory.getName());
+        updatedCategory.assertThat().field(FIELD_NAME).is(newName);
```
tpage-alfresco, line 69: Slightly redundant checks here, but maybe it makes the messages nicer if the obvious thing goes wrong.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
```diff
@@ -366,19 +379,165 @@ public void testCreateCategories_nonExistingParentNode()
...
+        given(nodesMock.getNode(any())).willReturn(prepareCategoryNode());
+        given(nodeServiceMock.getPrimaryParent(any())).willReturn(parentAssociation);
+
+        // when
+        assertThatExceptionOfType(InvalidArgumentException.class).isThrownBy(() -> objectUnderTest.updateCategoryById(CATEGORY_ID, categoryWithInvalidParentId))
+            .withMessageContaining(FIELD_NOT_MATCH, "parentId");
+    }
+
+    @Test
+    public void testUpdateCategoryById_notMatchingHasChildrenField()
```
mpichura, line 513: As I commented in the tested class - I believe we should either ignore all fields other than `name` or shouldn't allow any additional fields.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
```diff
@@ -366,19 +379,165 @@ public void testCreateCategories_nonExistingParentNode()
...
+        given(nodesMock.getNode(any())).willReturn(prepareCategoryNode());
+        given(nodeServiceMock.getPrimaryParent(any())).willReturn(parentAssociation);
+
+        // when
+        assertThatExceptionOfType(InvalidArgumentException.class).isThrownBy(() -> objectUnderTest.updateCategoryById(CATEGORY_ID, categoryWithInvalidId))
+            .withMessageContaining(FIELD_NOT_MATCH, "id");
+    }
+
+    @Test
+    public void testUpdateCategoryById_notMatchingParentIdField()
```
mpichura, line 497: As I commented in the tested class - I believe we should either ignore all fields other than name or shouldn't allow any additional fields.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
```diff
@@ -366,19 +379,165 @@ public void testCreateCategories_nonExistingParentNode()
...
+        given(nodesMock.getNode(any())).willReturn(prepareCategoryNode());
+        given(nodeServiceMock.getPrimaryParent(any())).willReturn(parentAssociation);
+
+        // when
+        assertThatExceptionOfType(InvalidArgumentException.class).isThrownBy(() -> objectUnderTest.updateCategoryById(CATEGORY_ID, categoryWithoutName))
+            .withMessageContaining(NOT_NULL_NOR_EMPTY);
+    }
+
+    @Test
+    public void testUpdateCategoryById_notMatchingIdField()
```
mpichura, line 481: As I commented in the tested class - I believe we should either ignore all fields other than name or shouldn't allow any additional fields.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
```diff
@@ -0,0 +1,235 @@
...
+        restClient.authenticateUser(dataUser.getAdminUser())
+            .withCoreAPI()
+            .usingCategory(createdCategory)
+            .updateCategory(fixedCategoryModel);
+
+        restClient.assertStatusCodeIs(BAD_REQUEST);
+    }
+
+    /**
+     * Try to update a category using not matching ID in path and receive 400 (Bad Request)
```
tpage-alfresco, line 197: This sounds like it should be a 404.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
```diff
@@ -0,0 +1,235 @@
...
+        restClient.authenticateUser(dataUser.getAdminUser())
+            .withCoreAPI()
+            .usingCategory(notRelatedCategory)
+            .updateCategory(fixedCategoryModel);
+
+        restClient.assertStatusCodeIs(BAD_REQUEST);
+    }
+
+    /**
+     * Try to update a category using not matching ID in request body and receive 400 (Bad Request)
```
tpage-alfresco, line 217: I'm pretty sure that we shouldn't validate the id in the body. It should be allowed to be there, but get ignored. I'm pretty sure this is the behaviour for other PUT endpoints.

mpichura (commented):
Overall - very good job, however, I have questions regarding this implementation versus our OPRN API design (BTW I think I've found one error in our design and will prepare a fix shortly)

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -106,17 +105,36 @@ public List<Category> createSubcategories(String parentCategoryId, List<Category
...
+        return mapToCategory(changeCategoryName(categoryNodeRef, fixedCategoryModel.getName()));
     }
 
-    private boolean isNotACategory(NodeRef nodeRef)
+    private void verifyAdminAuthority()
     {
-        return !nodes.isSubClass(nodeRef, ContentModel.TYPE_CATEGORY, false);
+        if (!authorityService.hasAdminAuthority())
+        {
+            throw new PermissionDeniedException(NO_PERMISSION_TO_CREATE_A_CATEGORY);
```
tpage-alfresco, line 128: I guess we should modify this message to be something like "No permission to manage categories."

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -106,17 +105,36 @@ public List<Category> createSubcategories(String parentCategoryId, List<Category
...
+        }
+    }
+
+    private NodeRef createCategoryNodeRef(NodeRef parentNodeRef, Category c)
+    {
+        if (StringUtils.isEmpty(c.getName())) {
```
tpage-alfresco, line 134: \[Formatting\] Trailing braces a couple of times in this file.

tpage-alfresco: Oh - incidentally we can use org.alfresco.util.ParameterCheck.mandatoryString for this.

remote-api/src/test/java/org/alfresco/rest/api/impl/CategoriesImplTest.java
tpage-alfresco: I think we should ignore them - at least I'm pretty sure that's what happens in other PUT APIs.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CategoriesRestTest.java
krdabrowski: I was also thinking about that, but didn't want to break the convention 😂 Now it's time for a new one - I will add the `builder`

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/CreateCategoriesTests.java
krdabrowski: Ok, I will fix that.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
krdabrowski: I will update the javadoc.

krdabrowski: Since we ignore other fields than `name` this test will be removed.

remote-api/src/main/java/org/alfresco/rest/api/Categories.java
```diff
@@ -31,12 +31,21 @@
...
 @Experimental
 public interface Categories
 {
     Category getCategoryById(String id, Parameters params);
 
     List<Category> createSubcategories(String parentCategoryId, List<Category> categories, Parameters parameters);
+
+    /**
+     * Update category by ID. Currently, it's possible only to update the name of category.
+     * Fixed category fields: id, parentId and hasChildren has to match the original category.
```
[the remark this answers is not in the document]
krdabrowski: I will remove checks for `id`, `parentId` and `hasChildren` fields.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
krdabrowski: Adjusted

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
krdabrowski: I think the same and would leave it as is.

PR #1637, opened 2022-12-20
Fix/mnt 23290 slow group membership

antoniojfelix:
Just to leave some context. The problem of this issue is that when the client has lots of groups (in this case 252000 groups), it takes more than a day to receive a response from the rest api call to get the group memberships.

During the investigation, I realized that the problem happened not only when getting the group memberships, but also in other two different cases, such as when getting a single specific group and when getting all the groups. 

The problem is the query select_ChildAssocsOfParentWithoutParentAssocsOfType that is being called to get the list of root groups. This query gets the set of child associations of a certain parent node without parent associations. In effect the 'orphans' with respect to a certain association type. Then, we have to retrieve the name of the root groups from this result set. This query it's also being used in other places, it's a "generic" query. After that, the list of the root groups will be used to be compared with the response groups in order to set the field `IsRoot` that will also be provided in the response of the rest API call.

To fix this, I basically created a new query specifically to get the list of the names of the root groups. The query is now much faster and it retrieves the list of names directly, so we don't need to do any other operations to retrieve the information.


jottley (approved):

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -1147,6 +1147,21 @@
...
+            on z.parent_node_id = #{parentNode.id}
+        where z.qname_localname not in
+        (
+            select distinct  a.qname_localname
+            from alf_child_assoc a
+            join alf_child_assoc z
+            on z.parent_node_id = #{parentNode.id}
+            and z.child_node_id = a.parent_node_id
+        )
+    </select>
```
dsibilio, line 1164: There's an extra space between `distinct` and `a.qname_localname`.

repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
```diff
@@ -5035,6 +5037,8 @@ protected abstract void selectChildAssocsWithoutParentAssocsOfType(
             Long parentNodeId,
             QName assocTypeQName,
             ChildAssocRefQueryCallback resultsCallback);
+    protected abstract List<String> selectRootGroupsNames(
+            Long parentNodeId);
```
tpage-alfresco, line 5041: I don't get why we have an overridden method that just passes through to an abstract method.

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -1147,6 +1147,21 @@
             parentNode.id = #{parentNode.id} and
             a.child_node_id IS NULL
     </select>
+
+    <select id="select_RootGroups" parameterType="ChildAssoc" resultType="java.lang.String">
```
tpage-alfresco, line 1151: This query looks more generic than just user groups, would it be possible to use it for something else too? Maybe the name could be made more generic?

tpage-alfresco (commented):
I don't think the node service should have any knowledge of user groups, so I think we should change the names of some of these new methods (the code looked fine to me though).

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -1147,6 +1147,21 @@
...
+            join alf_child_assoc z
+            on z.parent_node_id = #{parentNode.id}
+        where z.qname_localname not in
+        (
+            select distinct  a.qname_localname
+            from alf_child_assoc a
+            join alf_child_assoc z
+            on z.parent_node_id = #{parentNode.id}
+            and z.child_node_id = a.parent_node_id
+        )
```
dsibilio, line 1163: If I'm reading this correctly, we're trying to get all the _"orphan"_ parent rows (rows for the which there is no existing child identifying such row as its parent?).

Assuming this is the intended logic to apply here and I interpreted the query correctly, how would the current subquery fare against a `not exists ( select 1 ...` equivalent clause instead? I assume relying on `not exists` might make the intention of the query clearer + it might perform better since it doesn't have to actually fetch any values but just return a boolean.

Also, does the subquery really need to repeat the self-join, or could it be simplified to something like e.g.:

`where not exists
(
    select 1
    from alf_child_assoc aca
    where aca.parent_node_id = z.child_node_id
)`

-, while yielding the same results?

I may be missing some context here 🤔

repository/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java
antoniojfelix: I was following the same approach of the ChildAssocsOfParentWithoutParentAssocsOfType (the original method that was being called) and didn't notice that for the new method this wasn't needed.

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
antoniojfelix: Initially I thought that the name should be specific but now I realized that the queries can be used for other things. Changed the name to: **select_LocalNameListOfChildAssocsWithoutParent**

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
antoniojfelix: Thanks for the suggestion, after some tweaking, using the `not exists` instead of `not in` improved the query response time by more than 50% in my tests.

repository/src/main/resources/alfresco/ibatis/org.alfresco.repo.domain.dialect.Dialect/node-common-SqlMap.xml
```diff
@@ -1147,6 +1147,20 @@
...
+        select distinct c.qname_localname
+        from alf_child_assoc c
+        where c.parent_node_id = #{parentNode.id}
+        and not exists (
+            select 1
+            from alf_child_assoc a
+            join alf_child_assoc z
+            on z.parent_node_id = #{parentNode.id}
+            and z.child_node_id = a.parent_node_id
+            where c.child_node_id = a.child_node_id
```
tpage-alfresco, line 1161: Going through this query then it will pick out E as a "child assoc without parent":
```
A
|\
B \
C  |
D /
|/
E
```
I don't think this is what you wanted.

tpage-alfresco: We discussed this and it turns out to be a limitation with the original query. Consequently there's no need to fix it, we can just name the method appropriately.

tpage-alfresco (approved):
LGTM

dsibilio (approved):

PR #1654, opened 2023-01-04
MNT-23263 Changes done for missing properties after moving record between categories

atkumar14:
AGS - Missing properties after moving record between categories


tpage-alfresco (approved):
LGTM. Can a customer move their record folder a couple of times to restore the properties or do we need a patch too?

suneet-gupta:
> LGTM. Can a customer move their record folder a couple of times to restore the properties or do we need a patch too?

We would be creating a patch to restore the properties.

PR #1655, opened 2023-01-04
ACS-4034: Assign a piece of content to a category

krdabrowski:
POST /nodes/{nodeId}/category-links
https://alfresco.atlassian.net/browse/ACS-4034
- added also one TAS test for update category covering repeated name case

[ACS-4034]: https://alfresco.atlassian.net/browse/ACS-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
@@ -211,7 +211,7 @@ public class LinkToCategoriesTests extends CategoriesRestTest
     }
 
     /**
-     * Try to link content to category as a owner and expect 201
+     * Try to link content to category as owner and expect 201
      */
     @Test(groups = { TestGroup.REST_API})
     public void testLinkContentToCategory_asOwner()
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
@@ -31,23 +31,18 @@ import static org.alfresco.utility.report.log.Step.STEP;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.HttpStatus.CREATED;
 import static org.springframework.http.HttpStatus.FORBIDDEN;
+import static org.springframework.http.HttpStatus.METHOD_NOT_ALLOWED;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
 
 import javax.json.Json;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 
 import org.alfresco.dataprep.CMISUtil;
-import org.alfresco.rest.model.RestActionBodyExecTemplateModel;
 import org.alfresco.rest.model.RestCategoryLinkBodyModel;
 import org.alfresco.rest.model.RestCategoryModel;
 import org.alfresco.rest.model.RestCategoryModelsCollection;
-import org.alfresco.rest.model.RestCommentModel;
 import org.alfresco.rest.model.RestNodeModel;
-import org.alfresco.rest.model.RestRatingModel;
-import org.alfresco.rest.model.RestRenditionInfoModelCollection;
-import org.alfresco.rest.model.RestRuleModel;
 import org.alfresco.rest.model.RestTagModel;
 import org.alfresco.utility.model.FileModel;
 import org.alfresco.utility.model.FolderModel;
@@ -278,19 +273,32 @@ public class LinkToCategoriesTests extends CategoriesRestTest
     }
 
     /**
-     * Try to link non-content node to category and expect 201 (Created)
+     * Try to link folder node to category and expect 201 (Created)
      */
     @Test(groups = { TestGroup.REST_API})
-    public void testLinkContentToCategory_usingTagInsteadOfContentAndExpect201()
+    public void testLinkContentToCategory_usingFolderInsteadOfContentAndExpect200()
     {
         STEP("Try to link a tag to category and expect 201");
+        final RestCategoryLinkBodyModel categoryLink = createCategoryLinkWithId(category.getId());
+        restClient.authenticateUser(user).withCoreAPI().usingNode(folder).linkToCategory(categoryLink);
+
+        restClient.assertStatusCodeIs(CREATED);
+    }
+
+    /**
+     * Try to link non-content node to category and expect 405 (Method Not Allowed)
+     */
+    @Test(groups = { TestGroup.REST_API})
+    public void testLinkContentToCategory_usingTagInsteadOfContentAndExpect405()
+    {
+        STEP("Try to link a tag to category and expect 405");
         final RestCategoryLinkBodyModel categoryLink = createCategoryLinkWithId(category.getId());
         final RestTagModel tag = restClient.authenticateUser(user).withCoreAPI().usingNode(file).addTag("someTag");
         final RepoTestModel tagNode = new RepoTestModel() {};
         tagNode.setNodeRef(tag.getId());
         restClient.authenticateUser(dataUser.getAdminUser()).withCoreAPI().usingNode(tagNode).linkToCategory(categoryLink);
 
-        restClient.assertStatusCodeIs(CREATED);
+        restClient.assertStatusCodeIs(METHOD_NOT_ALLOWED);
     }
 
     /**
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
@@ -211,10 +211,10 @@ public class LinkToCategoriesTests extends CategoriesRestTest
     }
 
     /**
-     * Try to link content to category as a owner and expect 403 (Forbidden)
+     * Try to link content to category as a owner and expect 201
      */
     @Test(groups = { TestGroup.REST_API})
-    public void testLinkContentToCategory_asOwnerAndExpect403()
+    public void testLinkContentToCategory_asOwner()
     {
         STEP("Use admin to create a private site");
         final SiteModel privateSite = dataSite.usingAdmin().createPrivateRandomSite();
@@ -225,7 +225,7 @@ public class LinkToCategoriesTests extends CategoriesRestTest
         final FileModel privateFile = dataContent.usingUser(user).usingResource(privateFolder).createContent(CMISUtil.DocumentType.TEXT_PLAIN);
         dataUser.removeUserFromSite(user, privateSite);
 
-        STEP("Try to link content to a category as owner and expect 403");
+        STEP("Try to link content to a category as owner and expect 201");
         final RestCategoryLinkBodyModel categoryLink = createCategoryLinkWithId(category.getId());
         restClient.authenticateUser(user).withCoreAPI().usingNode(privateFile).linkToCategory(categoryLink);
 
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
@@ -26,22 +26,32 @@
 
 package org.alfresco.rest.categories;
 
+import static org.alfresco.utility.constants.UserRole.SiteManager;
 import static org.alfresco.utility.report.log.Step.STEP;
 import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.HttpStatus.CREATED;
 import static org.springframework.http.HttpStatus.FORBIDDEN;
 import static org.springframework.http.HttpStatus.NOT_FOUND;
 
+import javax.json.Json;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import org.alfresco.dataprep.CMISUtil;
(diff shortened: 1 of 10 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
```diff
@@ -0,0 +1,276 @@
...
+        fileNode.assertThat().field(PROPERTIES_FIELD).contains(thirdCategory.getId());
+    }
+
+    /**
+     * Try to link content to category as an improper user and expect 403 (Forbidden)
+     */
+    @Test(groups = { TestGroup.REST_API})
+    public void testLinkContentToCategory_usingUserWithoutAccessRightsAndExpect403()
+    {
+        STEP("Try to link content to a category using user without read permission and expect 403");
```
tpage-alfresco, line 183: I think we should try a user without change permission instead (and expect 403).

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
```diff
@@ -0,0 +1,276 @@
...
+        STEP("Try to call link content API with empty category ID and expect 400");
+        final String nonExistingCategoryId = StringUtils.EMPTY;
+        final RestCategoryLinkBodyModel categoryLink = createCategoryLinkWithId(nonExistingCategoryId);
+        restClient.authenticateUser(user).withCoreAPI().usingNode(file).linkToCategory(categoryLink);
+
+        restClient.assertStatusCodeIs(BAD_REQUEST);
+    }
+
+    /**
+     * Try to link non-content node to category and expect 400 (Bad Request)
```
tpage-alfresco, line 232: I'm surprised at this. I thought that Share allowed folders to be linked to categories.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/UpdateCategoriesTests.java
```diff
@@ -93,6 +94,27 @@ public void testUpdateSubcategory_asAdmin()
         updatedCategory.assertThat().field(FIELD_NAME).is(categoryNewName);
     }
 
+    /**
+     * Try to update a category with a name, which is already present within the parent category
+     */
+    @Test(groups = { TestGroup.REST_API})
+    public void testUpdateCategory_usingRecurringName()
```
tpage-alfresco, line 101: This looks like a good extra test, but it seems unrelated to the change. I don't have an objection - more tests is good as you think of them, I just wanted to check that you intended to include this.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -236,15 +292,68 @@ private NodeRef changeCategoryName(final NodeRef categoryNodeRef, final String n
...
+    {
+        if (currentCategories == null)
+        {
+            return newCategories;
+        }
+
+        final Collection<NodeRef> allCategories = DefaultTypeConverter.INSTANCE.getCollection(NodeRef.class, currentCategories);
+        newCategories.stream()
+            .filter(category -> !allCategories.contains(category))
+            .forEach(allCategories::add);
```
tpage-alfresco, line 334: Do we know that `allCategories` is modifiable? Instead we could use the union of two sets to determine what the new value should contain.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -236,15 +292,68 @@ private NodeRef changeCategoryName(final NodeRef categoryNodeRef, final String n
...
+        return allCategories;
+    }
+
+    /**
+     * Add to or update node's property cm:categories containing linked category references.
+     *
+     * @param nodeRef Node reference.
+     * @param categoryNodeRefs Category node references.
+     */
+    private void linkNodeToCategories(final NodeRef nodeRef, final Collection<NodeRef> categoryNodeRefs)
```
tpage-alfresco, line 345: This looks fine to me, but can't we reuse some existing service code for this? (Maybe in the CategoryService?)

mpichura (commented):
I think we are not having parity with some Share functionality (applying categories to folders, possibly more restrictive check on node permissions).

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
tpage-alfresco: Yes - it's probably a good idea to look to see if there's any restriction in the type of node in the existing v0 webscript.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
tpage-alfresco: Ok - that sounds like a good idea, but if it's a more general change then we should get this PR merged first.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
krdabrowski: I don't know why but I assumed that read permission will be enough to create a link. Now after rethinking you are right - I will fix it.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
krdabrowski: I've just verified that and you are true. Share in did allows folder linking. I will fix that.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
krdabrowski: The `.getCollection()` always creates a new `ArrayList<T>` so yes - it's modifiable. In the other hand it's nice suggestion to use sets union - I will change my solution.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
krdabrowski: As Maciej already mentioned, unfortunately there doesn't exist any service code which could be reused. Similar solution is used in `LinkCategoryActionExecuter` and we could update it also there if we would introduce some new "CategoryService".

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
krdabrowski: Regarding other node types - v0 API allows to link different kinds of nodes to categories. I don't think there's any check for that. I've asked Stefan if we should support only Content and Folder nodes.

tpage-alfresco: It's a good idea to check. For now I would take the check out completely so that this can be merged. Then we can have a new ticket for the validation once we know what the requirements are.

remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -141,6 +152,35 @@ public void deleteCategoryById(String id, Parameters parameters)
...
+    public List<Category> linkContentNodeToCategories(final String nodeId, final List<Category> categoryLinks)
+    {
+        if (CollectionUtils.isEmpty(categoryLinks))
+        {
+            throw new InvalidArgumentException(NOT_A_VALID_CATEGORY);
+        }
+
+        final NodeRef contentNodeRef = nodes.validateNode(nodeId);
+        verifyReadPermission(contentNodeRef);
+        verifyNodeType(contentNodeRef, ContentModel.TYPE_CONTENT);
```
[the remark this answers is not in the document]
krdabrowski: Fixed - node type verification removed.

krdabrowski:
After talk with @jamalkm added node type verification. It should look the same way like for tags, see: https://github.com/Alfresco/alfresco-community-repo/blob/master/remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java#L86

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
```diff
@@ -0,0 +1,350 @@
...
+    public void testLinkContentToCategory_usingTagInsteadOfContentAndExpect405()
+    {
+        STEP("Try to link a tag to category and expect 405");
+        final RestCategoryLinkBodyModel categoryLink = createCategoryLinkWithId(category.getId());
+        final RestTagModel tag = restClient.authenticateUser(user).withCoreAPI().usingNode(file).addTag("someTag");
+        final RepoTestModel tagNode = new RepoTestModel() {};
+        tagNode.setNodeRef(tag.getId());
+        restClient.authenticateUser(dataUser.getAdminUser()).withCoreAPI().usingNode(tagNode).linkToCategory(categoryLink);
+
+        restClient.assertStatusCodeIs(METHOD_NOT_ALLOWED);
```
krdabrowski, line 301: `METHOT_NOT_ALLOWED` - to maintain consistency with tags API.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
```diff
@@ -0,0 +1,350 @@
...
+        category = prepareCategoryUnderRoot();
+    }
+
+    /**
+     * Link content to category and verify if this category is present in node's properties
+     */
+    @Test(groups = { TestGroup.REST_API})
+    public void testLinkContentToCategory()
+    {
+        STEP("Check if file is not linked to any category");
```
tpage-alfresco, line 92: It would be better to use a new file for this test, otherwise other tests may cause this to fail.

tpage-alfresco: Ah - I just noticed it's created in `BeforeMethod` - ignore this then.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
```diff
@@ -0,0 +1,350 @@
...
+        STEP("Use admin to create a private site");
+        final SiteModel privateSite = dataSite.usingAdmin().createPrivateRandomSite();
+
+        STEP("Add the user to the site, let him create a folder and then evict him from the site again");
+        dataUser.addUserToSite(user, privateSite, SiteManager);
+        final FolderModel privateFolder = dataContent.usingUser(user).usingSite(privateSite).createFolder();
+        final FileModel privateFile = dataContent.usingUser(user).usingResource(privateFolder).createContent(CMISUtil.DocumentType.TEXT_PLAIN);
+        dataUser.removeUserFromSite(user, privateSite);
+
+        STEP("Try to link content to a category as owner and expect 403");
```
tpage-alfresco, line 228: I don't understand this. The owner should be able to link the content to a category.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
tpage-alfresco: This is fine, but is it documented in the api-explorer?

remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -834,6 +834,8 @@
         <constructor-arg name="categoryService" ref="CategoryService"/>
         <constructor-arg name="nodes" ref="nodes"/>
         <constructor-arg name="nodeService" ref="NodeService"/>
+        <constructor-arg name="permissionService" ref="PermissionService"/>
+        <constructor-arg name="typeConstraint" ref="nodeTypeConstraint"/>
```
tpage-alfresco, line 838: This is a good discovery :+1:

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
mpichura: I see the point of being consistent with Tags API but I don't think this status code is meant for that. AFAIR, it should be returned when http method is not handled/accepted by the server. 
TBH, I don't think we should repeat mistake we made with tags and we should either return `400 Bad request` or `422 Unprocessable entity`

mpichura: As discussed - I believe we can move on with this code and log a ticket to get this on and (possibly tags) fixed to return a more suitable error code.

mpichura (approved):
Apart from the error code 405 part which we hopefully will handle soon, LGTM.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/categories/LinkToCategoriesTests.java
krdabrowski: Ticked created: https://alfresco.atlassian.net/browse/ACS-4362

tpage-alfresco (approved):
LGTM. Sorry - I thought I'd already +1ed this.

PR #1657, opened 2023-01-04
[MNT-23379] Set site manager permission if node belongs to a site and inheritance flag is disabled

tiagosalvado10:


remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -2763,6 +2766,38 @@ public Node updateContent(String fileNodeId, BasicContentInfo contentInfo, Input
...
+                {
+                    SiteInfo containingSite = siteService.getSite(nodeRef);
+
+                    if (containingSite != null)
+                    {
+                        String thisSiteGroupPrefix = siteService.getSiteGroup(containingSite.getShortName());
+                        final String siteManagerAuthority = thisSiteGroupPrefix + "_" + SiteModel.SITE_MANAGER;
+                        AuthenticationUtil.runAs(new AuthenticationUtil.RunAsWork<Void>()
+                        {
+                            public Void doWork() throws Exception
```
tpage-alfresco, line 2785: The `runAs` can be replaced by a lambda function if you want.

tpage-alfresco (approved):
LGTM

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
tiagosalvado10: Replaced the `runAs` call as suggested.

tpage-alfresco (approved):
Looks even better :+1: :+1:

jottley (approved):

PR #1658, opened 2023-01-04
ACS-2967: While LocalPipelineTransform was reported to cause a warnin…

mstrankowski:
…g issue, actually it is already fixed, LocalFailoverTransform however caused same warning


mstrankowski:
No "Could not extract header" can be found in logs after that fix

PR #1674, opened 2023-01-13
ACS-4160: Fixing and making GetTopCategories more re-usable

mpichura:
This PR (although not directly) is introducing changes for getting top categories in Elasticsearcgh setup. 
- created a common method to create search query parameters for top categories in `AbstractCategoryServiceImpl`
- fixed a (most common) case where `cm:generalclassifiable` aspect is used to get top categories
- added unit tests (E2E aren't yet to be added as this implementation is not used anywhere yet)
This PR makes it possible to query top categories with SOLR search in place which didn't seem to work properly before.
Another PR in ent-repo will be submitted to cover ES part of the search once this one is merged.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
+++ b/repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2020 Alfresco Software Limited
+ * Copyright (C) 2005 - 2023 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software. 
  * If the software was purchased under a paid Alfresco license, the terms of 
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
```diff
@@ -546,6 +550,57 @@ public void deleteClassification(StoreRef storeRef, QName aspectName)
...
+     * @param aspectName Aspect name. "cm:generalclassifiable" aspect should be used for usual cases.
+     *                   It is possible to use a custom aspect but it must have valid category property
+     * @param count Will be used as faceted results limit, when system has very many categories this must be reflecting that number
+     * @return SearchParameters to perform search for top categories.
+     */
+    protected SearchParameters createSearchTopCategoriesParameters(StoreRef storeRef, QName aspectName, int count) {
+        final AspectDefinition aspectDefinition = dictionaryService.getAspect(aspectName);
+        if(aspectDefinition == null)
+        {
+            throw new IllegalStateException("Unknown aspect");
```
tpage-alfresco, line 566: Maybe IllegalArgumentException?

mpichura: I didn't like that one too but I wanted to be backwards compatible and that particular part is taken from `SolrCategoryService` class. TBH, I didn't find any useful usage of `getTopCategories` (I think `jscript` part is not useed anywhere) but it may be hidden somehow I can't find it and I am just cautious...
Do you think it is safe to change the type of Exception?

tpage-alfresco (approved):
LGTM

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
```diff
@@ -546,6 +550,57 @@ public void deleteClassification(StoreRef storeRef, QName aspectName)
...
+            final SearchParameters.FieldFacet ff = new SearchParameters.FieldFacet(field);
+            ff.setLimitOrNull(count < 0 ? null : count);
+            sp.addFieldFacet(ff);
+            return sp;
+        })
+                .orElseThrow(() -> new IllegalStateException("Aspect does not have category property mirroring the aspect name"));
+    }
+
+    /**
+     * Checks whether give aspect property definition is valid category property
```
krdabrowski, line 592: I think you forgot here about 'n' in `give` word.

krdabrowski (approved):
LGTM

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
mpichura:
```suggestion
     * Checks whether given aspect property definition is valid category property

```

PR #1697, opened 2023-01-20
ACS-4137 Fix building a multiarch image

damianujma:
The `docker-container` driver is needed to use for multi-platform builds but the driver doesn't support accessing local images. 
We are not able to build the `alfresco/alfresco-governance-repository-community-base` with local image `alfresco/alfresco-community-repo-base:latest` access.
We need to run a local registry to push and pull the local image.


wojtekswieton (approved):

amps/ags/rm-community/rm-community-repo/pom.xml
```diff
@@ -15,6 +15,12 @@
       <app.amp.client.war.folder>${project.build.directory}/${project.build.finalName}-war</app.amp.client.war.folder>
 
       <image.name>alfresco/alfresco-governance-repository-community-base</image.name>
+
+      <local.registry.host>127.0.0.1</local.registry.host>
+      <local.registry.port>5000</local.registry.port>
+      <base.image>alfresco/alfresco-community-repo-base</base.image>
+      <base.image.tag>latest</base.image.tag>
```
dsibilio, line 22: Why is this hard-coded to `latest`? What about the scenarios where `image.tag` is pointing to a specific version rather than latest?

amps/ags/rm-community/rm-community-repo/scripts/prepare_buildx.sh
```diff
@@ -0,0 +1,24 @@
...
+BASE_IMAGE_TAG="${4}"
+BUILDER_NAME="${5}"
+LOCAL_REGISTRY="${LOCAL_REGISTRY_HOST}":"${LOCAL_REGISTRY_PORT}"
+
+#Run a local registry server
+docker run -d -p "${LOCAL_REGISTRY_PORT}":"${LOCAL_REGISTRY_PORT}" --restart=always --name registry registry:2
+
+#Push base image to the local repository
+docker tag "${BASE_IMAGE}":"${BASE_IMAGE_TAG}" "${LOCAL_REGISTRY}"/"${BASE_IMAGE}":"${BASE_IMAGE_TAG}"
+while [ "$( docker container inspect -f '{{.State.Running}}' registry )" != "true" ]
```
dsibilio, line 15: We should add a reasonable timeout condition here to fail the script in case we wait on the registry to come up for too long.

dsibilio (changes requested):

amps/ags/rm-community/rm-community-repo/pom.xml
damianujma: Right! I replaced it with `${image.tag}`.

amps/ags/rm-community/rm-community-repo/scripts/prepare_buildx.sh
damianujma: I added a timeout condition (2 minutes).

dsibilio (changes requested):
Shouldn't the `SLEEP_SECONDS` variable be initialized to `0` before checking for the first time that it's less than `120`?

Running this `while [ $SLEEP_SECONDS -lt 10 ]; do ((SLEEP_SECONDS++)); sleep 1; done` locally, outputs: `bash: [: -lt: unary operator expected`.
Setting `SLEEP_SECONDS=0` just before executing it should be enough to make it work correctly.

damianujma:
> Shouldn't the `SLEEP_SECONDS` variable be initialized to `0` before checking for the first time that it's less than `120`?
> 
> Running this `while [ $SLEEP_SECONDS -lt 10 ]; do ((SLEEP_SECONDS++)); sleep 1; done` locally, outputs: `bash: [: -lt: unary operator expected`. Setting `SLEEP_SECONDS=0` just before executing it should be enough to make it work correctly.

I initialized the `SLEEP_SECONDS` variable to `0` in this commit: https://github.com/Alfresco/alfresco-community-repo/commit/add8fb003f022e69f8ce3c9cd8751ca4fba53709#diff-7e8077050235226fbaf62fe601f0ceeb0ae084068d19d4cf7e43aad078ff44c6R9

dsibilio (approved):
LGTM. But before we merge [feature/ACS-4137_Produce_Multi-Arch_Docker_images](https://github.com/Alfresco/alfresco-community-repo/tree/feature/ACS-4137_Produce_Multi-Arch_Docker_images) into master we definitely need to run more checks when building this image starting from tagged base images.

PR #1703, opened 2023-01-25
MNT-23411 Option to disable events2

evasques:
* Added configuration repo.event2.enabled (true by default)
* Prevent the event2 behaviours from binding on bootstrap if disabled
* Implemented enable and disable methods that can be called in runtime on EventGenerator - this disables/enables the binded behaviours


jottley (approved):

jamalkm (approved):
LGTM, great work!

PR #1709, opened 2023-01-27
ACS-4027: Introduce changes to support removing tag from all associated nodes

gevagel:


remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -128,37 +138,49 @@ public void deleteTag(String nodeId, String tagId)
...
-        {
-            Tag selectedTag = new Tag(pair.getFirst(), pair.getSecond());
-            selectedTag.setCount(tagsByCountMap.get(selectedTag.getTag()));
-            tags.add(selectedTag);
-        }
-
-        return CollectionWithPagingInfo.asPaged(paging, tags, results.hasMoreItems(), (totalItems == null ? null : totalItems.intValue()));
-    }
+    @Override
+	public void deleteTagById(StoreRef storeRef, String tagId) {
```
tpage-alfresco, line 142: Tabs

gevagel: Thanks for pointing that out Tom. Just to make sure, you are referring to this game with an acronym, right? [https://en.wikipedia.org/wiki/Totally_Accurate_Battle_Simulator](https://en.wikipedia.org/wiki/Totally_Accurate_Battle_Simulator) 😅

tpage-alfresco: Ah yes of course!

tpage-alfresco (approved):
Other than the whitespace this looks good to me.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertThrows;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.then;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TagsImplTest
```
tpage-alfresco, line 48: Hmm - why doesn't this new test class need to be added to a test suite? There's a test OmittedTestClassFinderUnitTest which I would have expected to pick this up.

krdabrowski (approved):

gevagel:
After discussing with @krdabrowski, I will merge this to allow for other work to proceed, will follow up with a separate PR for E2E tests

PR #1717, opened 2023-01-31
[MNT-22633] Creation, Modification TimeStamp values setted for download as zip

pjoshi31:
Creation and Last Modification Timestamp changed during "Download As ZIP"


repository/src/main/java/org/alfresco/repo/download/ZipDownloadExporter.java
```diff
@@ -137,13 +144,30 @@ public void start(final ExporterContext context)
...
         this.currentName = (String)nodeService.getProperty(nodeRef, ContentModel.PROP_NAME);
+        Date createdUtcTimestamp = (Date)nodeService.getProperty(nodeRef, ContentModel.PROP_CREATED);
+        Date modifiedUtcTimestamp = (Date)nodeService.getProperty(nodeRef, ContentModel.PROP_MODIFIED);
+        TimeZone time_zone_default = TimeZone.getDefault();
+        SimpleDateFormat formatter= new SimpleDateFormat("dd MMM yyyy HH:mm:ss z");
+        formatter.setTimeZone(time_zone_default);
+        try
+        {
+            this.zipTimestampCreated = formatter.parse(formatter.format(createdUtcTimestamp));
+            this.zipTimestampModified = formatter.parse(formatter.format(modifiedUtcTimestamp));
```
tpage-alfresco, line 155: What is this about?? Why are we running the date through the formatter twice?

repository/src/main/java/org/alfresco/repo/download/ZipDownloadExporter.java
```diff
@@ -137,13 +144,30 @@ public void start(final ExporterContext context)
...
+        }
         path.push(new Pair<String, NodeRef>(currentName, nodeRef));
         if (dictionaryService.isSubClass(nodeService.getType(nodeRef), ContentModel.TYPE_FOLDER))
         {
             String path = getPath() + PATH_SEPARATOR;
             ZipArchiveEntry archiveEntry = new ZipArchiveEntry(path);
             try
             {
+                archiveEntry.setTime(zipTimestampCreated.getTime());
+                archiveEntry.setCreationTime(FileTime.fromMillis(zipTimestampCreated.getTime()));
```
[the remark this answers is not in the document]
pjoshi31: We can use the createdUtcTimestamp and modifiedUtcTimestamp but the both variables are having UTC Time, I don't think that will be a correct for expected behaviour.

repository/src/main/java/org/alfresco/repo/download/ZipDownloadExporter.java
pjoshi31: This is all for converting the UTC time to local time and we are using formatter twice because the Format method returns string and we are converting to date via parse method. 
As date is required for the setting time in the apache archive library.

repository/src/main/java/org/alfresco/repo/download/ZipDownloadExporter.java
```diff
@@ -167,6 +176,9 @@ public void contentImpl(NodeRef nodeRef, QName property, InputStream content, Co
         {
             // ALF-2016
             ZipArchiveEntry zipEntry=new ZipArchiveEntry(getPath());
+            zipEntry.setTime(zipTimestampCreated.getTime());
+            zipEntry.setCreationTime(FileTime.fromMillis(zipTimestampCreated.getTime()));
+            zipEntry.setLastModifiedTime(FileTime.fromMillis(zipTimestampModified.getTime()));
             zipStream.putArchiveEntry(zipEntry);
```
tpage-alfresco, line 182: Is it worth refactoring out all this common code into a method?

PR #1720, opened 2023-02-01
ACS-4155 Update webscripts since we only index full category path in ES.

tpage-alfresco:


mpichura (approved):
Assuming this won't break anything SOLR related.

krdabrowski (approved):

KikoUK (approved):

tpage-alfresco:
> Assuming this won't break anything SOLR related.

It shouldn't do, since Solr indexes two paths per category - one starting with `/cm:categoryRoot/cm:generalclassifiable` and the other starting with `/cm:generalclassifiable`.

PR #1735, opened 2023-02-07
[ACS-4459] Investigate and extend/universalize current custom Solr mTLS implementation in Repository

kmagdziarz:


core/src/main/java/org/alfresco/httpclient/HttpClient4Factory.java
```diff
@@ -0,0 +1,104 @@
...
+
+            SSLConnectionSocketFactory sslConnectionSocketFactory = new SSLConnectionSocketFactory(
+                    createSSLContext(config),
+                    new String[] { TLS_V_1_2, TLS_V_1_3 },
+                    null,
+                    SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+            clientBuilder.setSSLSocketFactory(sslConnectionSocketFactory);
+        }
+
+        clientBuilder.setMaxConnTotal(Integer.parseInt(config.getConfig().get("maxTotalConnections")));
```
pzhyland, line 92: Consider applying [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) here by hiding the internals of the config class. What about something like:
```java
config.getMaxTotalConnections()
```
instead of
```java
Integer.parseInt(config.getConfig().get("maxTotalConnections"))
```

kmagdziarz: I had similar approach earlier, but I agreed with @mstrankowski to pass raw map end retreive properties directly in place of usage - such approach was considered as more beneficial if in the future someone will add more properties to config. But I can change that to yours suggestion @pzhyland

pzhyland: So at least it could be change to something like this:
```java
config.getIntegerProperty("maxTotalConnections")
config.getBooleanProperty(...)
config.getKeyManagers()
config.getTrustManagers()
```

kmagdziarz: That is a good idea. +1

core/src/main/java/org/alfresco/httpclient/HttpClientException.java
```diff
@@ -0,0 +1,11 @@
+package org.alfresco.httpclient;
```
pzhyland, line 1: No cpyright header

repository/src/main/java/org/alfresco/transform/registry/CombinedConfig.java
```diff
@@ -67,8 +68,11 @@ public class CombinedConfig extends CombinedTransformConfig
     private ConfigFileFinder configFileFinder;
     private int tEngineCount;
 
-    public CombinedConfig(Log log, AbstractTransformRegistry registry)
+    HttpClientConfig httpClientConfig;
```
pzhyland, line 71: `private final`?

repository/src/main/java/org/alfresco/transform/registry/CombinedConfig.java
```diff
@@ -89,10 +93,11 @@ public boolean addLocalConfig(String path)
 
     public boolean addRemoteConfig(List<String> urls, String remoteType)
     {
+        CloseableHttpClient httpclient = HttpClient4Factory.createHttpClient(httpClientConfig);
```
pzhyland, line 96: `Closeable...` is not closed which might lead to a resource leak.

pzhyland (changes requested):
This PR seems to be incomplete. It only deals with getting the configuration. Missing parts are requesting a transformation and getting results from the SFS.

core/src/main/java/org/alfresco/httpclient/HttpClient4Factory.java
```diff
@@ -0,0 +1,117 @@
...
+    public static CloseableHttpClient createHttpClient(HttpClientConfig config, HttpClientConnectionManager connectionManager)
+    {
+        HttpClientBuilder clientBuilder = HttpClients.custom();
+
+        if(config.getBooleanProperty("mTLSEnabled"))
+        {
+            clientBuilder.addInterceptorFirst((HttpRequestInterceptor) (request, context) -> {
+                if (!((HttpHost) context.getAttribute(HTTP_TARGET_HOST)).getSchemeName().equals(HTTPS_PROTOCOL))
+                {
+                    String msg = "mTLS is enabled but provided URL has not secured protocol";
```
mstrankowski, line 85: does not use

core/src/main/java/org/alfresco/httpclient/HttpClient4Factory.java
```diff
@@ -0,0 +1,117 @@
...
+            clientBuilder.setSSLSocketFactory(sslConnectionSocketFactory);
+        }
+
+        if(connectionManager != null)
+        {
+            clientBuilder.setConnectionManager(connectionManager);
+        }
+
+        RequestConfig requestConfig = RequestConfig.custom()
+               .setConnectTimeout(config.getIntegerProperty("connectionTimeout"))
```
mstrankowski, line 104: I know that I recommended this approach previously, but it might be worth it to move it down, so that config.getConnectionTimeout() could be called for example.

It is defined anyway and we'd gain the ability to validate the properties set. If someone provides an unsupported property or makes a typo in its name, we could call it out in an exception immediately.

repository/src/main/resources/alfresco/encryption-context.xml
```diff
@@ -126,4 +126,5 @@
     <bean id="keyStoreChecker" class="org.alfresco.encryption.KeyStoreChecker">
         <property name="mainKeyStore" ref="keyStore"/>
     </bean>
+
```
mstrankowski, line 129: Doesn't really make a difference, but removing that could remove the file mention in this PR. I don't mind it either way though

mstrankowski (approved):
LGTM

pzhyland (approved):

core/src/main/java/org/alfresco/httpclient/HttpClientConfig.java
```diff
@@ -0,0 +1,215 @@
...
+        String keyValue;
+        if(property.isRequired)
+        {
+            keyValue = Optional.ofNullable(config.get(property.name)).orElseThrow(() -> {
+                String msg = String.format("Required property: '%s' is empty.", property.name);
+                LOGGER.error(msg);
+                throw new HttpClientException(msg);
+            });
+        } else {
+            keyValue = config.getOrDefault(property.name, "0");
```
mstrankowski, line 144: What about different property value types? Shouldn't it return null and the processing of a null return be done in the getInteger/getBoolean, or possibly in the future getStringProperty?

Is 0 parsed to false with parseBoolean?

core/src/main/java/org/alfresco/httpclient/HttpClient4Factory.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * #%L
+ * Alfresco Records Management Module
```
mstrankowski, line 3: Are the copyright modules accurate? It doesn't seem to fit with Records Management here

scripts/ci/generate_keystores.sh
```diff
@@ -0,0 +1,27 @@
...
+#! /bin/bash
+
+# SETTINGS
+# Alfresco Format: "classic" / "current" is supported only from 7.0
+ALFRESCO_FORMAT=current
+
+#Contains directory settings
+source ${GITHUB_WORKSPACE}/alfresco-ssl-generator/ssl-tool/utils.sh
+
+# Cleanup previous output of script
```
mstrankowski, line 11: that might not be needed on GHA, it clears everything by default

mstrankowski: It's not really stopping anything either though, so little importance :D

PR #1742, opened 2023-02-08
Added tagById validation if node is a tag

suneet-gupta:
Added tagById validation if node is a tag


tpage-alfresco:
It looks like this will break `TagsImpl.getTags` which I think might erroneously check if the passed node is a tag. Please can you investigate what this method is used for?

PR #1766, opened 2023-02-22
ACS-4023: Update GET /tags to support getting tags by name

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-4023


remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -81,36 +97,129 @@
...
+
+    @Test
+    public void testGetTags_verifyIfCountIsZero()
+    {
+        given(parametersMock.getPaging()).willReturn(pagingMock);
+        given(taggingServiceMock.getTags(any(StoreRef.class), any(PagingRequest.class), any())).willReturn(pagingResultsMock);
+        given(pagingResultsMock.getTotalResultCount()).willReturn(new Pair<>(Integer.MAX_VALUE, 0));
+        given(pagingResultsMock.getPage()).willReturn(List.of(new Pair<>(TAG_NODE_REF, TAG_NAME)));
         given(parametersMock.getInclude()).willReturn(List.of("count"));
-        final List<Tag> actualCreatedTags = objectUnderTest.createTags(tagsToCreate, parametersMock);
```
krdabrowski, line 93: I've fixed this test, because is should test `count` in method `.getTags()` not `.createTags()`

remote-api/src/main/java/org/alfresco/rest/workflow/api/impl/MapBasedQueryWalker.java
```diff
@@ -247,25 +248,31 @@ else if (type == WhereClauseParser.LESSTHANOREQUALS)
...
+        }
+        else if (negated)
+        {
+            // Throw error for the unsupported negation only if the property was valid for comparison, show the more meaningful error first.
+            throw new InvalidArgumentException("Cannot use NOT for " + WhereClauseParser.tokenNames[type] + " comparison.");
         }
     }
 
+    @Override
+    public Collection<String> getProperty(String propertyName, int type, boolean negated)
```
krdabrowski, line 271: This my be hard to notice, but adding this method should be the only change in this class.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -226,4 +228,91 @@ public void checkThatDeletedTagIsNotRetrievedAnymore() throws Exception
...
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+            .withParams("where=(name IN ('" + documentTag.getTag() + "', '" + folderTag.getTag() + "'))")
+            .withCoreAPI()
+            .getTags();
+
+        restClient.assertStatusCodeIs(HttpStatus.OK);
+        returnedCollection.assertThat()
+            .entriesListCountIs(2).and()
+            .entriesListContains("tag", documentTagValue.toLowerCase()).and()
+            .entriesListContains("tag", folderTagValue.toLowerCase());
```
tpage-alfresco, line 266: Idea for the future: I think this would be neater if we had support for:
```
returnedCollection.assertThat().entrySetMatches("tag", Set.of("tag1", "tag2"));
```
or
```
returnedCollection.map(entry -> entry.get("tag")).assertThat().equals(Set.of("tag1", "tag2"));
```

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -226,4 +228,91 @@ public void checkThatDeletedTagIsNotRetrievedAnymore() throws Exception
...
+    {
+        STEP("Get tags with names filter using EQUALS (in fact contains) and expect two item in result");
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+            .withParams("where=(name=orphan)")
+            .withCoreAPI()
+            .getTags();
+
+        restClient.assertStatusCodeIs(HttpStatus.OK);
+        returnedCollection.assertThat()
+            .entriesListIsNotEmpty().and()
```
tpage-alfresco, line 283: This validation looks like it's covered by the next line.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -226,4 +228,91 @@ public void checkThatDeletedTagIsNotRetrievedAnymore() throws Exception
...
+
+    /**
+     * Try to get tags with names filter using EQUALS and wrong property name and expect 400 (Bad Request).
+     */
+    @Test(groups = { TestGroup.REST_API, TestGroup.TAGS, TestGroup.REGRESSION })
+    public void testGetTags_withWrongWherePropertyNameAndExpect400()
+    {
+        STEP("Try to get tags with names filter using EQUALS and wrong property name and expect 400");
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+            .withParams("where=(tag=gat)")
```
tpage-alfresco, line 295: This is a good point. Do you think `where=(tag=...)` would be clearer than `where=(name=...)` for the client? Maybe we should propose this change in the API explorer?

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -226,4 +228,91 @@ public void checkThatDeletedTagIsNotRetrievedAnymore() throws Exception
...
+
+    /**
+     * Try to get tags with names filter using EQUALS and IN at the same time and expect 400 (Bad Request).
+     */
+    @Test(groups = { TestGroup.REST_API, TestGroup.TAGS, TestGroup.REGRESSION })
+    public void testGetTags_bothEqualsAndInNotAllowed()
+    {
+        STEP("Try to get tags with names filter using EQUALS and IN at the same time and expect 400");
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+            .withParams("where=(name=tag AND name IN ('tag-', 'gat'))")
```
tpage-alfresco, line 311: Is there a technical reason why we can't use these together? Can we use `where=(name=tag AND name=tag)`?  I would have expected it to work, but I don't think it's very useful functionality to double filter the field like this.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -183,7 +192,7 @@ public CollectionWithPagingInfo<Tag> getTags(StoreRef storeRef, Parameters param
             tags.add(selectedTag);
         }
 
-        return CollectionWithPagingInfo.asPaged(paging, tags, results.hasMoreItems(), (totalItems == null ? null : totalItems.intValue()));
+        return CollectionWithPagingInfo.asPaged(paging, tags, results.hasMoreItems(), totalItems);
```
tpage-alfresco, line 195: This seems better :+1:

tpage-alfresco (changes requested):
Looks good. As we discussed we should update this PR to use something other than `=` and `IN` because those are anticipated to be exact matches.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
krdabrowski: Yes, `tag` in where clause will be better as in result entry we are also returning field 'tag'.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
krdabrowski: You are right, the first approach looks way better.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
krdabrowski: Will fix it.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
krdabrowski: The issue is that I don't know which filter should be picked up for e.g. `where=(tag=hr AND tag=is)`. We cannot determine if tag should be "hr" or "is". This is the reason `AND` is not supported for tags filtering. I think that we can support `OR` operator instead.

tpage-alfresco: I think that `where=(tag=hr AND tag=is)` should match no tags. Anyway - I think we discussed already that it's fine not to support boolean operators at all if they're not trivial. The main feature we want is for the FE to be able to filter by exact match (`where=(tag in ('hr', 'is'))`).

krdabrowski: Hmm... 🤔 good point with no tags! Anyway `AND` is not supported. `OR` wasn't too complex, so it's supported.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/BasicQueryWalker.java
```diff
@@ -0,0 +1,259 @@
...
+        verifyPropertyExpectedness(propertyName);
+        verifyClausesNegatability(negated, propertyName);
+        addProperties(propertyName, WhereClauseParser.MATCHES, negated, propertyValue);
+    }
+
+    @Override
+    public void and()
+    {
+        // Don't need to do anything here - it's enough to enable AND operator.
+        // OR is not supported at the same time.
```
tpage-alfresco, line 144: Higher up you seemed to say the opposite - that OR is supported, but AND isn't.

remote-api/src/main/java/org/alfresco/rest/workflow/api/impl/MapBasedQueryWalker.java
```diff
@@ -247,25 +248,31 @@ else if (type == WhereClauseParser.LESSTHANOREQUALS)
...
+    {
+        return Set.of(this.getProperty(propertyName, type));
```
tpage-alfresco, line 273: What does `negated` mean in this method? Why is it ignored?

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
```diff
@@ -1,623 +1,677 @@
...
+                    childNodes.addAll(nodeService.getChildrenByName(nodeRef, ContentModel.ASSOC_SUBCATEGORIES, exactNamesFilter));
+                }
+                if (CollectionUtils.isNotEmpty(alikeNamesFilter))
+                {
+                    // lookup using search engin filtering by name
+                    childNodes.addAll(getChildren(nodeRef, Mode.SUB_CATEGORIES, Depth.IMMEDIATE, sortByName, alikeNamesFilter, skipCount + maxItems + 1));
+                }
+            }
+
+            Stream<ChildAssociationRef> childNodesStream = childNodes.stream().distinct();
```
tpage-alfresco, line 450: I think I mentioned this before, but if we used a Set then we wouldn't need to use `distinct` here.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/BasicQueryWalker.java
krdabrowski: Class `BasicQueryWalker` by default supports `AND` operations, but it's overwritten by:
`QueryHelper
	.resolve(namesQuery)
	.usingOrOperator()`
To support `OR` operations instead.

remote-api/src/main/java/org/alfresco/rest/workflow/api/impl/MapBasedQueryWalker.java
krdabrowski: This field defines if the way of comparison (let say "EQUALS") should be negated - "NOT EQUALS". Class `MapBasedQueryWalker` doesn't support `NOT` operations at all, and this is the reason why this field is ignored. Maybe an exception should be thrown for `negated == true` 🤔

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
krdabrowski: Will fix it.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/QueryHelper.java
```diff
@@ -61,63 +75,67 @@
...
-        void or();
+         */
+        default void or()
+        {
+            throw UNSUPPORTED;
+        }
+
+        default Collection<String> getProperty(String propertyName, int type, boolean negated)
+        {
+            return Collections.emptyList();
```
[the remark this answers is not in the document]
krdabrowski: After re-thinking it could throw an exception by default. Thanks for the hint.

remote-api/src/main/java/org/alfresco/rest/workflow/api/impl/MapBasedQueryWalker.java
tpage-alfresco: Either that or just make the javadoc clearer that negated will be ignored?

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/BasicQueryWalker.java
tpage-alfresco: Ah - I missed that. Thanks!

remote-api/src/main/java/org/alfresco/rest/workflow/api/impl/MapBasedQueryWalker.java
krdabrowski: Added javadoc with explanation.

krdabrowski:
Fixes to above comments pushed.

remote-api/src/main/java/org/alfresco/rest/api/model/Tag.java
```diff
@@ -101,36 +101,20 @@ public int compareTo(Tag o)
...
-		final int prime = 31;
-		int result = 1;
-		result = prime * result + ((nodeRef == null) ? 0 : nodeRef.hashCode());
-		return result;
+		if (this == o)
+			return true;
+		if (o == null || getClass() != o.getClass())
+			return false;
+		Tag tag1 = (Tag) o;
+		return Objects.equals(nodeRef, tag1.nodeRef) && Objects.equals(tag, tag1.tag) && Objects.equals(count, tag1.count);
```
mpichura, line 111: Is that really the case? I think `NodeRef` should be enough as `count` field not always gets populated but I may be wrong. When comparing a tag without `count` with the same tag with`count` populated - they should still be equal, shouldn't they? Tag's `name` (aka `tag`) must be the same for the same node as it points to the same node property.
Unless there are some read/write inconsistencies nothing more than `NodeRef` (tag's `NodeRef` is the node reference of tag node) should be needed here to make sure two tags are equal, shoud there?

tpage-alfresco (approved):

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/QueryHelper.java
```diff
@@ -286,4 +304,180 @@ public static String stripQuotes(String toBeStripped)
...
+
+    public static QueryResolver.WalkerSpecifier resolve(final Query query)
+    {
+        return new QueryResolver.WalkerSpecifier(query);
+    }
+
+    /**
+     * Helper class allowing WHERE query resolving using query walker. By default {@link BasicQueryWalker} is used, but different walker can be supplied.
+     */
+    public static abstract class QueryResolver<S extends QueryResolver<?>>
```
mpichura, line 316: The level of class nesting is a bit of an overkill to me here but don't mind my opinion too much as I may not be getting the whole picture right.

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
```diff
@@ -1,623 +1,677 @@
...
+                {
+                    moreItems = true;
+                    break OUTER_LOOP;
+                }
+
+                associations.add(child);
+            }
+        }
+
+        return createPagingResult(associations, moreItems);
```
mpichura, line 480: Wouldn't it be easier her to use `ListBackedPagingResults` class here?

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
```diff
@@ -1,623 +1,677 @@
...
+        {
+            throw new CategoryServiceException(CATEGORY_ROOT_NODE_NOT_FOUND);
+        }
+        return categoryRootAssocs.stream()
+                .filter(ca -> ca.getQName().equals(ContentModel.ASPECT_GEN_CLASSIFIABLE))
+                .map(ChildAssociationRef::getChildRef)
+                .findFirst();
+    }
+
+    private <T> PagingResults<T> createPagingResult(final List<T> associations, final boolean hasMoreItems)
```
mpichura, line 648: This method may not be neede if we used `ListBackedPagingResults` class where it is called.

mpichura (changes requested):
I had some questins and suggestions.

packaging/tests/tas-restapi/src/main/java/org/alfresco/rest/core/assertion/ModelsCollectionAssertion.java
```diff
@@ -143,6 +146,53 @@ public C entriesListDoesNotContain(String key, String value)
     return (C) modelCollection;
   }
 
+  public C entrySetContains(String key, String... expectedValues)
```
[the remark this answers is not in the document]
krdabrowski: You're right, but this is the way how TAS assertions methods are written - using builder pattern. Later other assertions can be applied on this object.

packaging/tests/tas-restapi/src/main/java/org/alfresco/rest/core/assertion/ModelsCollectionAssertion.java
```diff
@@ -143,6 +146,53 @@ public C entriesListDoesNotContain(String key, String value)
...
+  {
+    return entrySetContains(key, Arrays.stream(expectedValues).collect(Collectors.toSet()));
+  }
+
+  @SuppressWarnings("unchecked")
+  public C entrySetContains(String key, Collection<String> expectedValues)
```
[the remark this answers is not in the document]
krdabrowski: You're right, but this is the way how TAS assertions methods are written - using builder pattern. Later other assertions can be applied on this object.

packaging/tests/tas-restapi/src/main/java/org/alfresco/rest/core/assertion/ModelsCollectionAssertion.java
```diff
@@ -143,6 +146,53 @@ public C entriesListDoesNotContain(String key, String value)
...
+        .collect(Collectors.toSet());
+
+    Assert.assertTrue(actualValues.containsAll(expectedValues), String.format("Entry with key: \"%s\" is expected to contain values: %s, but actual values are: %s",
+        key, expectedValues, actualValues));
+
+    return (C) modelCollection;
+  }
+
+  @SuppressWarnings("unchecked")
+  public C entrySetMatches(String key, Collection<String> expectedValues)
```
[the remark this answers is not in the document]
krdabrowski: You're right, but this is the way how TAS assertions methods are written - using builder pattern. Later other assertions can be applied on this object.

remote-api/src/main/java/org/alfresco/rest/framework/resource/parameters/where/QueryHelper.java
krdabrowski: IMO nested helper classes are meant to use only within `QueryResolver` using builder pattern, so they don't have to be placed outside it. In general I agree with you that in different use case it would be an overkill.

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
krdabrowski: Good hint, I've used it - thanks!

repository/src/main/java/org/alfresco/repo/search/impl/AbstractCategoryServiceImpl.java
krdabrowski: Fixed

remote-api/src/main/java/org/alfresco/rest/api/model/Tag.java
mpichura: We probably don't use `Tag` instances in hash-based structures (do we?) but if we did - then this `equals`/`hasCode` implementation could result in having unwanted duplicates.

krdabrowski: No, `Tag`s are used only with lists. Regarding `count` field - almost all the time it is `null`, so I assumed that it shouldn't be a huge deal even for hash-based structures.

mpichura (approved):

PR #1767, opened 2023-02-23
[ACS-4636] "?include=count" parameter is not supported for creation a tag for a node

pjoshi31:


remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -104,6 +104,8 @@ public List<Tag> addTags(String nodeId, final List<Tag> tags)
 			{
 				throw new UnsupportedResourceOperationException("Cannot tag this node");
 			}
+			List<Pair<String, Integer>> tagsByCount = null;
+			Map<String, Integer> tagsByCountMap = new HashMap<String, Integer>();
```
tpage-alfresco, line 108: Neither of these things are named clearly. They both look likely to contain the same information, so maybe we don't need both. I think a better name might be `tagsAndCounts` or `tagCountMap`.

tpage-alfresco: Yes - looking below I can see that the list shouldn't be defined here.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -124,9 +126,19 @@ public int size()
             {
 		        List<Pair<String, NodeRef>> tagNodeRefs = taggingService.addTags(nodeRef, tagValues);
 		        List<Tag> ret = new ArrayList<Tag>(tags.size());
+				tagsByCount = taggingService.findTaggedNodesAndCountByTagName(nodeRef.getStoreRef());
```
tpage-alfresco, line 129: Is this the only method to find the count for tags? I don't think it returns the full list of tags if there are too many in the system.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -124,9 +126,19 @@ public int size()
...
+					for (Pair<String, Integer> tagByCountElem : tagsByCount)
+					{
+						tagsByCountMap.put(tagByCountElem.getFirst(), tagByCountElem.getSecond());
+					}
+				}
 		        for(Pair<String, NodeRef> pair : tagNodeRefs)
 		        {
-		        	ret.add(new Tag(pair.getSecond(), pair.getFirst()));
+					Tag createdTag=new Tag(pair.getSecond(), pair.getFirst());
+					createdTag.setCount(Optional.ofNullable(tagsByCountMap.get(createdTag.getTag())).orElse(0));
```
tpage-alfresco, line 140: As discussed this will usually be out by one, since the tagged node has not been indexed yet. It might be better to always assume the count will be one more than whatever the index says?

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -159,10 +159,10 @@ public void testCreateTags()
         then(authorityServiceMock).shouldHaveNoMoreInteractions();
         then(taggingServiceMock).should().createTags(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, tagNames);
         then(taggingServiceMock).shouldHaveNoMoreInteractions();
-        final List<Tag> expectedTags = createTagsWithNodeRefs(tagNames);
-        assertThat(actualCreatedTags)
-            .isNotNull()
-            .isEqualTo(expectedTags);
+        final List<Tag> expectedTags = createTagsWithNodeRefs(tagNames).stream()
+                                            .peek(tag -> tag.setCount(0))
```
krdabrowski, line 163: This test verifies a different method -> `.createTags(List<Tag> tags, Parameters parameters)`, and you was working on method `.addTags(String nodeId, List<Tag> tags)`. You should introduce new tests which will test right method and your fix.

krdabrowski: Interesting why this test didn't fail 🤔

atkumar14: Reverted the changes in the original test cases. Added a new test case for the above scenario.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
atkumar14: done

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
atkumar14: done

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -81,9 +89,22 @@
...
+        given(nodesMock.validateNode(NODE_ID)).willReturn(TAG_NODE_REF);
         given(nodesMock.validateNode(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, TAG_ID)).willReturn(TAG_NODE_REF);
+        given(nodesMock.getNode(any())).willReturn(createdNodeMock);
+        given(createdNodeMock.getNodeId()).willReturn(NODE_ID);
+        given(typeConstraint.matches(any())).willReturn(true);
         given(taggingServiceMock.getTagName(TAG_NODE_REF)).willReturn(TAG_NAME);
+        given(taggingServiceMock.findTaggedNodesAndCountByTagName(any())).willReturn(createTagsCountPairList());
     }
+
+    public List<Pair<String,Integer>> createTagsCountPairList()
```
tpage-alfresco, line 101: I guess this doesn't need to be public?
```suggestion
    private List<Pair<String,Integer>> createTagsCountPairList()
```

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -81,9 +89,22 @@
     public void setup()
     {
         given(authorityServiceMock.hasAdminAuthority()).willReturn(true);
+        given(nodesMock.validateNode(NODE_ID)).willReturn(TAG_NODE_REF);
         given(nodesMock.validateNode(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, TAG_ID)).willReturn(TAG_NODE_REF);
+        given(nodesMock.getNode(any())).willReturn(createdNodeMock);
+        given(createdNodeMock.getNodeId()).willReturn(NODE_ID);
+        given(typeConstraint.matches(any())).willReturn(true);
         given(taggingServiceMock.getTagName(TAG_NODE_REF)).willReturn(TAG_NAME);
+        given(taggingServiceMock.findTaggedNodesAndCountByTagName(any())).willReturn(createTagsCountPairList());
```
tpage-alfresco, line 98:
```suggestion
        given(taggingServiceMock.findTaggedNodesAndCountByTagName(any())).willReturn(List.of(new Pair<>("tag-node-id",1)));
```

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -302,4 +323,19 @@ private static Tag createTagWithNodeRef(final String tagName)
             .tag(tagName)
             .create();
     }
+
+    @Test
+    public void testAddTagsToNode()
```
tpage-alfresco, line 328: Is there value in a test for applying multiple tags? How about a test where the tag is not in the response from `findTaggedNodesAndCountByTagName`?

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -124,9 +125,13 @@ public int size()
             {
 		        List<Pair<String, NodeRef>> tagNodeRefs = taggingService.addTags(nodeRef, tagValues);
 		        List<Tag> ret = new ArrayList<Tag>(tags.size());
+				List<Pair<String, Integer>> tagsCountPairList = taggingService.findTaggedNodesAndCountByTagName(nodeRef.getStoreRef());
+				Map<String, Integer> tagsCountMap = tagsCountPairList.stream().collect(Collectors.toMap(Pair::getFirst,Pair::getSecond));
 		        for(Pair<String, NodeRef> pair : tagNodeRefs)
 		        {
-		        	ret.add(new Tag(pair.getSecond(), pair.getFirst()));
+					Tag createdTag=new Tag(pair.getSecond(), pair.getFirst());
+					createdTag.setCount(Optional.ofNullable(tagsCountMap.get(createdTag.getTag())).map(cnt->cnt+1).orElse(1));
```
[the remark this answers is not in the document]
suneet-gupta: Added Comment

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
suneet-gupta: Done

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
suneet-gupta: Removed the method

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
suneet-gupta: Added testcase for multiple tags and findTaggedNodesAndCountByTagName returning empty list

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -302,4 +314,34 @@ private static Tag createTagWithNodeRef(final String tagName)
...
+        final List<String> tagNames = List.of("tag1","tag2");
+        final List<Tag> tagsToCreate = createTags(tagNames);
+        given(taggingServiceMock.addTags(any(), any())).willAnswer(invocation -> createTagAndNodeRefPairs(invocation.getArgument(1)));
+        given(taggingServiceMock.findTaggedNodesAndCountByTagName(any())).willReturn(Collections.emptyList());
+        final List<Tag> actualCreatedTags = objectUnderTest.addTags(nodesMock.getNode(any()).getNodeId(),tagsToCreate);
+        then(taggingServiceMock).should().addTags(TAG_NODE_REF, tagNames);
+        final List<Tag> expectedTags = createTagsWithNodeRefs(tagNames);
+        assertThat(actualCreatedTags)
+                .isNotNull()
+                .isEqualTo(expectedTags);
```
tpage-alfresco, line 344: This test should not be passing, because the count was not set in the tags. Please look to see if we can reasonably include the count in the `equals()` method of `Tag`.

krdabrowski: Nice catch Tom! 👍

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -279,7 +285,7 @@ public List<Tag> createTags(final StoreRef storeRef, final List<Tag> tags, final
 			.peek(tag -> {
 				if (parameters.getInclude().contains(PARAM_INCLUDE_COUNT))
 				{
-					tag.setCount(0);
+					tag.setCount(1);
```
tpage-alfresco, line 288: This looks wrong to me. Either way please add a comment to explain why the count is being set to 1 or to 0.

suneet-gupta: The reason we have changed count from 0 to 1 as we are adding a tag here with noderef.

krdabrowski: This is wrong - this method creates orphan tags, which are not related with any content. This should be 0.

suneet-gupta:
Reverted the count to 0 as it was an orphan tag.

remote-api/src/main/java/org/alfresco/rest/api/model/Tag.java
```diff
@@ -130,6 +130,20 @@ public boolean equals(Object obj)
...
+		}
+		else if (!tag.equals(other.tag))
+			return false;
+		if (count == null)
+		{
+			if (other.count != null)
+				return false;
+		}
+		else if (!count.equals(other.count))
+			return false;
```
tpage-alfresco, line 146: Update to `equals` without corresponding update to `hashCode` but I think it's fine in this case.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -302,4 +314,34 @@ private static Tag createTagWithNodeRef(final String tagName)
...
+    }
+
+    @Test
+    public void testAddTagsToNodeWithResponseNotIndexed()
+    {
+        final List<String> tagNames = List.of("tag1","tag2");
+        final List<Tag> tagsToCreate = createTags(tagNames);
+        given(taggingServiceMock.addTags(any(), any())).willAnswer(invocation -> createTagAndNodeRefPairs(invocation.getArgument(1)));
+        given(taggingServiceMock.findTaggedNodesAndCountByTagName(any())).willReturn(Collections.emptyList());
```
tpage-alfresco, line 338: Why is there no corresponding line in `testAddTagsToNode`?

tpage-alfresco, line 338: Why is there no corresponding line in `testAddTagsToNode`?

suneet-gupta: Agreed. A nice catch. It's not required here.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -302,4 +314,33 @@ private static Tag createTagWithNodeRef(final String tagName)
...
+        final List<Tag> actualCreatedTags = objectUnderTest.addTags(nodesMock.getNode(any()).getNodeId(),tagsToCreate);
+        then(taggingServiceMock).should().addTags(TAG_NODE_REF, tagNames);
+        final List<Tag> expectedTags = createTagsWithNodeRefs(tagNames).stream().peek(tag -> tag.setCount(1)).collect(Collectors.toList());;
+        assertThat(actualCreatedTags)
+                .isNotNull()
+                .isEqualTo(expectedTags);
+    }
+
+    @Test
+    public void testAddTagsToNodeWithResponseNotIndexed()
```
tpage-alfresco, line 333: This test is now identical to the previous test. Please fix the tests to do what they're meant to do.

suneet-gupta:
will go through the changes to fix the unidentified issues. will raise a new PR after fixing all issues.

PR #1785, opened 2023-03-02
ACS-3975 Fix Intermittent Test Failure

mikolajbrzezinski:


dsibilio (changes requested):
Please also remove the `Ignore` import if it's not unused as I suspect.

dsibilio:
P.S.: let's wait for the new `7.4.0-Mx` release to be completed before merging this change.

damianujma (approved):

mikolajbrzezinski:
The test was not observed to fail on the test branches, so for now it's being reintroduced. If any failures reappear then we can try to fix it, but for now it is working fine

PR #1793, opened 2023-03-09
[ACS-3917][ACS-3916] Added tagById Validation in get and remove method

pksingh41:
Added tagById Validation in get and remove method


remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -188,24 +188,20 @@ public CollectionWithPagingInfo<Tag> getTags(StoreRef storeRef, Parameters param
...
-    	if(tagNodeRef == null)
-    	{
-    		throw new EntityNotFoundException(tagId);
-    	}
-    	return tagNodeRef;
-    }
-
+		NodeRef tagNodeRef = nodes.validateNode(storeRef,tagId);
+		String tagName = taggingService.getTagName(tagNodeRef);
+		if( tagNodeRef == null || !taggingService.isTag(tagNodeRef.getStoreRef(), tagName))
```
tpage-alfresco, line 199: Can we improve this? Currently we are doing the following:
* Check there is a node for `storeRef` and `tagId` and get the `tagNodeRef`.
* Use `tagNodeRef` to get the name of that node, ignoring whether it's actually a tag or not.
* Check if `tagNodeRef` was null.
* Get the store ref from `tagNodeRef`.
* Use search to check if there's a node with the `cm:taggable` aspect and the given `tagName`.

This is pretty complex and I don't even think this works properly as it is. If we have a tag called "apples" and call this method with a node id for a folder called "apples" then I think it will say the folder is a valid tag.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -31,7 +31,7 @@
 import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
-import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.*;
```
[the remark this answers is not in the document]
pksingh41: Agreed Addressed the review comment

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: Addressed review comment

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -97,6 +100,13 @@ public void testGetTags() {
         assertEquals(expectedTags, actualCreatedTags);
     }
 
+    @Test
+    public void testTagNotFoundValidation()
+    {
+        given(nodeServiceMock.hasAspect(any(),any())).willReturn(true);
+        assertThrows(EntityNotFoundException.class, () ->objectUnderTest.validateTag(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, TAG_ID));
```
tpage-alfresco, line 107: It's not clear to me why this method should throw the EntityNotFoundException. Please add comments or some more suggestive mock methods to explain this.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -97,6 +100,13 @@ public void testGetTags() {
         assertEquals(expectedTags, actualCreatedTags);
     }
 
+    @Test
+    public void testTagNotFoundValidation()
+    {
+        given(nodeServiceMock.hasAspect(any(),any())).willReturn(true);
```
tpage-alfresco, line 106: I'm sure we don't really want to test a node that has every aspect. I think there's actually a single node and a single aspect we're interested in here. In cases like this then please can we focus on what's actually relevant to the test case?

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -141,8 +151,6 @@ public void testDeleteTagById_nonExistentTag()
 
         then(nodesMock).should().validateNode(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE, "dummy-id");
         then(nodesMock).shouldHaveNoMoreInteractions();
-
-        then(taggingServiceMock).shouldHaveNoInteractions();
```
tpage-alfresco, line 145: This seems like a sensible assertion for a test involving a non-existent tag. Why did we remove it? What method in `taggingService` gets called now?

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -188,24 +195,19 @@ public CollectionWithPagingInfo<Tag> getTags(StoreRef storeRef, Parameters param
...
     
     public NodeRef validateTag(StoreRef storeRef, String tagId)
     {
-    	NodeRef tagNodeRef = nodes.validateNode(storeRef, tagId);
-    	if(tagNodeRef == null)
-    	{
-    		throw new EntityNotFoundException(tagId);
-    	}
-    	return tagNodeRef;
+        NodeRef tagNodeRef = nodes.validateNode(storeRef,tagId);
```
tpage-alfresco, line 204: This looks wrong to me. In both places that you've added a call to this method we're now calling:
```
NodeRef tagNodeRef = nodes.validateNode(tagId)
NodeRef tagNodeRef = nodes.validateNode(storeRef,tagId);
```
(The first call outside this method and second call inside this method).

Please could you add some justification if you think this is correct?

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -188,24 +195,19 @@ public CollectionWithPagingInfo<Tag> getTags(StoreRef storeRef, Parameters param
...
+        if (tagNodeRef == null || nodeService.hasAspect(tagNodeRef, ContentModel.ASPECT_TAGGABLE))
```
tpage-alfresco, line 205: I don't understand what the aim behind this change is.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -244,8 +246,11 @@ public Tag getTag(StoreRef storeRef, String tagId)
 
     public CollectionWithPagingInfo<Tag> getTags(String nodeId, Parameters params)
     {
-		NodeRef nodeRef = validateTag(nodeId);
-
+		NodeRef nodeRef = nodes.validateNode(nodeId);
+		if( nodeRef == null )
+		{
+			throw new EntityNotFoundException(nodeId);
```
tpage-alfresco, line 252: Doesn't `nodes.validateNode` already throw this?

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -244,8 +246,11 @@ public Tag getTag(StoreRef storeRef, String tagId)
 
     public CollectionWithPagingInfo<Tag> getTags(String nodeId, Parameters params)
     {
-		NodeRef nodeRef = validateTag(nodeId);
-
+		NodeRef nodeRef = nodes.validateNode(nodeId);
```
tpage-alfresco, line 249: This looks like an improvement (assuming the methods do what I think). I'd expect `validateTag(nodeId)` to want the nodeId for a tag, rather than a document/folder.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Remove the test case as there was a preexisting EntityNotFoundException test case. So removed the duplicate test case

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Remove the test case, it is no longer required.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Yes it was by mistake, we reverted the changes.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: Yes in current implementation we are calling Node validation method twice, So we have now completely independent
implementation for both the methods .

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: In this change, we are checking cm:taggable aspect to decide whether a node is tag or not.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: Reverted the above changes.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: Yes, that's true

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
krdabrowski: I'm also confused with this check - you are telling here: if node doesn't exist or if it is a tag then throw an exception. You probably forgot about the NOT part.

suneet-gupta: As per QA testing, it's providing correct result. Please let us know the scenario if it's not working correctly

krdabrowski: Ohh, now I understand it - or at least I think so - you are checking here if node exists or if it's the root tag, am I right?

pksingh41: As mentioned above, In this change we are checking cm:taggable aspect to decide whether a node is tag or not.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
MateuszHY: Why it is no longer required? Do we have any other test cases to cover those changes?

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
krdabrowski: Ok @pksingh41, so if a node has this `cm:taggable` aspect then it is not a tag, yes?

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Yes, there is a preexisting test case which cover above scenario. The test case name is "testDeleteTagById_nonExistentTag()".

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
pksingh41: Can you please help us with the scenario where it may fail. We will try to incorporate the required fix.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
tpage-alfresco: This can't really be the case though. You've made a change to the production code so either that test doesn't check your change, or perhaps it's disabled.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
tpage-alfresco: Can you please explain the logic you're trying to write here. We will then be able to help tell you if this really does what you think it does.

pksingh41: In this change we are checking cm:taggable aspect to decide whether a node is tag or not. Please suggest if this is not a correct approach.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Yes, we will add a new test case for the PR changes.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
tpage-alfresco: ...and if it is a tag then you throw an exception? Or if it's not a tag then you throw an exception? Please can you explain a bit more thoroughly what the if statement does.

pksingh41: Only a node which is not a tag can have cm:taggable aspect So we are throwing exception only when a node is not a tag
and that's the validation we are making here.

krdabrowski: I think that a node which is not a tag and is not tagged yet (simple file for example) could successfully pass this verification - it is a valid node and has not `cm:taggable` aspect.

krdabrowski: But maybe I'm going to deep into it. In general you are right @pksingh41 that tags shouldn't have this aspect present.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Added a new test case for getTagByIdNotFound validation. Could you please review and approve the PR if no more changes required. Thanks !

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
tpage-alfresco: > Only a node which is not a tag can have cm:taggable aspect

This doesn't sound like the right logic though. You care whether the node is a tag or not.

This is like:
* I want to check if something is edible
* I know that anything made of concrete is not edible
* The logic for something being edible: `if (object != null || object.isMadeOfConcrete()) { throw NotEdibleException(); }`

pksingh41: Can you please suggest the correct approach to validate the tag Id. We will make the required changes.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -27,37 +27,26 @@
...
 import org.alfresco.rest.framework.core.exceptions.PermissionDeniedException;
-import org.alfresco.rest.framework.resource.parameters.CollectionWithPagingInfo;
-import org.alfresco.rest.framework.resource.parameters.Paging;
 import org.alfresco.rest.framework.resource.parameters.Parameters;
-import org.alfresco.rest.framework.resource.parameters.where.InvalidQueryException;
-import org.alfresco.rest.framework.tools.RecognizedParamsExtractor;
-import org.alfresco.service.cmr.repository.DuplicateChildNodeNameException;
-import org.alfresco.service.cmr.repository.NodeRef;
-import org.alfresco.service.cmr.repository.StoreRef;
+import org.alfresco.service.cmr.repository.*;
```
tpage-alfresco, line 49: Please try not to use wildcard imports.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -97,145 +86,38 @@
...
-        final List<Tag> expectedTags = createTagsWithNodeRefs(List.of(TAG_NAME)).stream()
+        final List<Tag> actualCreatedTags = objectUnderTest.createTags(tagsToCreate, parametersMock);
+        final List<Tag> expectedTags = createTagsWithNodeRefs(tagNames).stream()
             .peek(tag -> tag.setCount(0))
             .collect(Collectors.toList());
-        assertEquals(expectedTags, actualTags.getCollection());
-    }
-
-    @Test
-    public void testGetTags_withEqualsClauseWhereQuery()
```
tpage-alfresco, line 139: I don't understand why you've deleted all these tests - was that deliberate?

PR #1806, opened 2023-03-16
ACS-4025: Support include count and orderBy count for GET /tags

gevagel:
This PR introduces changes to support `?orderBy=count` when the count is included for `GET /tags` API endpoint. The `TagsImpl` method logic has been changed in order to follow the structure suggested in a comment in the associated JIRA issue [link](https://alfresco.atlassian.net/browse/ACS-4025).


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -72,6 +77,130 @@ public void getTagsWithConsumerRole()
...
 
+    /**
+     * Include count in the query parameters and ensure count is as expected for returned tags.
+     */
+    @Test(groups = { TestGroup.REST_API, TestGroup.TAGS, TestGroup.REGRESSION })
+    public void testGetTags_withIncludeCount()
+    {
+        STEP("Get tags including count filter and ensure count is as expected for returned tags");
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+                .withParams("include=count")
```
tpage-alfresco, line 88: You can use `include("count")` instead of this if you like.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -72,6 +77,130 @@ public void getTagsWithConsumerRole()
...
+    /**
+     * Get tags and order results by count. Default sort order should be ascending
+     */
+    @Test(groups = { TestGroup.REST_API, TestGroup.TAGS, TestGroup.REGRESSION })
+    public void testGetTags_withOrderByCountDefaultOrderShouldBeAsc()
+    {
+
+        STEP("Get tags and order results by count. Default sort order should be ascending");
+        returnedCollection = restClient.authenticateUser(adminUserModel)
+                .withParams("include=count&orderBy=count")
```
tpage-alfresco, line 111: Out of interest what happens if count is not included, but we request to sort by it? I'm not sure if we need a test for this, but we should document the behaviour in the API explorer.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
```diff
@@ -194,7 +323,7 @@ public void useHighSkipCountCheckPagination()
                 .and().field("hasMoreItems").is("false")
                 .and().field("count").is("0")
                 .and().field("skipCount").is(20000)
-                .and().field("totalItems").is(0);
+                .and().field("totalItems").is(4);
```
tpage-alfresco, line 326: Maybe even better to delete this line of the assertion?

azakrzewski-hy (approved):
LGTM

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/tags/GetTagsTests.java
gevagel: The way I wrote the code, we can only sort by count when count is also included. That being said, do you think that when count isn't included and we request to sort by it, we should let the user know why it isn't possible?

tpage-alfresco: That would be ideal. This is an edge case, but it would be good to document the behaviour and provide a helpful error message.

repository/src/main/java/org/alfresco/service/cmr/tagging/TaggingService.java
```diff
@@ -88,10 +89,7 @@
      * @return PagingResults
      */
     @NotAuditable
-    default PagingResults<Pair<NodeRef, String>> getTags(StoreRef storeRef, PagingRequest pagingRequest, Collection<String> exactNamesFilter, Collection<String> alikeNamesFilter)
-    {
-        return new EmptyPagingResults<>();
-    }
+    List<Pair<NodeRef, String>> getTags(StoreRef storeRef, PagingRequest pagingRequest, Collection<String> exactNamesFilter, Collection<String> alikeNamesFilter);
```
mpichura, line 92: See my comment for above method

repository/src/main/java/org/alfresco/service/cmr/search/CategoryService.java
```diff
@@ -122,7 +122,7 @@ public enum Depth {IMMEDIATE, ANY};
      * @param sortByName boolean
      */
     @Auditable(parameters = {"storeRef", "aspectName", "pagingRequest", "sortByName"})
-    PagingResults<ChildAssociationRef> getRootCategories(StoreRef storeRef, QName aspectName, PagingRequest pagingRequest, boolean sortByName);
+    Collection<ChildAssociationRef> getRootCategories(StoreRef storeRef, QName aspectName, PagingRequest pagingRequest, boolean sortByName);
```
mpichura, line 125: As I commented for `TaggingService` - I would strongly prefer adding methods than changing their signatures.

repository/src/main/java/org/alfresco/service/cmr/search/CategoryService.java
```diff
@@ -135,7 +135,7 @@ public enum Depth {IMMEDIATE, ANY};
      * @return
      */
     @Auditable(parameters = {"storeRef", "aspectName", "pagingRequest", "sortByName", "filter"})
-    PagingResults<ChildAssociationRef> getRootCategories(StoreRef storeRef, QName aspectName, PagingRequest pagingRequest, boolean sortByName, String filter);
+    Collection<ChildAssociationRef> getRootCategories(StoreRef storeRef, QName aspectName, PagingRequest pagingRequest, boolean sortByName, String filter);
```
mpichura, line 138: See my comment for the above method.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -158,36 +161,33 @@ public void deleteTagById(StoreRef storeRef, String tagId) {
...
-            {
-                for (Pair<String, Integer> tagByCountElem : tagsByCount)
-                {
-                    tagsByCountMap.put(tagByCountElem.getFirst(), tagByCountElem.getSecond());
-                }
-            }
-            tags.forEach(tag -> tag.setCount(Optional.ofNullable(tagsByCountMap.get(tag.getTag())).orElse(0)));
-        }
+		Map<Integer, Collection<String>> namesFilters = resolveTagNamesQuery(params.getQuery());
+		List<Pair<NodeRef, String>> results = taggingService.getTags(storeRef, Util.getPagingRequest(paging), namesFilters.get(EQUALS), namesFilters.get(MATCHES));
```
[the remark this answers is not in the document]
mpichura: To some extent I agree with Tom - the sorting must be done before paging. 
The question is whether it should be done inside this method or service method is not very relevant to me as in both cases we seem do be doing  the sorting in memory on the whole result set. 
So it's either getting all the tags (call the service method to get all tags for `StoreRef`) here and then sort and page them or deal with that on the service method side.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -158,36 +161,33 @@ public void deleteTagById(StoreRef storeRef, String tagId) {
 		taggingService.deleteTag(storeRef, tagValue);
 	}
 
-    @Override
+	@Override
```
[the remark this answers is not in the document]
tpage-alfresco: At some point we should reformat this entire file to get rid of the tabs. It's a nightmare in every PR :sweat_smile:

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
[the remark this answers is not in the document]
tpage-alfresco: We could do it in memory after the service call, but I thought that if we do it inside the service call then we're one step closer to replacing it with a DB query that handles these things without loading all tags in the system. I'm happy with either approach as long as the paging actually works with custom sort orders.

remote-api/src/main/java/org/alfresco/rest/api/impl/TagsImpl.java
```diff
@@ -335,4 +336,25 @@ private NodeRef checkTagRootAsNodePrimaryParent(String tagId, NodeRef tagNodeRef
 		}
 		return tagNodeRef;
 	}
+
+	private List<Tag> sortTags(List<Tag> tagsList, String sortBy, boolean sortAsc)
```
mpichura, line 340: Is this method ever used?

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -121,13 +127,14 @@ public void setup()
     public void testGetTags()
     {
         given(parametersMock.getPaging()).willReturn(pagingMock);
-        given(taggingServiceMock.getTags(any(StoreRef.class), any(PagingRequest.class), any(), any())).willReturn(pagingResultsMock);
-        given(pagingResultsMock.getTotalResultCount()).willReturn(new Pair<>(Integer.MAX_VALUE, 0));
-        given(pagingResultsMock.getPage()).willReturn(List.of(new Pair<>(TAG_NODE_REF, TAG_NAME)));
+        given(parametersMock.getSorting()).willReturn(new ArrayList<>());
```
mpichura, line 130: I think `Collections.emptyList()` would be more meaningful here and in all other places you've used `(new ArrayList<>()` to mock `include` and `sorting`.

repository/src/main/java/org/alfresco/service/cmr/tagging/TaggingService.java
```diff
@@ -327,8 +330,15 @@ default PagingResults<Pair<NodeRef, String>> getTags(StoreRef storeRef, PagingRe
...
-    
-    
+
+    /**
+     *
+     * @param storeRef              node reference
+     * @param parametersInclude     the 'include' part of the parameters
+     * @return List of pairs of tag nodeRefs and tag names
+     */
+    List<Pair<NodeRef, String>> getTags(StoreRef storeRef, List<String> parametersInclude);
```
mpichura, line 340: Same here regarding possible use of `Map` instead of `List<Pair>`

repository/src/main/java/org/alfresco/service/cmr/search/CategoryService.java
```diff
@@ -150,11 +150,19 @@ public enum Depth {IMMEDIATE, ANY};
...
+                                                              Collection<String> exactNamesFilter, Collection<String> alikeNamesFilter) {
         return new EmptyPagingResults<>();
     }
 
+    /**
+     * Get a collection of the root categories for an aspect/classification supporting multiple name filters.
+     */
+    @Auditable(parameters = {"storeRef", "aspectName", "exactNamesFilter", "alikeNamesFilter"})
+    default Collection<ChildAssociationRef> getRootCategories(StoreRef storeRef, QName aspectName, Collection<String> exactNamesFilter, Collection<String> alikeNamesFilter) {
+        return null;
```
mpichura, line 162: It woluld be better if `emptyList` (or `emptySet` but I see that the impementation returns `List`) was returned here instead of `null`.

mpichura (commented):
I had some questions and remarks. 
Also, as an aside - please check curly brackets formatting in some places.

tpage-alfresco:
Late to the party, but on the topic of modifying the interfaces then I agree with Maciej about TaggingService and CategoryService as these are both marked `@AlfrescoPublicApi`. For interfaces without this annotation then we should feel free to modify the interface as we feel is best.

MateuszHY (approved):

krdabrowski (approved):

repository/src/main/java/org/alfresco/service/cmr/tagging/TaggingService.java
```diff
@@ -94,6 +95,18 @@ default PagingResults<Pair<NodeRef, String>> getTags(StoreRef storeRef, PagingRe
     {
         return new EmptyPagingResults<>();
     }
+
+    /**
+     * Get a map of tags filtered by name and sorted by tag name or count
```
mpichura, line 100: Minor one - can be fixed in another iteration - javadoc here should explain what key and values are (either in general method description on in `@return` part.

repository/src/main/java/org/alfresco/service/cmr/tagging/TaggingService.java
```diff
@@ -362,6 +374,13 @@ default List<Pair<String, NodeRef>> createTags(StoreRef storeRef, List<String> t
     {
         return Collections.emptyList();
     }
+
+    /**
+     *
+     * @param storeRef
+     * @return a map with each tag name and its count
```
mpichura, line 381: Minor - can come in another iteration.
```suggestion
     * @return a map with each tag name and its usage count

```

mpichura (approved):
Had some very minor comments - they may be resolved in another iteration.

PR #1810, opened 2023-03-17
ACS-4857: Lookup for root tag in DB and NOT using search engine

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-4857

### DIFF
--- a/amps/share-services/pom.xml
+++ b/amps/share-services/pom.xml
@@ -32,10 +32,6 @@
                     <groupId>javax.servlet</groupId>
                     <artifactId>jstl</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>commons-collections</groupId>
-                    <artifactId>commons-collections</artifactId>
-                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
@@ -102,10 +98,6 @@
                     <groupId>javax.servlet</groupId>
                     <artifactId>jstl</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>commons-collections</groupId>
-                    <artifactId>commons-collections</artifactId>
-                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


amps/share-services/pom.xml
```diff
@@ -32,10 +32,6 @@
                     <groupId>javax.servlet</groupId>
                     <artifactId>jstl</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>commons-collections</groupId>
-                    <artifactId>commons-collections</artifactId>
```
krdabrowski, line 37: Removed exclusion due to failing test:
`Error:    ReadOnlyTransactionInGetSlingshotApiTest.testGetDoclibCategoryNode:208->BaseWebScriptTest.sendRequest:321->BaseWebScriptTest.sendRequest:372 Status code 500 returned, but expected 200 for /slingshot/doclib/categorynode/node/workspace/SpacesStore/61b4570b-f000-42d5-b17a-e8528198673d (get)`
From stack trace:
`Caused by: java.lang.ClassNotFoundException: org.apache.commons.collections.CollectionUtils`

mpichura: I think these exclusions may have been important as we seem to be setting the version of `commons-collections` in the main POM' `dependencyManagement` explicitly (and we set it to the newest version which is vulnerability free). 
On the other hand - it doesn't look that your changes have introduced this error.
I'd say that we need to take care of that in Webscripts project but maybe exclusion on that Webscript dependency with `test` scope would be enough?

mpichura (changes requested):

amps/share-services/pom.xml
krdabrowski: Dependency scope changed to: `commons-collections:commons-collections:jar:3.2.2:test`

tpage-alfresco (approved):
LGTM

PR #1820, opened 2023-03-21
[PRODSEC-6575] Changes done for validating expiry date for shared Link.

atkumar14:
https://alfresco.atlassian.net/browse/PRODSEC-6575


kavitshah-gl (approved):
Seems to be fine.

tpage-alfresco (changes requested):
It's not clear to me how this change fixes the ticket.

Please add unit tests to cover the new logic and an E2E test for the API calls that the FE is making.

atkumar14:
As per discussion with client, there is a design change in the front end to fix the above issue. Hence closing the PR.

PR #1822, opened 2023-03-22
ACS-4863 Allow referencing nodes by aliases for tag and category application.

tpage-alfresco:


remote-api/src/main/java/org/alfresco/rest/api/impl/CategoriesImpl.java
```diff
@@ -189,7 +189,7 @@ public void deleteCategoryById(final StoreRef storeRef, final String id, final P
     @Override
     public List<Category> listCategoriesForNode(final String nodeId, final Parameters parameters)
     {
-        final NodeRef contentNodeRef = nodes.validateNode(nodeId);
+        final NodeRef contentNodeRef = nodes.validateOrLookupNode(nodeId, null);
```
mpichura, line 192: I assume that the reason to change the method to retrieve the `NodeRef` throughout the whole PR might be that it handles getting `NodeRef` more properly but passing `null` as a parameter doesn't look great (not only here - I know we do that in other places too). Maybe we should refactor `validateNode` method a bit instead - but that's jus  an idea - if that could be a risky effort - then just disregard this comment?

krdabrowski (approved):

krdabrowski:
Nice job with fixing this check for all tags and categories methods!

mpichura (approved):
LGTM - had a minor remark/idea but it's not crucial.

PR #1826, opened 2023-03-23
ACS-4863 Add method validateOrLookupNode without path.

tpage-alfresco:


mpichura (approved):
Great job

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -649,6 +649,11 @@ private NodeRef getParentNodeRef(NodeRef nodeRef)
         return nodeService.getPrimaryParent(nodeRef).getParentRef();
     }
 
+    public NodeRef validateOrLookupNode(String nodeId)
```
mpichura, line 652: That could as well be a default method in the interface.

tpage-alfresco: Another good idea - thanks.

PR #1833, opened 2023-03-27
[ACS-4636] create tag count param support

pksingh41:
https://alfresco.atlassian.net/browse/ACS-4636


remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
```diff
@@ -453,7 +453,7 @@ public void testAddTags()
         List<Tag> tags = tagNames.stream().map(name -> Tag.builder().tag(name).create()).collect(toList());
         given(taggingServiceMock.addTags(CONTENT_NODE_REF, tagNames)).willReturn(pairs);
```
kavitshah-gl, line 455: given(parametersMock.getInclude()).willReturn(List.of("count")); should be added to initialize the mock value for count.

kavitshah-gl (changes requested):
Looks fine with small change required in testing class.

remote-api/src/test/java/org/alfresco/rest/api/impl/TagsImplTest.java
pksingh41: Done

kavitshah-gl (approved):
Seems to be okay.

PR #1847, opened 2023-03-31
ACS-4932 Fix issues around renaming tags.

tpage-alfresco:
This change refactors the existing code so that:
* The tag id is not changed by the rename
* The tag is renamed in the index (Solr or ES)
* All documents are updated with the new tag name in the index (Solr or ES)
* The tag scopes are updated.


tpage-alfresco:
This PR hopefully also addresses https://alfresco.atlassian.net/browse/ACS-3915

azakrzewski-hy (approved):
LGTM

PR #1864, opened 2023-04-06
[ACS-4462]  Add connection manager builder for httpClient4Factory

kmagdziarz:


core/src/main/java/org/alfresco/httpclient/HttpClient4Factory.java
```diff
@@ -23,17 +23,23 @@
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
 
+import java.util.Objects;
+
```
mstrankowski, line 27:
```suggestion
```

PR #1875, opened 2023-04-13
Set property values to the ones being set for sharedFileStoreConnector

mstrankowski:
since the only difference is the number of connections, which doesn't work in community-repo due to how the http clients are created per request of transform

### DIFF
--- a/repository/src/main/resources/alfresco/repository.properties
+++ b/repository/src/main/resources/alfresco/repository.properties
@@ -752,8 +752,8 @@ encryption.ssl.truststore.keyMetaData.location=
 ## HttpClient config for Transform Service
 #enable mtls in HttpClientFactory for Transform Service.
 httpclient.config.transform.mTLSEnabled=false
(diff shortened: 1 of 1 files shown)


repository/src/main/resources/alfresco/repository.properties
```diff
@@ -752,8 +752,8 @@ encryption.ssl.truststore.keyMetaData.location=
 ## HttpClient config for Transform Service
 #enable mtls in HttpClientFactory for Transform Service.
 httpclient.config.transform.mTLSEnabled=false
-httpclient.config.transform.maxTotalConnections=40
-httpclient.config.transform.maxHostConnections=40
+httpclient.config.transform.maxTotalConnections=20
+httpclient.config.transform.maxHostConnections=20
```
mstrankowski, line 756: I'm adjusting the common config to the values from sharedFileStoreConnector, since those values here aren't even used properly due to how httpClients are created.

kmagdziarz (approved):

PR #1989, opened 2023-06-13
[ACS-5392] migrated to http5 in IdentityServiceFacadeFactoryBean

kcichonczyk:


pom.xml
```diff
@@ -71,6 +71,8 @@
         <dependency.gson.version>2.8.9</dependency.gson.version>
         <dependency.httpclient.version>4.5.13</dependency.httpclient.version>
         <dependency.httpcore.version>4.4.16</dependency.httpcore.version>
+        <dependency.httpcomponents-httpclient5.version>5.2.1</dependency.httpcomponents-httpclient5.version>
```
pzhyland, line 74: Can we use one property for both dependencies?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -232,7 +235,7 @@ private IdentityServiceFacade createIdentityServiceFacade()
 
             //ACS-5392
```
pzhyland, line 236: Please remove the comment.

pzhyland (approved):

pom.xml
kcichonczyk: They tended to differ in the past (httpcore vs httpclient) so I would suggest leaving as it is.

PR #2035, opened 2023-07-03
ACS-5506 Add description and hasSubgroups to groups API

MichalKinas:
See https://alfresco.atlassian.net/browse/ACS-5506

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
@@ -118,7 +118,7 @@ public class GroupsTests extends RestTest
         restClient.assertStatusCodeIs(HttpStatus.OK);
 
         //AddChildGroup
-        restClient.authenticateUser(adminUser).withCoreAPI().usingParams("include=zones").usingGroups().createGroup(subgroupBodyCreate);
+        restClient.authenticateUser(adminUser).withCoreAPI().usingParams().usingGroups().createGroup(subgroupBodyCreate);
         restClient.assertStatusCodeIs(HttpStatus.CREATED);
         JsonObject groupMembershipGroupBody = Json.createObjectBuilder().add("id", "GROUP_"+subGroupName).add("memberType", "GROUP").build();
         String groupMembershipGroupBodyCreate = groupMembershipGroupBody.toString();
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
@@ -120,8 +120,7 @@ public class GroupsTests extends RestTest
 
         //ListPersonMembership
         restClient.authenticateUser(userModel).withCoreAPI().usingUser(userModel).listGroupMemberships()
-                                              .assertThat().entriesListContains("id", "GROUP_"+groupName)
-                                              .and().entriesListContains("id", "GROUP_"+subGroupName);
+                                              .assertThat().entriesListContains("id", "GROUP_"+groupName);
         restClient.assertStatusCodeIs(HttpStatus.OK);
 
         //CheckListDetails
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
@@ -105,7 +105,7 @@ public class GroupsTests extends RestTest
 
         JsonObject groupMembershipBody = Json.createObjectBuilder().add("id", userModel.getUsername()).add("memberType", "PERSON").build();
         String groupMembershipBodyCreate = groupMembershipBody.toString();
-        JsonObject groupMembershipGroupBody = Json.createObjectBuilder().add("id", subGroupName).add("memberType", "GROUP").build();
+        JsonObject groupMembershipGroupBody = Json.createObjectBuilder().add("id", "GROUP_"+subGroupName).add("memberType", "GROUP").build();
         String groupMembershipGroupBodyCreate = groupMembershipGroupBody.toString();
 
         //MembershipCreation:
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
@@ -45,7 +45,7 @@ public class GroupsTests extends RestTest
         restClient.authenticateUser(userModel).withCoreAPI().usingGroups().createGroup(groupBodyCreate);
         restClient.assertStatusCodeIs(HttpStatus.FORBIDDEN);
         //+ve
-        restClient.authenticateUser(adminUser).withCoreAPI().usingParams("include=zones").usingGroups().createGroup(groupBodyCreate)
+        restClient.authenticateUser(adminUser).withCoreAPI().usingParams("include=zones,hasSubgroups,description").usingGroups().createGroup(groupBodyCreate)
                                               .assertThat().field("zones").contains("APP.DEFAULT")
                                               .and().field("isRoot").is(true)
                                               .and().field("displayName").is(groupName)
@@ -82,7 +82,7 @@ public class GroupsTests extends RestTest
         restClient.assertStatusCodeIs(HttpStatus.OK);
 
         //GetGroupDetails:
-        restClient.withCoreAPI().usingParams("include=zones").usingGroups().getGroupDetail("GROUP_"+groupName)
+        restClient.withCoreAPI().usingParams("include=zones,hasSubgroups,description").usingGroups().getGroupDetail("GROUP_"+groupName)
                   .assertThat().field("id").is("GROUP_"+groupName)
                   .and().field("zones").contains("APP.DEFAULT")
                   .and().field("isRoot").is(true)
@@ -94,7 +94,7 @@ public class GroupsTests extends RestTest
         restClient.assertStatusCodeIs(HttpStatus.NO_CONTENT);
 
         //VerifyIfParentHasNoSubgroups:
-        restClient.withCoreAPI().usingParams("include=zones").usingGroups().getGroupDetail("GROUP_"+groupName)
+        restClient.withCoreAPI().usingParams("include=zones,hasSubgroups").usingGroups().getGroupDetail("GROUP_"+groupName)
                 .assertThat().field("id").is("GROUP_"+groupName)
                 .and().field("hasSubgroups").is(false);
         restClient.assertStatusCodeIs(HttpStatus.OK);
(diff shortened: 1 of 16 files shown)


remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -852,7 +887,7 @@ public void deleteGroupMembership(String groupId, String groupMemberId)
 
         // Verify if groupMemberId is member of groupId
         AuthorityType authorityType = AuthorityType.getAuthorityType(groupMemberId);
-        Set<String> parents = authorityService.getContainingAuthorities(AuthorityType.GROUP, groupMemberId, true);
```
mpichura, line 855: I'm not sure if this is correct.  I'd say that `authorityType` variable is unnecessary here and could be removed,

mpichura (commented):
I've had some remarks and thin the code needs some changes.

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
MichalKinas: Right, removed

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -89,6 +94,7 @@ public class GroupsImpl implements Groups
     private static final String ZONE = "zone";
     private static final String AUTHORITY_NAME = "authorityName";
     private static final String ERR_MSG_MODIFY_FIXED_AUTHORITY = "Trying to modify a fixed authority";
+    private static final QName PROP_DESCRIPTION = QName.createQName(NamespaceService.CONTENT_MODEL_1_0_URI, "description");
```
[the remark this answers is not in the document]
MichalKinas: Removed

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -151,6 +167,12 @@ public Group create(Group group, Parameters parameters)
             authorityService.addAuthority(group.getParentIds(), authority);
         }
 
+        if (group.getDescription() != null && !group.getDescription().isEmpty())
```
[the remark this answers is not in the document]
MichalKinas: Changed

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -584,6 +612,13 @@ private Group getGroup(AuthorityInfo authorityInfo, List<String> includeParam, S
         group.setDisplayName(authorityDisplayName);
 
         group.setIsRoot(isRootAuthority(rootAuthorities, authorityInfo.getAuthorityName()));
+        group.setHasSubgroups(!authorityService.getContainedAuthorities(AuthorityType.GROUP, authorityInfo.getAuthorityName(), true).isEmpty());
+
+        NodeRef groupNodeRef = authorityService.getAuthorityNodeRef(authorityInfo.getAuthorityName());
+        String description = nodeService.getProperty(groupNodeRef, PROP_DESCRIPTION) != null ?
+                nodeService.getProperty(groupNodeRef, PROP_DESCRIPTION).toString() :
+                "";
```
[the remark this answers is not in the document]
MichalKinas: Changed to `null`

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -168,6 +190,12 @@ public Group update(String groupId, Group group, Parameters parameters)
             handleAuthorityException(ae);
         }
 
+        if (group.getDescription() != null && !group.getDescription().isEmpty())
```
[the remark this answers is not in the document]
MichalKinas: Done

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -151,6 +167,12 @@ public Group create(Group group, Parameters parameters)
             authorityService.addAuthority(group.getParentIds(), authority);
         }
 
+        if (group.getDescription() != null && !group.getDescription().isEmpty())
+        {
+            NodeRef groupNodeRef = authorityService.getAuthorityNodeRef(authority);
+            nodeService.setProperty(groupNodeRef, PROP_DESCRIPTION, group.getDescription());
```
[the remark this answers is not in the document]
MichalKinas: Right, added changes to `AuthorityService`

remote-api/src/main/java/org/alfresco/rest/api/impl/GroupsImpl.java
```diff
@@ -584,6 +613,13 @@ private Group getGroup(AuthorityInfo authorityInfo, List<String> includeParam, S
         group.setDisplayName(authorityDisplayName);
 
         group.setIsRoot(isRootAuthority(rootAuthorities, authorityInfo.getAuthorityName()));
+        group.setHasSubgroups(!authorityService.getContainedAuthorities(AuthorityType.GROUP, authorityInfo.getAuthorityName(), true).isEmpty());
```
mpichura, line 616: I'm pretty sure that code added in lines 616-621 is causing the tests to fail. 
I've seen that `authorityService.getContainedAuthorities` as well as `nodeService.getProperty` can throw an exception which is not handled.
Secondly, calling `isEmpty()` on `Set` can cause NPE (line 616) - it's much safer to use `CollectionUtils.isNotEmpty` there.
I'll try to come with some more findings soon as I am almost ready with hints on how to bring this closer to a working solution.

MichalKinas: Really good catch thanks! Added required changes

tpage-alfresco (changes requested):
I don't like the asymmetry of this solution. We are relying on the `AuthorityDAO` to persist the group description, but I think the only way to retrieve it is with the `NodeService`. I think description is a reasonable field to add to the group object `AuthorityInfo` (providing it can continue to be unset) or alternatively we could just use the `NodeService` to set and retrieve it? (within a single transaction) Nb. `AuthorityInfo` is part of the `@AlfrescoPublicApi` and so we need to preserve the existing constructor.

The `hasSubgroups` flag requires extra service calls to populate and so should be an `include` parameter in the API.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/groups/GroupsTests.java
```diff
@@ -31,44 +31,72 @@ public void dataPreparation() throws Exception
     @Test(groups = { TestGroup.REST_API, TestGroup.GROUPS, TestGroup.SANITY })
     @TestRail(section = { TestGroup.REST_API, TestGroup.NODES }, executionType = ExecutionType.SANITY,
             description = "Verify creation, listing, updating and deletion of groups.")
-    public void createListUpdateAndDeleteGroup() throws Exception
-    {
-        String groupName = "ZtestGroup" + UUID.randomUUID().toString();
-        JsonObject groupBody = Json.createObjectBuilder().add("id", groupName).add("displayName", groupName).build();
+    public void createListUpdateAndDeleteGroup() {
```
tpage-alfresco, line 34: Generally in our Java projects we have opening braces on a new line. The two exceptions I know of are in the Elasticsearch Connector and in lambda expressions.

repository/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -378,27 +379,37 @@ public void addAuthority(Collection<String> parentNames, String childName)
...
+    @Override
+    public void createAuthority(String name, String authorityDisplayName, Set<String> authorityZones, Map<QName, Serializable> properties)
     {
-        HashMap<QName, Serializable> props = new HashMap<QName, Serializable>();
+        Map<QName, Serializable> props = new HashMap<>();
         /* MNT-11749 : Alfresco allows to create authorities with different char cases, but disallow duplicates */
         props.put(ContentModel.PROP_NAME, DigestUtils.md5Hex(name));
         props.put(ContentModel.PROP_AUTHORITY_NAME, name);
         props.put(ContentModel.PROP_AUTHORITY_DISPLAY_NAME, authorityDisplayName);
+        if (MapUtils.isNotEmpty(properties))
```
tpage-alfresco, line 395: Is this checking for `properties == null`? If not then we can just omit the if statement and always putAll the empty map.

repository/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -1431,7 +1442,31 @@ public void setAuthorityDisplayName(String authorityName, String authorityDispla
             return;
         }
         nodeService.setProperty(ref, ContentModel.PROP_AUTHORITY_DISPLAY_NAME, authorityDisplayName);
+    }
 
+    @Override
+    public Pair<String, String> getAuthorityDisplayNameAndDescription(String authorityName)
```
tpage-alfresco, line 1448: I think you could call through to `getAuthorityDisplayName` to reduce the duplication here.

repository/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -1431,7 +1442,31 @@ public void setAuthorityDisplayName(String authorityName, String authorityDispla
...
+    @Override
+    public void setAuthorityDisplayNameAndDescription(String authorityName, String authorityDisplayName, String description)
+    {
+        NodeRef ref = getAuthorityOrNull(authorityName);
+        if (ref == null)
+        {
+            return;
+        }
+        nodeService.setProperty(ref, ContentModel.PROP_AUTHORITY_DISPLAY_NAME, authorityDisplayName);
+        nodeService.setProperty(ref, ContentModel.PROP_DESCRIPTION, description);
```
tpage-alfresco, line 1469: There's a `nodeService` method to set several properties at the same time which should be faster than this.

tpage-alfresco (approved):
This looks much better now and I don't think it should have any performance implications for existing integrations (since both new fields are behind `include` flags). You mentioned looking at canned queries, which could improve the performance when loading both the description and display name.

LGTM :+1:

repository/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
MichalKinas: Okay I was actually considering doing so but then I think it would mean the second `getAuthorityOrNull` call in `getAuthorityDisplayName` would it be fine for us? I mean as far as I checked it uses `authorityLookupCache` so it might be fine what do you think?

repository/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
MichalKinas: Good point it actually should check `properties != null`, fixed now

tpage-alfresco (approved):
LGTM

PR #2039, opened 2023-07-04
[ACS-5547] jakarta-transaction-api version bump

kcichonczyk:


kcichonczyk:
Before merging, remember to replace:
`clone_and_install alfresco-aos-module feature/acs-5547_transaction_migration`
with 
`clone_and_install alfresco-aos-module`

PR #2071, opened 2023-07-14
ACS-5449 Bump mockito-core from 4.9.0 to 5.4.0

viepovsky:
First failure was caused by `CamelComponentsTest`class , there were problems with bean initialization from `test-messaging-context.xml`. Above mentioned class along with `CamelRoutesTest` class uses this xml file. ~~These classes had annotation `@Category(NeverRunsTests.class)`, and after adding `@Ignore` to deactivate them, now tests pass. (I tried many ways of fixing `test-messaging-context.xml` file but didn't find any solution, only `@Ignore` makes it pass).~~

Edit: As @damianujma found out, that we cannot delete these tests, I backed to work to find solution for failing tests.
Finally I have found a solution for these camel tests, the error which caused problems was : `No qualifying bean of type 'org.alfresco.repo.rawevents.TransactionAwareEventProducer' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}` and whatever I did it was not possible to mock certain classes (apart from `TransactionAwareEventProducer` class there were more exceptions for other mocked classes). Anyway after many tries, finally I inserted mocked beans on top of the application context xml file(`test-messaging.xml`) added type of argument and it does work now. 
What caused the problem? Upgrading `mockito-core` from `4.9.0` to `4.10.0` and above causes this. I found one releted issue which I think is connected with this, as someone also had this problem, but they did not find solution. Anyway the solution is just putting mocked beans on the top of xml file before scanning spring components. Here is link to the issue: https://github.com/mockito/mockito/pull/2779#issuecomment-1382439975 , there is also mentioned that now when mocking and using xml, the type of argument is needed(it is also connected with the last failure with `NodeLocatorServiceImplTest`)

Second failure with access modifier, `LockServicePoliciesImpl` class was private, and after changing it to public there are no more `IllegalAccessExceptions`.

Third failure connected with `RuleServiceImplUnitTest` class, which tests `RuleServiceImpl`, this class has got private field `SimpleCache<NodeRef, List<Rule>> nodeRulesCache;`. And in `RuleServiceImplUnitTest` was field with raw interface without declaration just `SimpleCache nodeRulesCache;`, this caused problems in exception: `java.lang.NullPointerException: Cannot invoke "org.alfresco.repo.cache.SimpleCache.remove(java.io.Serializable)" because "this.nodeRulesCache" is null`.
After adding types to interface tests are passing.

The last failure was connected with `NodeLocatorServiceImplTest` class due to its bean initialization file `test-nodeLocatorServiceImpl-context.xml`, the problem was with recognizing proper overloaded method, I had to provide argument type (`type=java.lang.Class`), so the compiler could know which method to call.

### DIFF
--- a/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
+++ b/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2022 Alfresco Software Limited
+ * Copyright (C) 2005 - 2023 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software.
  * If the software was purchased under a paid Alfresco license, the terms of
--- a/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
+++ b/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
@@ -43,7 +43,6 @@ import static org.alfresco.service.namespace.RegexQNamePattern.MATCH_ALL;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.nullable;
@@ -62,7 +61,6 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.alfresco.repo.action.RuntimeActionService;
-import org.alfresco.repo.cache.DefaultSimpleCache;
 import org.alfresco.repo.cache.SimpleCache;
 import org.alfresco.service.cmr.action.Action;
 import org.alfresco.service.cmr.action.ActionServiceException;
--- a/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
+++ b/repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
@@ -62,6 +62,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.alfresco.repo.action.RuntimeActionService;
+import org.alfresco.repo.cache.DefaultSimpleCache;
 import org.alfresco.repo.cache.SimpleCache;
 import org.alfresco.service.cmr.action.Action;
 import org.alfresco.service.cmr.action.ActionServiceException;
@@ -94,7 +95,7 @@ public class RuleServiceImplUnitTest
     @Mock
     private PermissionService permissionService;
     @Mock
-    private SimpleCache nodeRulesCache;
+    private SimpleCache<NodeRef, List<Rule>> nodeRulesCache;
     @Mock
     private NodeService runtimeNodeService;
     @Mock
@@ -127,7 +128,7 @@ public class RuleServiceImplUnitTest
         ChildAssociationRef ruleAssociation = mock(ChildAssociationRef.class);
         when(ruleAssociation.getChildRef()).thenReturn(RULE_NODE);
         when(nodeService.createNode(eq(RULE_SET_NODE), eq(ASSOC_CONTAINS), any(QName.class), eq(TYPE_RULE))).thenReturn(ruleAssociation);
-        doNothing().when(nodeRulesCache).remove(FOLDER_NODE);
+        doNothing().when(nodeRulesCache).remove(any());
         // Set the rule title and action.
         when(mockRule.getTitle()).thenReturn("Rule title");
         when(mockRule.getAction()).thenReturn(mockAction);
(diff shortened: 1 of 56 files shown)


dsibilio (commented):
Overall LGTM, good fixes.

Given that we're not running these Camel tests for years now as @tpage-alfresco [disabled them](https://github.com/Alfresco/alfresco-community-repo/commit/b9e4557973fdbe119c015377390e4608ccaa123e) since they weren't working anymore + we wouldn't be able to run them anyways with the latest Mockito because there's issues with bringing up the context which we didn't figure out yet, I wonder if we should just go ahead and remove these 2 Camel-related tests + the `test-messaging-context.xml` file that was only used by these 2 tests in the first place as far as I understand.

Also, this section of the CI doesn't really make any sense anymore since these 2 tests are not being run: https://github.com/Alfresco/alfresco-community-repo/blob/master/.github/workflows/ci.yml#L278-L297, and it should most likely be removed as well.

@tpage-alfresco do you have any thoughts on this?

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
```diff
@@ -43,16 +43,13 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
```
tpage-alfresco, line 46: PMD says this is unused

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
```diff
@@ -43,16 +43,13 @@
...
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.BDDMockito.then;
-import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
tpage-alfresco, line 52: PMD (wrongly) says this is unused, but I think the issue is the wildcard import.

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
```diff
@@ -65,6 +62,7 @@
 import java.util.stream.IntStream;
 
 import org.alfresco.repo.action.RuntimeActionService;
+import org.alfresco.repo.cache.DefaultSimpleCache;
```
tpage-alfresco, line 65: PMD says this is unused.

tpage-alfresco (commented):
Please can you review the new PMD issues

tpage-alfresco:
@dsibilio I'm in favour of removing those too :+1: Having tests that aren't run seems like a bad idea in the first place - probably I should have pushed for them to be removed rather than ignoring them.

tpage-alfresco (approved):
Other than the PMD findings then this looks good. As discussed we could also remove the unused test.

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
viepovsky: Yes, you are right, I have removed it

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
viepovsky: Yes, you are right, I have removed it

repository/src/test/java/org/alfresco/repo/rule/RuleServiceImplUnitTest.java
viepovsky: I leave it as it is because it is used in this class

dsibilio (approved):

dsibilio (changes requested):
LGTM but there's some more stuff we can clean up for sure.

Are all the files here https://github.com/Alfresco/alfresco-community-repo/tree/fix/ACS-5449_investigate-bump-mockito-core-test-failure/repository/src/test/java/org/alfresco/messaging/camel still used? Or are most if not all of them rendered useless as things stand now?
If they are useless they should be deleted as well.

Same most likely applies to https://github.com/search?q=repo%3AAlfresco%2Falfresco-community-repo%20testRoutes.xml&type=code.

viepovsky:
Yes, it seems like these classes inside package `messaging/camel` are unused. I have deleted them as well as delete `testRoutes.xml` which was also only connected with these `Camel` tests and everything passes after all deletions.

Removed classes from `messaging/camel` were called(apart from previous removed `Camel` classes):
`MockConsumer`, `MockExceptionProcessor`, `MockExceptionThrowingConsumer` and `TestingRouteBuilder` from `messaging/camel/routes` folder.

dsibilio (approved):

aonych (dismissed):
LGTM, good job!

damianujma (approved):

dsibilio (changes requested):
See https://github.com/Alfresco/alfresco-community-repo/actions/runs/5609730718/job/15198015845#step:7:474.

Thanks @damianujma for noticing that the Camel tests were actually being run anyways successfully before the Mockito upgrade... so we actually kind of need to maintain them and fix issues that come with the Mockito upgrade rather than removing the tests. The name for the `NeverRunsTests` category was misleading.

damianujma (changes requested):
Even though the tests are "NeverRunsTest" annotated, these tests are run actually (e.g. https://github.com/Alfresco/alfresco-community-repo/actions/runs/5609730718/job/15198015845#step:7:474). I think we shouldn't remove them.

mikolajbrzezinski (approved):

aonych (approved):

damianujma (approved):

PR #2085, opened 2023-07-25
MNT-23763 Fix image preview webscript for use with ES.

tpage-alfresco:
This also fixes the query logic in two other webscripts. The default boolean operator is OR but these webscripts require AND.


mpichura (approved):
LGTM

PR #2081, opened 2023-07-20
Add file name to DirectAccessUrl

canpan14:
This will cause it to be returned in the response

PR #2090, opened 2023-07-26
Add file name to DirectAccessUrl

tpage-alfresco:
This will cause it to be returned in the response


tpage-alfresco:
Copy of PR #2081 so that it can be built by the CI.

jottley:
@tpage-alfresco Is this PR still relevant or needed

PR #2093, opened 2023-07-27
ACS-5449 Add commons-net dependency to main pom.xml file

viepovsky:
Solution eliminating test failures in alfresco-enterprise-repo: https://github.com/Alfresco/alfresco-enterprise-repo/actions/runs/5669575699/job/15390101

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -386,7 +386,12 @@
                 <artifactId>commons-fileupload</artifactId>
                 <version>1.5</version>
             </dependency>
-
+            <dependency>
+                <groupId>commons-net</groupId>
+                <artifactId>commons-net</artifactId>
+                <version>3.9.0</version>
+                <scope>test</scope>
+            </dependency>
             <dependency>
                 <groupId>org.apache.httpcomponents</groupId>
                 <artifactId>httpclient</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -390,7 +390,6 @@
                 <groupId>commons-net</groupId>
                 <artifactId>commons-net</artifactId>
                 <version>3.9.0</version>
-                <scope>test</scope>
             </dependency>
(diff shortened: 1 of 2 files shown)


pzhyland (approved):

dsibilio (changes requested):
Why are we shifting a test dependency to compile scope? I don't think this is the right solution.

My understanding so far is that the `org.reflections` solution we're using in the enterprise-repository is finnicky and scans more than it should even though it's appropriately configured to scan only the `org.alfresco` package. I've found several bug mentioning that the package filtering is not working as intended: https://github.com/ronmamo/reflections/issues?q=is%3Aissue+is%3Aopen+package.

Ideally we should eventually shift off of using this `org.reflections` package since it's not maintained anymore, however for a quick fix we don't need to elevate the `commons-net` dependency to `compile`, we can simply add it as a `test` dependency here: https://github.com/Alfresco/alfresco-enterprise-repo/blob/a86d09dfeafe713c7caafd07c06c1fd8309d8117/repository/pom.xml#L125, like so:

```xml
        <dependency>
            <groupId>commons-net</groupId>
            <artifactId>commons-net</artifactId>
            <scope>test</scope>
        </dependency>
```

Please let me know if that works for you.

viepovsky:
Seems like only adding `commons-net` as `test` dependency to ent-repo solves the problem:
https://github.com/Alfresco/alfresco-enterprise-repo/pull/1515
https://github.com/Alfresco/alfresco-enterprise-repo/actions/runs/5680124842/job/15393683660

dsibilio (approved):
@pzhyland made me notice that you didn't really add the dependency, you just declared it in the `dependenciesManagement` section. So this solution should be fine as well 👍 Let's go ahead with this if you've tested it and it works.

PR #2098, opened 2023-07-28
ACS-5020 Investigate test failures

viepovsky:
Could not replicate error from: https://alfresco.atlassian.net/browse/ACS-5020 did not happen even once.

But there was other failure happening couple of times which after re-runing turned green. 
It is not connected with that test class which I have investigate but it is from the same job AGS Community Rest API Tests:
error

Apart from this I have only reduced PMD issues. And everything passes tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
   push:
     branches:
       - feature/**
-#      - fix/**
+      - fix/**
   workflow_call:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


dsibilio (approved):
LGTM.

Regarding the unrelated test that you noticed the intermittent failure of, if you have any CI/CD occurrences / logs please track this intermittent failure as a new bug under https://alfresco.atlassian.net/browse/ACS-5384.

PR #2120, opened 2023-08-07
There was no override of properties related to HttpClientConfig, whic…

mstrankowski:
…h meant that if a property wasn't present in global-properties files, it wouldn't be registered and couldn't be provided through command line. I've ammended that here


mstrankowski:
Only failures are instabilities, still need to propagate those

PR #2133, opened 2023-08-10
[PRODSEC-7274] bump bouncycastle version, add enforce rule for old bo…

kcichonczyk:
…uncycastle versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1110,10 +1110,15 @@
                                             <exclude>jakarta.xml.ws:jakarta.xml.ws-api:(, 3.0.1)</exclude>
                                             <exclude>jakarta.xml.soap:jakarta.xml.soap-api:(, 2.0.1)</exclude>
                                             <exclude>jakarta.jws:jakarta.jws-api:(, 3.0.0)</exclude>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1110,6 +1111,8 @@
                                             <exclude>jakarta.xml.ws:jakarta.xml.ws-api:(, 3.0.1)</exclude>
                                             <exclude>jakarta.xml.soap:jakarta.xml.soap-api:(, 2.0.1)</exclude>
                                             <exclude>jakarta.jws:jakarta.jws-api:(, 3.0.0)</exclude>
+                                            <exclude>org.bouncycastle:bcprov-jdk15on</exclude>
```
pzhyland, line 1114: What do you think about excluding **everything** from `org.bouncycastle` and allowing only selected components. It would allow us to fail te build if some dependency brings something like `org.bouncycastle:bcprov-jdk16`

kcichonczyk: Updated according to your suggestion.

pzhyland (approved):
Looks fine but please monitor how it integrates into the enterprise and packaging. Both these projects inherit the configuration from the community repo.

PR #2185, opened 2023-09-07
MNT-23950 Cleanup after review

pzhyland:


.github/workflows/ci.yml
```diff
@@ -396,11 +396,14 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Alfresco/alfresco-build-tools/.github/actions/get-build-info@v1.33.0
       - uses: Alfresco/alfresco-build-tools/.github/actions/setup-java-build@v1.33.0
+        with:
+          java-version: ${{ matrix.jre-version }}
```
dsibilio, line 400: We should define the jre-version as 17 for the other matrix objects otherwise this won't evaluate to anything meaningful

dsibilio (changes requested):

dsibilio (approved):

.github/workflows/ci.yml
pzhyland: Done

PR #2206, opened 2023-09-25
ACS-6057 Gytheio removal

pzhyland:


repository/src/main/java/org/alfresco/messaging/LoggingDeadLetterQueue.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class LoggingDeadLetterQueue
+{
+    private static final Logger LOG = LoggerFactory.getLogger(LoggingDeadLetterQueue.class);
+
+    public void onReceive(Object message)
+    {
+        if (LOG.isDebugEnabled() && message != null)
+        {
+            LOG.debug("Received:\n\n" + message + "\n\n");
```
dsibilio, line 39:
```suggestion
            LOG.debug("Received:\n\n{}\n\n", message);
```

dsibilio (approved):
LGTM. I assume most of this boils down to replicating the original Gytheio classes that are going to be replaced by our own so there's not much to comment on.

PR #2245, opened 2023-10-12
ACS-6070: Support for highlighting prefix and postfix

krdabrowski:
- added prefix and postfix to field model (TAS)

More info: https://alfresco.atlassian.net/browse/ACS-6070


tpage-alfresco (approved):
LGTM

PR #2248, opened 2023-10-13
[MNT-23896] run handleClientAuth only when this.enforce variable is true

code4uuuu:
https://alfresco.atlassian.net/browse/MNT-23896


suneet-gupta (approved):
LGTM

PR #2260, opened 2023-10-20
Bump Camel to 4.1.0, Netty to 4.1.100.Final

dsibilio:
As per title.


dsibilio:
Closing this as it shouldn't be urgent, so it's better to wait for a new version of Camel that will rely on Netty >= 4.1.100.Final instead of this one that is still using an older version and may not be 100% compatible.

PR #2287, opened 2023-10-31
[hackathon] Enable build cache

gionn:
Requires https://github.com/Alfresco/alfresco-build-tools/pull/340


jottley:
@gionn Do we need to keep this PR or can we close it?

PR #2289, opened 2023-11-02
MNT-23927 Handle JsonMappingException and JsonGenerationException

evasques:
- Handle JsonMappingException and JsonGenerationException by throwing an IOException so we do not risk sending a partial JSON response
- Added logging to alert when a null key is found


jottley (approved):

jamalkm (approved):
LGTM.

PR #2299, opened 2023-11-09
Switch to HttpClient5

kmagdziarz:


jottley:
@kmagdziarz As the JIRA is marked Done, do we need to keep the PR open?

PR #2311, opened 2023-11-15
[MNT-23960] Added new text to pdf pipelines

tiagosalvado10:


jottley:
@tiagosalvado10 Since the JIRA is marked Done do we need to keep this PR open?

tiagosalvado10:
@jottley I'm going to close this PR as it was created to provide an alternative approach.

PR #2340, opened 2023-11-24
MNT-24063 Fix GHA migration issues & disable failing AGS tests

dsibilio:
The GHA migration on this branch was done manually and is different from the usual, there were missing scripts that are required in enterprise-repo etc.

Hopefully this should address some of the issues and we don't find more down the line.


.github/workflows/ci.yml
```diff
@@ -68,23 +62,19 @@ jobs:
           java-version: "11"
```
aonych, line 62: Maybe it would be good to create a new env variable for java version?

aonych (approved):

pzhyland (approved):

.github/workflows/ci.yml
dsibilio: Normally yes, in this case not really IMO. It would be better to keep it as-is so it's as similar as possible to the what's on e.g.: `release/7.2.1` so that if the need to cherry-pick any more fixes this can be done without getting a headache + this branch will never move from java-version = 11.

PR #2379, opened 2023-12-19
ACS-6402: Add tests covering module compatibility check for new ACS versioning schema

mpichura:
As per [this comment](https://alfresco.atlassian.net/browse/ACS-6402?focusedCommentId=1223553) in ACS-6402 - I am adding tests which explicitly cover cases we may experience after applying new versioning schema in ACS on module compatibility checks.


krdabrowski (approved):
LGTM

Pawel-608 (approved):
LGTM, you may just want to split these test cases into different test methods (but it is readable now anyway)

PR #2426, opened 2024-02-01
ACS-6677 Enhance OIDC Configuration Flexibility

damianujma:


repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
```diff
@@ -146,10 +176,7 @@ public String getClientSecret()
 
     public String getIssuerUrl()
     {
-        return UriComponentsBuilder.fromUriString(getAuthServerUrl())
-                                   .pathSegment(REALMS, getRealm())
-                                   .build()
-                                   .toString();
+        return issuerUrl;
```
dsibilio, line 179: Can you clarify how this is not going to be a breaking change?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
```diff
@@ -103,9 +108,34 @@ public int getConnectionPoolSize()
         return connectionPoolSize;
     }
 
+    public void setIssuerUrl(String issuerUrl)
+    {
+        this.issuerUrl = issuerUrl;
+    }
```
dsibilio, line 114: This should be right below the `getIssuerUrl` method as they belong to the same theme.

repository/src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication-context.xml
```diff
@@ -89,6 +89,12 @@
    </bean>
 
    <bean name="identityServiceConfig" class="org.alfresco.repo.security.authentication.identityservice.IdentityServiceConfig">
+      <property name="issuerUrl">
+         <value>${identity-service.issuer-url:#{null}}</value>
```
dsibilio, line 93: I'm assuming we need to declare `null` as a default otherwise Spring might complain about not finding the property at all? If it's not the case and the default mapping is unnecessary of course let's get rid of it. But I'm assuming it was done for a reason so it should be good -- just double-checking.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
```diff
@@ -0,0 +1,263 @@
...
+    private static final String DISCOVERY_PATH_SEGEMENTS = "/.well-known/openid-configuration";
```
dsibilio, line 56:
```suggestion
    private static final String DISCOVERY_PATH_SEGMENTS = "/.well-known/openid-configuration";
```

Typo, let's also fix all references to this variable please.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
```diff
@@ -0,0 +1,263 @@
...
+            assertThat(clientRegistration.getProviderDetails().getIssuerUri()).isNotNull();
+            assertThat(requestEntityCaptor.getValue().getUrl().toASCIIString()).isEqualTo(
+                AUTH_SERVER + DISCOVERY_PATH_SEGEMENTS);
+        }
+    }
+
+    @Test
+    public void shouldThrowIdentityServiceExceptionIfIssuerIsNotValid()
+    {
+        config.setIssuerUrl("https://invalidissuer.com");
```
dsibilio, line 129: How does this test work? I thought the underlying RestTemplate was mocked but I don't see any special mocking for this usecase.
If it's actually doing calls to this `https://invalidissuer.com`, can we replace it instead with something else that's safer? E.g.: some completely gibberish string that is not compatible with the URL syntax or e.g.: `https://hyland.com` if it's absolutely necessary?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -351,59 +360,104 @@ else if (isDefined(config.getTruststore()))
...
         }
     }
 
-    private static class ClientRegistrationProvider
+    static class ClientRegistrationProvider
     {
         private final IdentityServiceConfig config;
 
-        private ClientRegistrationProvider(IdentityServiceConfig config)
+        public static final Set<String> SUPPORTED_SCOPES = Set.of("openid", "profile", "email", "offline_access");
```
dsibilio, line 376: Should have default/package private visibility as well if possible

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -351,59 +360,104 @@ else if (isDefined(config.getTruststore()))
...
+                .map(URI::toASCIIString)
+                .orElse(null);
 
             final String issuerUri = Optional.of(metadata)
-                    .map(OIDCProviderMetadata::getIssuer)
-                    .map(Issuer::getValue)
-                    .orElseGet(config::getIssuerUrl);
+                .map(OIDCProviderMetadata::getIssuer)
+                .map(Issuer::getValue)
+                .orElse(null);
```
dsibilio, line 450: I'm assuming this default-change is wanted as we only want to trust the Issuer Url recovered via the `OIDCProviderMetadata` calls?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -449,9 +511,19 @@ private Optional<OIDCProviderMetadata> extractMetadata(RestOperations rest, URI
 
         private Collection<URI> possibleMetadataURIs()
         {
-            return List.of(UriComponentsBuilder.fromUriString(config.getIssuerUrl())
-                                               .pathSegment(".well-known", "openid-configuration")
-                                               .build().toUri());
+            if (StringUtils.isBlank(config.getAuthServerUrl()) && StringUtils.isBlank(config.getIssuerUrl()))
+            {
+                throw new IdentityServiceException(
+                    "Failed to create ClientRegistration. The values of issuer url and auth server url cannot be empty.");
```
dsibilio, line 517:
```suggestion
                    "Failed to create ClientRegistration. The values of issuer url and auth server url cannot both be empty.");
```

I think it would make it clearer that only 1 of the 2 properties is necessary at least.

dsibilio (changes requested):
Good work! Left a few comments to double-check and address.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
```diff
@@ -103,9 +108,34 @@ public int getConnectionPoolSize()
...
+        {
+            return UriComponentsBuilder.fromUriString(authServerUrl)
+                .pathSegment(REALMS, getRealm())
+                .build()
+                .toString();
+        }
+        else
+        {
+            return authServerUrl;
+        }
```
dsibilio, line 138:
```suggestion

            return authServerUrl;
```

No need for the `else` clause. Just return `authServerUrl` immediately.

dsibilio (changes requested):

repository/src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication-context.xml
damianujma: I can see there is not difference for that syntax. It seems declaring `null`, it doesn't change anything, the behavior is the same. I got rid of it.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
damianujma: Done

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
damianujma: You are right. It's better to trust the Issuer URL from the configuration. Anyway, before that piece of code we validate the OIDC Discovery document endpoints because the Issuer URL is required.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
damianujma: Done.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
damianujma: Done. I replaced the whole logic with that Optional flow. There was a need to change the last line.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
damianujma: This line mocks the RestTemplate to return always the 200 OK status code with an empty body.  https://github.com/Alfresco/alfresco-community-repo/pull/2426/files#diff-f7fe0229c772ae958a59d19f8011a277d6ee5c62f511558aced4a975476514c0R76
This test also mocks the static method - OIDCProviderMetadata.parse to parse a prepared openid discovery document for tests.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
damianujma: Done.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
damianujma: Before the changes the Issuer URL was auto-generated - now we can configure the Issuer URL. That logic related to a realm has been moved to the `getAuthServerUrl()` method. Now we use authServerUrl to create the OIDC Discovery Endpoint (See: https://github.com/Alfresco/alfresco-community-repo/pull/2426/files/bf3dfc382cbd68088ee76627db8260bd13798315#diff-fdc9f584f86648fc72e778dca91ae854f9340b996d7837e3b441578c0232b8e4R520)
The only case where a compatibility would not be achived has been fixed -> https://github.com/Alfresco/alfresco-community-repo/pull/2426/commits/ba98794eda5280d7b90ac091bb30f75dfdf8df67
Anyway, that `orElseGet` part should not be executed because we validate the OIDC Discovery document endpoints and the Issuer URL is required.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -351,59 +360,106 @@ else if (isDefined(config.getTruststore()))
...
+            }
+
+            return metadata;
+        }
+
+        private void validateOIDCEndpoint(URI value, String endpointName)
+        {
+            if (value == null || value.toASCIIString().isBlank())
+            {
+                throw new IdentityServiceException("The `%s` Endpoint cannot be empty.".formatted(endpointName));
```
jamalkm, line 436: Please update the exception message to more clearly indicate that the value retrieved from the discovery endpoint document is invalid.

jamalkm (changes requested):
Great work! I just left a couple of minor comments that need to be addressed.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
dsibilio: If it's all mocked anyways then let's please use an invalid syntax for the domain name such as `https://invalid-issuer` so that there is 0% chance of hitting a real domain in case something in the implementation changes and the mocks are not updated appropriately.

damianujma: Done.

repository/src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication-context.xml
damianujma: I double-checked that. There is a need to declare the default value `null` because not delaring that syntax causes returning string from the value declaration (e.g. `identity-service.issuer-url`) if the property is not set.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -527,21 +604,30 @@ private void reconfigureJWKSCache(ConfigurableJWTProcessor<SecurityContext> jwtP
...
+            List<OAuth2TokenValidator<Jwt>> validators = new ArrayList<>();
+            validators.add(new JwtTimestampValidator(Duration.of(0, ChronoUnit.MILLIS)));
+            validators.add(new JwtIssuerValidator(providerDetails.getIssuerUri()));
+            if (!config.isClientIdValidationDisabled())
+            {
+                new JwtClaimValidator<String>("azp", config.getResource()::equals);
+            }
+            if (StringUtils.isNotBlank(config.getAudience()))
+            {
+                new JwtClaimValidator<String>(JwtClaimNames.AUD, config.getAudience()::equals);
```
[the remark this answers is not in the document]
damianujma: [Added JwtAudienceValidator](https://github.com/Alfresco/alfresco-community-repo/pull/2426/commits/be8691803d17e0ffdaf8581eb432088a9b379b44).

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
damianujma: Done.

jamalkm (approved):
Great job!

dsibilio (approved):
Good work 👍

aonych (approved):

PR #2441, opened 2024-02-08
Map includeChildAssociations request parameter

AFaust:
This PR provides a very minor/trivial change that will have ACS actually use the parameter `includeChildAssociations` as provided by Search Services since at least version 1.4.2 (https://github.com/Alfresco/SearchServices/commit/943fafcca383a995db221805fed2cd9036ec66e2). This will significantly reduce the overhead when indexing metadata changes on any node that contains large amount of child nodes as combined with the `includeChildIds` flag which is already handled and provided as `false` as well, this change results in the elimination of the extremely expensive `getChildAssocs` call in the `SOLRTrackingComponentImpl` `getNodesMetadata` method.


tpage-alfresco:
LGTM. Thanks @AFaust !

PR #2445, opened 2024-02-09
ACS-6309 Integration with Report Portal

wojtekswieton:


amps/ags/rm-community/rm-community-repo/pom.xml
```diff
@@ -84,6 +84,12 @@
          <artifactId>junit</artifactId>
          <scope>test</scope>
       </dependency>
+       <dependency>
+           <groupId>com.epam.reportportal</groupId>
+           <artifactId>agent-java-testng</artifactId>
+           <version>5.4.0</version>
```
damianujma, line 90: Can't we set the `agent-java-testng` and `okhttp` versions as version variables? This applies to all declarations in the code.

damianujma (changes requested):

dsibilio (changes requested):
Please let's check if we can put the `agent-java-testng` version in a common place along with the `okhttp` version override so we don't have to repeat these declarations that many times.

Also `okhttp` version override is missing the `test` scope.

damianujma (approved):

.github/workflows/ci.yml
```diff
@@ -517,16 +895,50 @@ jobs:
...
+        id: rp-summarize
+        with:
+          tests-outcome: ${{ steps.tests.outcome }}
+          rp-launch-key: ${{ steps.rp-prepare.outputs.key }}
+          rp-project: ${{ env.RP_PROJECT }}
+          rp-token: ${{ secrets.REPORT_PORTAL_TOKEN }}
+      - name: "Exit on failure"
+        if: steps.tests.outcome != 'success'
+        run: |
+          echo "::error title=
```
dsibilio, line 941: Incomplete piece of code?

dsibilio (changes requested):

dsibilio:
I've noticed that some reports from tests that passed are actually marked as failed on ReportPortal due to "Surefire interrupted", it would be nice to investigate whether we can figure out and address the root cause for this so that we don't have red reports on ReportPortal for green test suites.

Otherwise even though quite verbose, it looks good, great work.

.github/workflows/ci.yml
```diff
@@ -199,8 +252,44 @@ jobs:
         run: echo "TRANSFORMERS_TAG=$(mvn help:evaluate -Dexpression=dependency.alfresco-transform-core.version -q -DforceStdout)" >> $GITHUB_ENV
       - name: "Set up the environment"
         run: docker-compose -f ./scripts/ci/docker-compose/docker-compose.yaml --profile ${{ matrix.compose-profile }} up -d
+      - name: "Prepare Report Portal"
+        uses: Alfresco/alfresco-build-tools/.github/actions/reportportal-prepare@v5.1.0
+        id: rp-prepare
+        with:
+          rp-launch-prefix: ${{ env.RP_LAUNCH_PREFIX }} - ${{ matrix.testSuite }}
+          rp-token: ${{ secrets.REPORT_PORTAL_TOKEN }}
```
dsibilio, line 260: Please double-check that if this secret is not available as it happens for DependaBot PRs then the build will still succeed and run all necessary tests. you can easily test this by temporarily changing this line to `rp-token: "{{ secrets.SECRET_THAT_DOES_NOT_EXIST }}`

dsibilio (changes requested):
Please also double-check that if ``

.github/workflows/ci.yml
wojtekswieton: Checked, all jobs still works correctly: https://github.com/Alfresco/alfresco-community-repo/actions/runs/8457827488

core/src/test/java/AllCoreUnitTestSuite.java
```diff
@@ -0,0 +1,59 @@
...
+        org.alfresco.util.EqualsHelperTest.class,
+        org.alfresco.util.GuidTest.class,
+        org.alfresco.util.ISO8601DateFormatTest.class,
+        org.alfresco.util.LogAdapterTest.class,
+        org.alfresco.util.LogTeeTest.class,
+        org.alfresco.util.PathMapperTest.class,
+        org.alfresco.util.TempFileProviderTest.class,
+        org.alfresco.util.VersionNumberTest.class
+})
+public class AllCoreUnitTestSuite {
```
dsibilio, line 58: I noticed there are several new TestSuites added as part of this PR.

For each TestSuite can you explain:
- how they were generated (e.g.: reverse-engineered from GHA logs/Surefire reports)
- when we'll need to update them (e.g.: update AllCoreUnitTestSuite if you need a new unit test to be executed as part of https://github.com/Alfresco/alfresco-community-repo/pull/2445/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR159-R161)
- since these tests already existed and they were just grouped by a TestSuite, how come we didn't have to modify any existing TestSuite e.g.: removing the tests that have now been moved to this new TestSuite? Were all the tests for each of these test suites never part of any other test suite before?

Thanks 🙏

data-model/src/test/java/org/alfresco/AllDataModelUnitTestSuite.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.junit.experimental.categories.Categories;
+import org.junit.runner.RunWith;
+import org.junit.runners.Suite;
+
+@RunWith(Categories.class)
+@Suite.SuiteClasses({
+        org.alfresco.opencmis.dictionary.CMISAbstractDictionaryServiceTest.class,
+        org.alfresco.repo.content.encoding.CharsetFinderTest.class,
+        //org.alfresco.repo.content.DataModelContentTestSuite.class,
```
dsibilio, line 36: Why is this commented out?

.github/workflows/ci.yml
```diff
@@ -142,24 +145,74 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - testName: Core
+            testModule: core
+            testAttributes: "-Dtest=AllCoreUnitTestSuite"
+          - testName: Data-Model
+            testModule: data-model
+            testAttributes: "-DskipCoreTests=true -Dtest=AllDataModelUnitTestSuite"
```
dsibilio, line 164: Out of curiosity, why does this need the `-DskipCoreTests=true` while other elements in the matrix don't?

dsibilio (commented):
Thanks for cleaning this up, added a couple of questions!

damianujma (approved):
LGTM

core/src/test/java/AllCoreUnitTestSuite.java
dsibilio: I still need answers to these questions @wojtekswieton

wojtekswieton: >-how they were generated (e.g.: reverse-engineered from GHA logs/Surefire reports)

I created 3 new test suites.
AllCoreUnitTestSuite and AllDataModelUnitTestSuite were necessary after split 'Core, Data-Model, Repository - AllUnitTestsSuite - Build and test' to 4 jobs.
Before that tests in that job were run by mixing 2 different ways:
- name: "Run tests" run: | mvn -B test -pl core,data-model -am -DfailIfNoTests=false mvn -B test -pl "repository,mmt" -am "-Dtest=AllUnitTestsSuite,AllMmtUnitTestSuite" -DfailIfNoTests=false
Repository tests and mmt - by directly calling test suite. Core, data-model by using surefire test including mechanism in core and data-model modules. So I created a test suites for them - and add all tests located inside those modules and also manually checked on GHA logs if all of them are added to test suites. Right now all tests in those 4 modules are runs by the same way.

Third new test suite: MessagingUnitTestSuite just grouped together 2 test classes that were before runs separately which generates 2 reports.
run: mvn -B test -pl repository -am -Dtest=CamelRoutesTest,CamelComponentsTest -DfailIfNoTests=false 

Most details about those new test suites I describe in a confluence page: https://hyland.atlassian.net/wiki/spaces/TECH/pages/1336050823/alfresco-community-repo+integration+with+Report+Portal

>-when we'll need to update them (e.g.: update AllCoreUnitTestSuite if you need a new unit test to be executed as part of https://github.com/Alfresco/alfresco-community-repo/pull/2445/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR159-R161)

When we need to add a new test cases or update just need to check if there are added to correct test suite.

>-since these tests already existed and they were just grouped by a TestSuite, how come we didn't have to modify any existing TestSuite e.g.: removing the tests that have now been moved to this new TestSuite? Were all the tests for each of these test suites never part of any other test suite before?

As I wrote above - those tests in new test suites weren't in any test suite before. They were runs by surefire mechanism or were directly indicated.

aonych (approved):

PR #2446, opened 2024-02-13
[ACS-5506] Fix string to node ref cast exception

MichalKinas:
Added a fix for `String` to `NodeRef` cast exception when `returnedObject` contains pair of strings.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/security/permissions/impl/acegi/ACLEntryAfterInvocationProvider.java
+++ b/repository/src/main/java/org/alfresco/repo/security/permissions/impl/acegi/ACLEntryAfterInvocationProvider.java
@@ -269,7 +269,18 @@ public class ACLEntryAfterInvocationProvider implements AfterInvocationProvider,
             }
             else if (Pair.class.isAssignableFrom(returnedObject.getClass()))
             {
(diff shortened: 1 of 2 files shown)


repository/src/main/java/org/alfresco/repo/security/permissions/impl/acegi/ACLEntryAfterInvocationProvider.java
```diff
@@ -269,7 +269,18 @@ else if (NodeRef.class.isAssignableFrom(returnedObject.getClass()))
...
+                {
+                    return decide(authentication, object, config, pair);
+                } else
+                {
+                    if (log.isDebugEnabled())
+                    {
+                        log.debug("Uncontrolled object - access allowed for " + object.getClass().getName());
+                    }
+                    return returnedObject;
+                }
```
tpage-alfresco, line 283: If we're adding extra logic around this then I think it should go into the `decide` method.

MichalKinas: Sounds good, done

tpage-alfresco (approved):
LGTM

PR #2460, opened 2024-02-20
ACS-6113: Fix secondary parents to never be set to null

krdabrowski:
https://alfresco.atlassian.net/browse/ACS-6113


tpage-alfresco (approved):
LGTM!

PR #2489, opened 2024-03-06
ACS-6931 Temporarily turn off SAST Scan

mikolajbrzezinski:
The Veracode Pipeline SAST Scan started timing out after an hour and no solution was found on our side. This is meant to be temporary and to be restored once we resolve this issue with the help of Veracode Support.

https://alfresco.atlassian.net/browse/ACS-6931

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
       - name: "Clean Maven cache"
         run: bash ./scripts/ci/cleanup_cache.sh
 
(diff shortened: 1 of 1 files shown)


dsibilio (changes requested):
Please let's track a new task under the Maintenance bucket for the current train and use that as a reference instead of the original ticket id.

Then let's make sure to update the PR title + use the new ticket id during squash & merge.

dsibilio (approved):

PR #2504, opened 2024-03-12
MNT-24270 Disable Faceting when we have unknown nodes in the search results

evasques:
Unknown nodes can appear in the search results if the nodes where deleted or permissions changed while the index is not yet up to date or if we disabled permission checks on the search engine. 

In this case the total items will not match the number of entries that are returned and the facet hits will be wrong. Changes made were:
 - In v1, correct the total items and don't return facets, facet queries, facet fields, pivots, etc when we detect unknown nodes in the results
 - In alfresco-share-services, correct the total items based in the failed nodes count and don't send facets
 - Added test

### DIFF
--- a/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/search/search.lib.js
+++ b/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/search/search.lib.js
@@ -674,6 +674,11 @@ function processResults(nodes, maxPageResults, startIndex, rootNode, meta)
       }
    }
 
+   if (failed != 0 && logger.isWarnLoggingEnabled() == true)
+   {
+      logger.warn("Faceting disabled as hits are innacurate due to unknown nodeRefs returned");
+   }
+
    if (logger.isLoggingEnabled())
       logger.log("Filtered resultset to length: " + results.length + ". Discarded item count: " + failed);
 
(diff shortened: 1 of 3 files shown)


jottley (approved):

tpage-alfresco:
> In v1, correct the total items and don't return facets, facet queries, facet fields, pivots, etc when we detect unknown nodes in the results
This sounds concerning. Isn't it better to return slightly wrong results than no results?

evasques:
@jottley was who discussed the issue with the customer and the incorrect facet results were confusing the users much more.

IMO from a security perspective, by showing the possibly real results (when we have permissions disabled on SOLR for example) we are disclosing that we have X number of documents with that term (and because its facets we are actually exposing dates, creators, modifiers, etc as we show the hits on each ) despite the user only having access to a subset of those. Its safer to not return that info than leak that we have a document that matches your search but you don't have permissions to see it.

jottley:
@tpage-alfresco Like Eva indicated, the customer is having issues with the fact that a single document could be returned and they are seeing facets/counts from across the repo indicating the total documents are much higher.  They also don't like seeing facets that don't even apply to the result set. For example a single document returned but a facet count of 4 but nothing returned when the navigate to the facet.  We discussed with them a customization/configuration to remove the facets from being seen but they still want/rely on the facets.  This appears to be an acceptable approach.

jottley:
As the related JIRA has been closed and the decision to no go forward with this approach, I am going to close this PR

PR #2593, opened 2024-04-16
ACS-7598 Backward Compatible ATS/AIS

pzhyland:


SaraAspery (approved):

repository/pom.xml
```diff
@@ -123,6 +123,10 @@
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path</artifactId>
+        </dependency>
```
mstrankowski, line 129: Don't need the json-smart dependency here?

mstrankowski: Or accessors-smart for that matter, judging by the jira comment linked in secure-the-security?

pzhyland: We need three of them ;) It's the same approach we have for the ACS 7.4.2 & 23.X. See https://hyland.atlassian.net/browse/ACS-7522?focusedCommentId=1231805

PR #2656, opened 2024-05-17
ACS-7587 Implement v1 Bulk API to add items to a hold

damianujma:
Added the following properties:

* `rm.hold.bulk.threadCount` (default: 2) - The number of worker threads.
* `rm.hold.bulk.maxItems` (default: 1000) - The maximum number of total items to process in a single bulk operation.
* `rm.hold.bulk.logging.interval` (default: 1000) - The number of entries to process before reporting progress.
* `rm.hold.bulk.batchSize` (default: 100) - The number of entries to be fetched from the Search Service in the [BatchProcessWorkProvider.getNextWork()](https://github.com/Alfresco/alfresco-community-repo/blob/d907461972e2d5bb97f17006583232559b2fdb97/repository/src/main/java/org/alfresco/repo/batch/BatchProcessWorkProvider.java#L60) method.
* `rm.hold.bulk.itemsPerTransaction` (default: 1) - The number of entries we process at a time in a transaction. If the umber is 1, the advantage of this is that when a transaction fails, correctly added entries are not rollbacked. If the number of the items per transaction is e.g. 100 then if one entry fails in a transaction, 99 entries are rollbacked and the [BatchMonitor](https://github.com/Alfresco/alfresco-community-repo/blob/d907461972e2d5bb97f17006583232559b2fdb97/repository/src/main/java/org/alfresco/repo/batch/BatchMonitor.java#L35) returns only 1 error and 99 successfully processed entries. **It's worth thinking.**

Note: _Support for the clustered environment will be added in the Enterprise Repo._


amps/ags/rm-community/rm-community-repo/config/alfresco/module/org_alfresco_module_rm/alfresco-global.properties
```diff
@@ -139,3 +139,16 @@ content.metadata.async.extract.6.enabled=false
...
 rm.recordSearch.maxItems=500
+
+#
+# Hold bulk
+#
+rm.hold.bulk.threadCount=2
+rm.hold.bulk.maxItems=1000
+rm.hold.bulk.batchSize=100
+rm.hold.bulk.logging.interval.ms=1000
+rm.hold.bulk.itemsPerTransaction=1
```
tpage-alfresco, line 150: You wrote some great documentation for these in the PR description. Please include them here so that customers/partners (and we) have an easy place to see what each property does.

amps/ags/rm-automation/rm-automation-community-rest-api/src/main/java/org/alfresco/rest/rm/community/model/hold/HoldBulkStatus.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public enum Status
+    {
+        PENDING,
+        IN_PROGRESS,
+        DONE
+    }
+
+    @Override
+    public boolean equals(Object o)
```
tpage-alfresco, line 82: You already defined `equals` and `hashCode` with `@Data`

amps/ags/rm-automation/rm-automation-community-rest-api/src/main/java/org/alfresco/rest/rm/community/model/hold/HoldBulkOperationEntry.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author Damian Ujma
+ */
+@Builder
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class HoldBulkOperationEntry
+{
+    @JsonProperty
```
tpage-alfresco, line 47: Are these `@JsonProperty` annotations needed?

amps/ags/rm-automation/rm-automation-community-rest-api/src/main/java/org/alfresco/rest/rm/community/requests/gscore/api/HoldsAPI.java
```diff
@@ -287,4 +291,113 @@ public void deleteHoldChild(String holdId, String holdChildId)
...
     }
+
+    /**
+     * Starts a bulk process for a hold.
+     *
+     * @param holdBulkOperation The bulk operation details
+     * @param hold The identifier of a hold
+     * @param parameters The URL parameters to add
+     * @return The {@link HoldBulkOperationEntry} for the started bulk process
+     * @throws RuntimeException for the following cases:
```
tpage-alfresco, line 302: Generally we use (or extend) AlfrescoRuntimeException. I'm not sure whether that's what you've done or not here.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
```diff
@@ -0,0 +1,519 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * API tests for adding items to holds via the bulk process
+ */
+public class AddToHoldsBulkV1Tests extends BaseRMRestTest
+{
+    private static final String ACCESS_DENIED_ERROR_MESSAGE = "Access Denied.  You do not have the appropriate " +
+        "permissions to perform this operation.";
+    private static final int NUMBER_OF_FILES = 30;
```
tpage-alfresco, line 83: This sounds like it might take quite a while to run? Can we achieve the same level of testing with e.g. 3?

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
```diff
@@ -0,0 +1,519 @@
...
+    private RoleService roleService;
+    @Autowired
+    private ContentActions contentActions;
+
+    @BeforeClass(alwaysRun = true)
+    public void preconditionForAddContentToHold()
+    {
+        STEP("Create a hold.");
+        hold = getRestAPIFactory().getFilePlansAPI(getAdminUser()).createHold(
+            Hold.builder().name("HOLD" + generateTestPrefix(AddToHoldsV1Tests.class)).description(HOLD_DESCRIPTION)
```
tpage-alfresco, line 102: In theory the idea was to create a constant:
```
private static final String PREFIX = generateTestPrefix(AddToHoldsV1Tests.class);
```
and then use it as a prefix:
```
.name(PREFIX + "HOLD")
```
but it doesn't really matter if you just want a random name.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
```diff
@@ -0,0 +1,519 @@
...
+
+        rootFolder = dataContent.usingAdmin().usingSite(testSite).createFolder();
+        FolderModel folder1 = dataContent.usingAdmin().usingResource(rootFolder).createFolder();
+        FolderModel folder2 = dataContent.usingAdmin().usingResource(folder1).createFolder();
+
+        // Add files to subfolders in the site
+        for (int i = 0; i < NUMBER_OF_FILES; i++)
+        {
+            FileModel documentHeld = dataContent.usingAdmin()
+                .usingResource(new Random().nextBoolean() ? folder1 : folder2)
```
tpage-alfresco, line 117: It would be better to add them alternately, otherwise we might end up with test failures based on the random distribution of files.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
```diff
@@ -0,0 +1,519 @@
...
+        STEP("Check the bulk status.");
+        HoldBulkStatus holdBulkStatus = getRestAPIFactory().getHoldsAPI(userAddHoldPermission)
+            .getBulkStatus(hold.getId(), bulkOperationEntry.getBulkStatusId());
+        assertBulkProcessStatus(holdBulkStatus, NUMBER_OF_FILES, 0, null);
+
+        STEP("Check the bulk statuses.");
+        HoldBulkStatusCollection holdBulkStatusCollection = getRestAPIFactory().getHoldsAPI(userAddHoldPermission)
+            .getBulkStatuses(hold.getId());
+        assertEquals(1, holdBulkStatusCollection.getEntries().size());
+        assertEquals(holdBulkStatus, holdBulkStatusCollection.getEntries().get(0).getEntry());
```
tpage-alfresco, line 175: Can't you test equality of the collection in a single call?

amps/ags/rm-community/rm-community-repo/config/alfresco/module/org_alfresco_module_rm/alfresco-global.properties
damianujma: Done.

amps/ags/rm-automation/rm-automation-community-rest-api/src/main/java/org/alfresco/rest/rm/community/model/hold/HoldBulkStatus.java
damianujma: Removed.

amps/ags/rm-automation/rm-automation-community-rest-api/src/main/java/org/alfresco/rest/rm/community/model/hold/HoldBulkOperationEntry.java
damianujma: Not really. Removed.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
damianujma: Changed to 5 to achieve the same level of testing.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
damianujma: I left it as it is because in this test class we need 3 unique hold names.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
```diff
@@ -0,0 +1,519 @@
...
+            .until(() -> getRestAPIFactory().getSearchAPI(null).search(searchRequest).getPagination()
+                .getTotalItems() == NUMBER_OF_FILES);
+
+        holdBulkOperation = HoldBulkOperation.builder()
+            .query(queryReq)
+            .op(HoldBulkOperationType.ADD).build();
+    }
+
+    /**
+     * Given a user with the add to hold capability and hold filling permission
```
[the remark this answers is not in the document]
damianujma: Done.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
damianujma: Changed to have a single call.

amps/ags/rm-automation/rm-automation-community-rest-api/src/test/java/org/alfresco/rest/rm/community/hold/AddToHoldsBulkV1Tests.java
tpage-alfresco: I get the idea, but IMO it's not good. Instead we should pick fixed scenarios to test which cover the cases we think are interesting.

I've seen really hard to investigate issues in Search Services tests when we used the underlying Solr test framework. It turned out that the failures were due to this function hidden away in the framework:
https://github.com/fnp/pylucene/blob/master/lucene-java-3.5.0/lucene/src/test-framework/java/org/apache/lucene/util/LuceneTestCase.java#L734
It would have been much better to simply test both cases every run, rather than having the failure happen one in ten(?) runs.

damianujma: Thank you for the explanation. I changed it to add test files alternately.

tpage-alfresco (approved):
LGTM

amps/ags/rm-community/rm-community-repo/config/alfresco/module/org_alfresco_module_rm/alfresco-global.properties
```diff
@@ -139,3 +139,21 @@ content.metadata.async.extract.6.enabled=false
...
+# Hold bulk
+#
+# The number of worker threads.
+rm.hold.bulk.threadCount=2
+# The maximum number of total items to process in a single bulk operation.
+rm.hold.bulk.maxItems=1000
+# The number of entries to be fetched from the Search Service as a next set of work object to process.
+rm.hold.bulk.batchSize=100
+# The number of entries to process before reporting progress.
+rm.hold.bulk.logging.interval=1000
```
dsibilio, line 153: Isn't this a bit high given that the default `maxItems` for a bulk operation is `1000` as well? I assume this means we would get progress information only after the whole operation has completed, which seems a bit extreme?

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkMonitor.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param nodeRef   the node reference
+     * @param processId the process id
+     */
+    void registerProcess(NodeRef nodeRef, String processId);
+
+    /**
+     * Get the bulk status
+     *
+     * @param bulkStatusId the process name
```
dsibilio, line 54: Is the `bulkStatusId` really the `process name`?

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkOperation.java
```diff
@@ -0,0 +1,44 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.module.org_alfresco_module_rm.bulk;
+
+import org.alfresco.rest.api.search.model.Query;
+
+/**
+ * An immutable POJO to represent a bulk operation
```
dsibilio, line 32:
```suggestion
 * An immutable POJO that represents a bulk operation
```

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/DefaultHoldBulkMonitor.java
```diff
@@ -0,0 +1,109 @@
...
+    public HoldBulkStatus getBulkStatus(String bulkStatusId)
+    {
+        return holdProgressCache.get(bulkStatusId);
+    }
+
+    @Override
+    public List<HoldBulkStatus> getBatchStatusesForHold(String holdId)
+    {
+        return Optional.ofNullable(holdProcessRegistry.get(holdId))
+            .map(list -> list.stream()
```
dsibilio, line 75: It would be good to avoid generic terms like `list` to make the code easier to read. What is this `list` made of? Is it a list of `bulkProcessDetails`? Then it should probably be called just `bulkProcessDetails` instead.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkMonitor.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface HoldBulkMonitor extends BulkMonitor<HoldBulkStatus>
+{
+    /**
+     * Get the batch statuses for a hold
+     *
+     * @param holdId the hold id
+     * @return the batch statuses
+     */
+    List<HoldBulkStatus> getBatchStatusesForHold(String holdId);
```
dsibilio, line 45: We always used the term `bulk` until now, is there a particular reason why we're referring to a `batch` here instead? If not, I'd suggest adhering to the semantic convention we just made up to avoid confusion.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkStatusUpdater.java
```diff
@@ -0,0 +1,69 @@
...
+            task.run();
+        }
+    }
+
+    @Override
+    public void publishEvent(Object event)
+    {
+        if (event instanceof BatchMonitorEvent batchMonitorEvent)
+        {
+            batchMonitor = batchMonitorEvent.getBatchMonitor();
```
dsibilio, line 66: Assuming `BatchMonitorEvents` can be published by multiple threads at the same time, this may yield unexpected results. I'd recommend double-checking this possibility and eventually:
- using thread-safe abstractions to contain the `batchMonitor` variable such as `AtomicReference`
- implementing a custom logic on top to only replace the `AtomicReference` pointer if the new event is actually found to be more up-to-date than the previous

in case it could be a real issue that events come from multiple threads and with different degrees of progress, that may not be necessarily ordered by default.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkStatusUpdater.java
```diff
@@ -0,0 +1,69 @@
...
+                batchMonitor.getTotalErrorsLong(), batchMonitor.getTotalResultsLong(),
+                batchMonitor.getLastError()));
+    }
+
+    @Override
+    public void update()
+    {
+        if (task != null)
+        {
+            task.run();
```
dsibilio, line 57: What if we run this task and the `batchMonitor` is still `null`?

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkBaseService.java
```diff
@@ -0,0 +1,249 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.InitializingBean;
+
+/**
+ * A base class for executing bulk operations on nodes based on search query results
+ */
+public abstract class BulkBaseService<T> implements InitializingBean
+{
+    private static final Log LOG = LogFactory.getLog(BulkBaseService.class);
```
dsibilio, line 52: Is there a reason why we're not using SLF4J here instead? (`private static final Logger LOGGER = LoggerFactory.GetLogger(...)`)

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkBaseService.java
```diff
@@ -0,0 +1,249 @@
...
+            {
+                if (LOG.isDebugEnabled())
+                {
+                    LOG.debug("Started processing batch with name: " + batchProcessor.getProcessName());
+                }
+                batchProcessor.processLong(batchProcessWorker, true);
+                if (LOG.isDebugEnabled())
+                {
+                    LOG.debug("Processing batch with name: " + batchProcessor.getProcessName() + " completed");
+                }
```
dsibilio, line 130: If we switch to SLF4J these should use appropriate logging format syntax (e.g.: `"Started processing batch with name: {}, batchProcessor.getProcessName()"`)

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkBaseService.java
```diff
@@ -0,0 +1,249 @@
...
+     * @param processId  process id
+     * @param totalItems total items
+     * @return bulk status
+     */
+    protected abstract T getInitBulkStatus(String processId, long totalItems);
+
+    /**
+     * Get bulk status updater
+     *
+     * @return task container
```
dsibilio, line 158: ?

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkServiceImpl.java
```diff
@@ -0,0 +1,242 @@
...
+    @Override
+    protected BatchProcessWorker<NodeRef> getWorkerProvider(NodeRef nodeRef, BulkOperation bulkOperation)
+    {
+        if (ADD.name().equals(bulkOperation.operationType()))
+        {
+            return new AddToHoldWorkerBatch(nodeRef);
+        }
+        throw new InvalidArgumentException(
+            "Unsupported action type when starting the bulk process: " + bulkOperation.operationType());
+    }
```
dsibilio, line 102: Enums are primary targets to use `switch` operators instead of `if-else` blocks.

aonych (approved):

mikolajbrzezinski (approved):

amps/ags/rm-community/rm-community-repo/config/alfresco/module/org_alfresco_module_rm/alfresco-global.properties
damianujma: I agree, I think `100` should be enough.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkMonitor.java
damianujma: This is leftover. Changed.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/DefaultHoldBulkMonitor.java
damianujma: Changed to `bulkProcessDetailsList`.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkMonitor.java
damianujma: This is leftover. Changed. Now we are referring to a `batch`.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkStatusUpdater.java
damianujma: This event is send once by one thread during a new batch processor instantiating. See: https://github.com/Alfresco/alfresco-community-repo/blob/master/repository/src/main/java/org/alfresco/repo/batch/BatchProcessor.java#L243
Moreover, we use synchronized methods to get information about the status from the `BatchMonitor` instance.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkStatusUpdater.java
damianujma: Good catch. Added the null-check for the `batchMonitor`.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkBaseService.java
damianujma: I wanted to provide the same logger for the batch processor. Sadly the batch processor requires `org.apache.commons.logging.Log` in the constructor.

See: https://github.com/Alfresco/alfresco-community-repo/blob/master/repository/src/main/java/org/alfresco/repo/batch/BatchProcessor.java#L36

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/BulkBaseService.java
damianujma: This is leftover. Changed.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/bulk/hold/HoldBulkServiceImpl.java
damianujma: Since we have only one case, IDE suggested to use the `if-else` block for now. I refactored it to use `switch`.

dsibilio (approved):

PR #2762, opened 2024-07-08
ACS-8352 Migrate to docker compose v2

aonych:
As per title.

### DIFF
--- a/amps/ags/ci/scripts/getLogs.sh
+++ b/amps/ags/ci/scripts/getLogs.sh
@@ -4,9 +4,9 @@ set -x
 # Display running containers
 docker ps
 
-alfrescoContainerId=$(docker ps -a | grep '-alfresco-' | awk '{print $1}')
-shareContainerId=$(docker ps -a | grep '-share-' | awk '{print $1}')
-solrContainerId=$(docker ps -a | grep '-search-' | awk '{print $1}')
+alfrescoContainerId=$(docker ps -a | grep '\-alfresco\-' | awk '{print $1}')
+shareContainerId=$(docker ps -a | grep '\-share\-' | awk '{print $1}')
+solrContainerId=$(docker ps -a | grep '\-search\-' | awk '{print $1}')
 
 docker logs $alfrescoContainerId > alfresco.log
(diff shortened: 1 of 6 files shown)


amps/ags/ci/scripts/getLogs.sh
```diff
@@ -4,9 +4,9 @@ set -x
 # Display running containers
 docker ps
 
-alfrescoContainerId=$(docker ps -a | grep '_alfresco_' | awk '{print $1}')
-shareContainerId=$(docker ps -a | grep '_share_' | awk '{print $1}')
-solrContainerId=$(docker ps -a | grep '_search_' | awk '{print $1}')
+alfrescoContainerId=$(docker ps -a | grep '-alfresco-' | awk '{print $1}')
+shareContainerId=$(docker ps -a | grep '-share-' | awk '{print $1}')
+solrContainerId=$(docker ps -a | grep '-search-' | awk '{print $1}')
```
dsibilio, line 9: There are several instances in this PR of either `grep -something`, `grep '-something'`. I very much doubt these will work. `-` is a special character in bash, and when it's the first character of a string it can be misinterpreted by the shell as a command argument causing unpredictable behaviour. There are several ways to escape this, e.g.: `grep '\-alfresco-'` or `grep -- -alfresco-`.

I just noticed there is also a commit that addressed this very same issue on master and that should be backported along with the Docker Compose migration: https://github.com/Alfresco/alfresco-community-repo/commit/84948e051a5b4a79686b631f430e32f79dc7c320#diff-fd3cc0697d0958a3a823fcb2018e241731b1e71ec772b7e7159943abe8f3527a

dsibilio (changes requested):

amps/ags/ci/scripts/getLogs.sh
aonych: I totally missed that. Thank you! Updated.

dsibilio (approved):
Please remember these changes need to also be backported to the latest HF branch for 7.4, 7.3 and 7.2.

Also, let's use cherry-pick instead of manually replicating the changes.

PR #2779, opened 2024-07-12
[MNT-24137][ags][tas] Audit Bug Fix for 500 Internal Server Error

SathishK-T:
https://hyland.atlassian.net/browse/MNT-24137


SathishK-T:
This Fix is not valid. Need to do more analysis on the issue.

PR #2787, opened 2024-07-16
[MNT-24310] Fix AGS Rest API returning non RM objects

suneet-gupta:
https://hyland.atlassian.net/browse/MNT-24310


code4uuuu (approved):
LGTM!

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/rm/rest/api/impl/SearchTypesFactory.java
```diff
@@ -92,7 +92,7 @@ public Set<QName> buildSearchTypesForUnfiledEndpoint(Parameters parameters, Set<
         boolean includeRecords = false;
```
purusothaman-mm, line 92: By default boolean value will be false. Should we need to assign values.

suneet-gupta: Agrees but this is existing code.

PR #2812, opened 2024-07-25
[ACS-8474] Add new bean for intercepting current audit entries.

kmagdziarz:


p4535992:
There is any problem with this PR ?

PR #2891, opened 2024-09-02
Feature/acs 6309 bump test ng

wojtekswieton:


pom.xml
```diff
@@ -112,6 +112,7 @@
         <dependency.jakarta-ee-json-impl.version>1.1.6</dependency.jakarta-ee-json-impl.version>
         <dependency.jakarta-json-path.version>2.9.0</dependency.jakarta-json-path.version>
         <dependency.json-smart.version>2.5.1</dependency.json-smart.version>
+        <dependency.org.testng.version>7.9.0</dependency.org.testng.version>
```
dsibilio, line 115: Let's refactor this to just be `dependency.testng.version` if possible. We don't usually specify the groupId segments in the version properties here.

dsibilio: Also, any reason why we're using `7.9.0` specifically instead of the latest `7.10.2`?

.github/workflows/ci.yml
```diff
@@ -174,7 +174,7 @@ jobs:
       - name: "Init"
         run: bash ./scripts/ci/init.sh
       - name: "Prepare Report Portal"
-        if: github.ref_name == 'master'
+        if: github.ref_name == 'feature/ACS-6309_Bump_TestNG'
```
dsibilio, line 177: Remember to revert once approved and ready to merge.

dsibilio (changes requested):

pom.xml
wojtekswieton: 7.9.0 was last version of testNG released before current version of report-portal testng-agent. Its not working with a newer version

packaging/tests/tas-email/pom.xml
```diff
@@ -36,11 +36,22 @@
...
         <dependency>
             <groupId>com.epam.reportportal</groupId>
             <artifactId>agent-java-testng</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <scope>test</scope>
+        </dependency>
```
dsibilio, line 54: Why is this needed only here? 🤔

wojtekswieton: Rest of the tas- modules already got those dependency. it's set in a global pom DependencyManagement, so we need those dependencies also in a child poms to actually use it

dsibilio (approved):

PR #2905, opened 2024-09-11
ACS-8754 Return 405 status code when calling /entity/id/relationship against a resource that only supports ReadById.

tpage-alfresco:


krdabrowski (approved):
LGTM

PR #2915, opened 2024-09-15
[MNT-23379] Do not remove site manager permissions from local set when inheritance is disabled

tiagosalvado10:


tiagosalvado10:
Going to close this PR. Changes will be addressed in https://github.com/Alfresco/alfresco-community-repo/pull/2914

PR #2926, opened 2024-09-19
[MNT-23264] Added fix for running job as system user

suneet-gupta:
https://hyland.atlassian.net/browse/MNT-23264

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
+++ b/repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
@@ -157,8 +157,8 @@ public class ScriptActionExecuter extends ActionExecuterAbstractBase
                 // the default scripting model provides access to well known objects and searching
                 // facilities - it also provides basic create/update/delete/copy/move services
                 Map<String, Object> model = this.serviceRegistry.getScriptService().buildDefaultModel(
-                            null,
-                        null,
+                        personRef,
+                        getCompanyHome(),
                         homeSpaceRef,
                         scriptRef,
                         actionedUponNodeRef,
--- a/repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
+++ b/repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
@@ -197,6 +197,7 @@ public class ScriptActionExecuter extends ActionExecuterAbstractBase
     /**
      * @see org.alfresco.repo.action.ParameterizedItemAbstractBase#addParameterDefinitions(java.util.List)
      */
+    @Override
     protected void addParameterDefinitions(List<ParameterDefinition> paramList)
     {
         if (scriptLocation == null)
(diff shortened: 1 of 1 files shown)


mohit-singh4 (approved):
LGTM!

repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
```diff
@@ -198,6 +197,7 @@ protected void executeImpl(Action action, NodeRef actionedUponNodeRef)
     /**
      * @see org.alfresco.repo.action.ParameterizedItemAbstractBase#addParameterDefinitions(java.util.List)
      */
+    @Override
```
purusothaman-mm, line 200: Is this @Override is required for the fix ?

suneet-gupta: Yes. This is not related to the fix in the ticket rather an already existing PMD issue which now appeared in the build

purusothaman-mm (approved):
LGTM!

repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
```diff
@@ -152,9 +152,8 @@ protected void executeImpl(Action action, NodeRef actionedUponNodeRef)
             {
                 // get the references we need to build the default scripting data-model
                 String userName = this.serviceRegistry.getAuthenticationService().getCurrentUserName();
-                NodeRef personRef = this.personService.getPerson(userName);
-                NodeRef homeSpaceRef = (NodeRef)nodeService.getProperty(personRef, ContentModel.PROP_HOMEFOLDER);
-                
+                NodeRef personRef = this.personService.getPersonOrNull(userName);
```
code4uuuu, line 155: can we add logger here in case of personRef is null?

suneet-gupta: We can but TBH I don't think it will add much value as we are not fetching or using any property from Person object further. Person object is not required for script execution performed during rule action.

code4uuuu: okay, then it would be fine

jottley, line 155: Please add comments to explain these two line changes.  We need something to the affect of when a script is executed as a system user there is no person to home space to provide to the scripting framework.

jottley (commented):
We should also have have the documentation (https://docs.alfresco.com/content-services/latest/develop/reference/repo-root-objects-ref/).  We need to update the references to the person and userHome objects that this are not available when a script is executed as the system user.

repository/src/main/java/org/alfresco/repo/action/executer/ScriptActionExecuter.java
suneet-gupta: Added comments above the two lines. Similar Changes haven been already in-place for SAP Connector.
https://github.com/Alfresco/sap-content-connector-repo/blob/36bcd4efeecba7b7a529af49f6706f22aa8851fa/src/main/java/com/alfresco/integrations/sap/repo/action/PatchedScriptActionExecuter.java

jottley (approved):

PR #2947, opened 2024-09-27
ACS-8871: Bump Spring from 6.0.19 to 6.1.13

krdabrowski:


remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
```diff
@@ -72,42 +74,46 @@ public class ContentGet extends StreamContent implements ServletContextAware
     private List<String> nonAttachContentTypes = Collections.emptyList();
 
     /**
-     * @param nonAttachContentTypes List<String>
+     * @param nonAttachContentTypes
+     *            List<String>
```
tpage-alfresco, line 78: Is Spotless doing this? I guess I've not seen it because we have very little Javadoc in the HxI Connector.

tpage-alfresco (approved):
LGTM

remote-api/src/main/java/org/alfresco/repo/web/scripts/content/ContentGet.java
krdabrowski: Yes, this is made by Spotless

pzhyland (approved):

PR #3032, opened 2024-11-12
Fix/apps 2894 node size details automation test

mohit-singh4:
https://hyland.atlassian.net/browse/APPS-2894

### DIFF
--- a/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/nodes/NodeSizeDetailsTests.java
+++ b/packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/nodes/NodeSizeDetailsTests.java
@@ -59,8 +59,8 @@ public class NodeSizeDetailsTests extends RestTest
                 .and().field("content.mimeType").is(FileType.TEXT_PLAIN.mimeType);
         fileSize = Utility.getResourceTestDataFile("sampleLargeContent.txt").length();
 
-        STEP("3. Wait for 20 seconds so that the content is indexed in Search Service.");
-        Thread.sleep(20000);
+        STEP("3. Wait for 30 seconds so that the content is indexed in Search Service.");
+        Thread.sleep(30000);
 
         RestSizeDetailsModel restSizeDetailsModel = restClient.authenticateUser(user1).withCoreAPI().usingNode(folder).executeSizeDetails();
         restClient.assertStatusCodeIs(HttpStatus.ACCEPTED);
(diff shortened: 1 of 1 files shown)


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/nodes/NodeSizeDetailsTests.java
```diff
@@ -20,17 +26,147 @@ public class NodeSizeDetailsTests extends RestTest
...
+        Utility.waitToLoopTime(30);
+
+        RestSizeDetailsModel restSizeDetailsModel = restClient.authenticateUser(user1).withCoreAPI().usingNode(folder).executeSizeDetails();
+        restClient.assertStatusCodeIs(HttpStatus.ACCEPTED);
+        restSizeDetailsModel.assertThat().field("jobId").isNotEmpty();
+
+        jobId = restSizeDetailsModel.getJobId();
+
+        STEP("4. Wait for 3 seconds for the processing to complete.");
+        Utility.waitToLoopTime(3);
```
pzhyland, line 72: Please avoid waiting a fixed time. These test might be run on environment with different resources. Sometimes 3 seconds might be enough sometimes it might be not. It will lead to flaky tests and will require additional maintenance in the future. Try to increase the wait time but during that time check the assertions multiple times. You can do it manually in the loop or you can use `org.awaitility.Awaitility` for instance.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/nodes/NodeSizeDetailsTests.java
```diff
@@ -20,17 +26,147 @@ public class NodeSizeDetailsTests extends RestTest
     private UserModel user1;
     private SiteModel siteModel;
     private FolderModel folder;
+    private String jobId;
 
     @BeforeClass(alwaysRun = true)
     public void dataPreparation()
     {
-        user1 = dataUser.createRandomTestUser("User-1");
+        user1 = dataUser.getAdminUser();
```
pzhyland, line 34: Does it mean that only `admin` users can calculate the folder size?

mohit-singh4: No, the Random Test User is also able to calculate the folder size. 
We have updated it to use RandomUser instead.

packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/nodes/NodeSizeDetailsTests.java
mohit-singh4: Acknowledged.
We have replaced the fixed time with **Awaitility.await().atMost(Duration.ofSeconds())** and configured it to be used when retrieving the size details.

pzhyland (approved):

kavitshah-gl (approved):
Approved

PR #3058, opened 2024-11-27
Fix/MNT-24571-Site Membership API

varapathijanakiram:
The Site Membership API fails when the last name is null. This fix addresses the issue to ensure that the API can handle cases where the last name field is missing or null without throwing an error.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/site/SiteMembership.java
+++ b/repository/src/main/java/org/alfresco/repo/site/SiteMembership.java
@@ -25,14 +25,14 @@
  */
 package org.alfresco.repo.site;
 
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+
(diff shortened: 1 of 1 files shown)


rrajoria (approved):

repository/src/main/java/org/alfresco/repo/site/SiteMembership.java
```diff
@@ -52,7 +53,7 @@ public class SiteMembership extends AbstractSiteMembership
     public SiteMembership(SiteInfo siteInfo, String id, String firstName, String lastName,
             String role)
     {
-        super(siteInfo,id, role);
+        super(siteInfo, id, role);
```
MohinishSah, line 56: Why we have space here?

MohinishSah (changes requested):
LGTM

MohinishSah (approved):

PR #3065, opened 2024-11-29
[Fix/MNT-24209] logging for download restapi

mohit-singh4:
https://hyland.atlassian.net/browse/MNT-24209


repository/src/main/java/org/alfresco/repo/download/ZipDownloadExporter.java
```diff
@@ -1,329 +1,344 @@
...
+        }
+    }
+
+    @Override
+    public void contentImpl(NodeRef nodeRef, QName property, InputStream content, ContentData contentData, int index)
+    {
+        // if the content stream to output is empty, then just return content descriptor as is
+        if (content == null)
+        {
+            log.info("Archiving content has been removed or modified for the specified NodeReference: " + nodeRef
```
mohit-singh4, line 185: There's a logger, which has been configured for which we are getting content data as null!

SathishK-T (approved):
LGTM

PR #3083, opened 2024-12-11
[ACS-9045] Bump Spring Framework to 6.2.0

kmagdziarz:


repository/src/main/resources/alfresco/public-services-security-context.xml
```diff
@@ -32,12 +32,12 @@
...
         </property>
     </bean>
-    
+
     <!-- ======================= -->
     <!-- Support for permissions -->
     <!-- ========================-->
 
-    <bean id='permissionService' class="org.springframework.transaction.interceptor.TransactionProxyFactoryBean">
+    <bean id="permissionService" class="org.springframework.transaction.interceptor.TransactionProxyFactoryBean" depends-on="transactionManager, permissionServiceImpl">
```
dsibilio, line 40: How come we had to specify this `depends-on` clause when the 2 beans are already included as refs to set in the bean's properties?

dsibilio (approved):
LGTM assuming tests pass. Left one comment about something that's unclear.

repository/src/main/java/org/alfresco/repo/transaction/RetryingTransactionInterceptor.java
```diff
@@ -64,7 +64,7 @@ public Object invoke(final MethodInvocation invocation) throws Throwable
             final TransactionAttribute txnAttr = getTransactionAttributeSource().getTransactionAttribute(
                     method, invocation.getThis().getClass());
 
-            final TransactionManager tm = determineTransactionManager(txnAttr);
+            final TransactionManager tm = determineTransactionManager(txnAttr, null);
```
kmagdziarz, line 67: The on attribute method was deprecated.

repository/src/main/resources/alfresco/public-services-security-context.xml
kmagdziarz, line 40: We need to force order of bean creation after changes in spring alghoritm.

kmagdziarz: After tens of hours of debugging circular dependencies and nulls inside `transactionManager`, trying to understand what exactly Spring has change in bean creation algorithm, we came up with this solution to force the correct order by adding `depends-on`. 
It looked like Spring was trying to create `permissionService` before the beans that are injected into it.

pzhyland (approved):

PR #3084, opened 2024-12-11
ACS-9044 Remove excluded files from war file for SAST

SaraAspery:


scripts/ci/remove-sast-exclusions.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+echo "=========================== Excluding Files from Veracode SAST ==========================="
+set -ex
+pushd "$(dirname "${BASH_SOURCE[0]}")/../../"
+
+# Copy war file to temporary directory
+/bin/cp -f "$1" "$2"
```
pzhyland, line 7: Why do we need `/bin/cp` instead of just `cp`?

pzhyland (approved):

PR #3092, opened 2024-12-16
fix/MNT-24542 class cast exception

code4uuuu:
https://hyland.atlassian.net/browse/MNT-24542

Here in this ticket. We are encountering an issue when running ACS with Keycloak configured without the OpenID scope. Upon running the sample SDK with the attached files and executing the generated JAR, we encounter a ClassCastException.

The root cause appears to be that while a UserInfoResponse can be successfully cast to a UserInfoSuccessResponse, it cannot be cast to a UserInfoErrorResponse. This issue occurs in the map, where such an invalid cast is attempted.

To address this, I have implemented a fix. The updated logic first retrieves the UserInfoResponse and checks if it is an instance of UserInfoErrorResponse. If it is, the error message is extracted, and a UserInfoException is thrown. Otherwise, the response is returned as-is.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
+++ b/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
@@ -40,6 +40,7 @@ import java.util.function.Predicate;
 
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
+import com.nimbusds.openid.connect.sdk.UserInfoErrorResponse;
 import com.nimbusds.openid.connect.sdk.UserInfoRequest;
 import com.nimbusds.openid.connect.sdk.UserInfoResponse;
 import com.nimbusds.openid.connect.sdk.UserInfoSuccessResponse;
@@ -143,7 +144,18 @@ class SpringBasedIdentityServiceFacade implements IdentityServiceFacade
                 .flatMap(httpResponse -> {
                     try
                     {
-                        return Optional.of(UserInfoResponse.parse(httpResponse));
+                        UserInfoResponse userInfoResponse = UserInfoResponse.parse(httpResponse);
+                        if (userInfoResponse instanceof UserInfoErrorResponse)
+                        {
+                            UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;
+                            String errorMessage = userInfoErrorResponse.getErrorObject().getDescription();
+                            LOGGER.warn("User Info Request failed: " + errorMessage);
+                            throw new UserInfoException(errorMessage);
+                        }
+                        else
+                        {
+                            return Optional.of(userInfoResponse);
+                        }
                     }
                     catch (ParseException e)
                     {
--- a/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
+++ b/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
@@ -44,7 +44,6 @@ import com.nimbusds.openid.connect.sdk.UserInfoErrorResponse;
 import com.nimbusds.openid.connect.sdk.UserInfoRequest;
 import com.nimbusds.openid.connect.sdk.UserInfoResponse;
 import com.nimbusds.openid.connect.sdk.UserInfoSuccessResponse;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.core.convert.converter.Converter;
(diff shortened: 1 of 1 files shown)


repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
```diff
@@ -1,344 +1,355 @@
...
+                                {
+                                    UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;
+                                    String errorMessage = userInfoErrorResponse.getErrorObject().getDescription();
+                                    LOGGER.warn("User Info Request failed: " + errorMessage);
+                                    throw new UserInfoException(errorMessage);
+                                }
+                                else
+                                {
+                                    return Optional.of(userInfoResponse);
+                                }
```
code4uuuu, line 158: I have made the following changes for the fix. First we are getting the userInfoResponse after that checking it whether it is instance of UserInfoErrorResponse, if it is then we are extracting errorMessage and throwing UserInfoException otherwise returning as it is

kavitshah-gl (changes requested):
Need to revert the change, only the part where we are making the changes needs to be formatted. It is really difficult to know what is changed if we just format the whole java file.

code4uuuu:
> Need to revert the change, only the part where we are making the changes needs

Agreed but If we will not format the whole java file then our pre-commit stage is failing

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
```diff
@@ -1,344 +1,355 @@
...
+                                UserInfoResponse userInfoResponse = UserInfoResponse.parse(httpResponse);
+
+                                if (userInfoResponse instanceof UserInfoErrorResponse)
+                                {
+                                    UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;
+                                    String errorMessage = userInfoErrorResponse.getErrorObject().getDescription();
+                                    LOGGER.warn("User Info Request failed: " + errorMessage);
+                                    throw new UserInfoException(errorMessage);
+                                }
+                                else
```
kavitshah-gl, line 155: I don't think we need else here, if the exception is not thrown then by default it will come to this line.

kavitshah-gl (changes requested):
Kindly make the changes accordingly.
Also look to resolve the PMD issues.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
code4uuuu: Done. Removed the else part

code4uuuu:
> Kindly make the changes accordingly. 
**[Remarks]** I have done the changes for the same

Also look to resolve the PMD issues.
**[Remarks]**  These are the PMD warning not the issues, our PMD scan stage already passed in the github action.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
```diff
@@ -1,344 +1,352 @@
...
+                            try
+                            {
+                                UserInfoResponse userInfoResponse = UserInfoResponse.parse(httpResponse);
+
+                                if (userInfoResponse instanceof UserInfoErrorResponse)
+                                {
+                                    UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;
+                                    String errorMessage = userInfoErrorResponse.getErrorObject().getDescription();
+                                    LOGGER.warn("User Info Request failed: " + errorMessage);
+                                    throw new UserInfoException(errorMessage);
```
kavitshah-gl, line 153: Do we need to throw the error exception? Can't we extract the error message and use the LOGGER.warn to show the error and then return the empty Optional value as done for ParseException?

kavitshah-gl (changes requested):
Kindly look into the comment for change requested

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
```diff
@@ -1,344 +1,352 @@
...
+                        })
+                        .flatMap(httpResponse -> {
+                            try
+                            {
+                                UserInfoResponse userInfoResponse = UserInfoResponse.parse(httpResponse);
+
+                                if (userInfoResponse instanceof UserInfoErrorResponse)
+                                {
+                                    UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;
+                                    String errorMessage = userInfoErrorResponse.getErrorObject().getDescription();
```
damianujma, line 151: In general, it looks good to me.
Please replace `UserInfoErrorResponse userInfoErrorResponse = (UserInfoErrorResponse) userInfoResponse;`
with pattern variable:
`if (userInfoResponse instanceof UserInfoErrorResponse userInfoErrorResponse)`.
Moreover, I think we can check if `userInfoErrorResponse.getErrorObject()` is `null`.
For example: 
```
String errorMessage = Optional.ofNullable(userInfoErrorResponse.getErrorObject())
                                            .map(ErrorObject::getDescription)
                                            .orElse("No error description found");
```

damianujma (changes requested):

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
damianujma: I agree with @kavitshah-gl .

code4uuuu: > Do we need to throw the error exception? Can't we extract the error message and use the LOGGER.warn to show the error and then return the empty Optional value as done for ParseException?

If we send the empty response then the Client SDK is not receiving the error at all which does not help users. Further if we continue with the UserInfoErrorResponse further to the map method then it gives ClassCastException. This is happening since UserInfoResponse can be converted to UserInfoSuccessResponse but UserInfoErrorResponse can't ne converted to UserInfoSuccessResponse and in the map it tries to do that . Thus rather than Spring throw the ClassCastException we are throwing UserInfoException which is of type IdentityServiceFacadeException which is caught gracefully in authenticateImpl method.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
code4uuuu: I have addressed the following changes

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/SpringBasedIdentityServiceFacade.java
kavitshah-gl: Okay I think throwing UserInfoException is fine since we are catching it.

kavitshah-gl (approved):
Approved.

damianujma (approved):

purusothaman-mm (approved):
LGTM

pzhyland (approved):

PR #3108, opened 2024-12-20
[APPS-3154] fix for number of file field in response

mohit-singh4:
https://hyland.atlassian.net/browse/APPS-3154
Here in this ticket, The "numberOfFiles" field in the GET/size-details response does not return accurate data for values exceeding 500.
The root cause appears to be the default limit of 500 in the search parameter when executing size-details for more than 500 files.
To address this issue, I made some changes by taking the value from the "NumberFound" field in the result entity, and it is now working correctly.


SathishK-T (approved):
LGTM

purusothaman-mm (approved):
LGTM

vedantmehra9 (approved):

kavitshah-gl (approved):
Looks ok.

pzhyland (approved):

PR #3147, opened 2025-01-15
[MNT-24807] repo event2 is exposing user password hash and salt version 2

cezary-witkowski:
Change description:
By default properties: usr:salt, usr:passwordHash, usr:password and trx:password will have their values replaced with "SENSITIVE_DATA_REMOVED" text.
Example: [3.json](https://github.com/user-attachments/files/18440951/3.json)

Configuration:
repo.event2.filter.nodeProperties (default value empty). Allows to specify properties that should be filtered out. Filtering always happens before mapping.
repo.event2.mapper.enabled (default value: true). Allows to disable property mapping entirely.
repo.event2.mapper.overrideReplacementText (default value: empty). Allows to change the text that will be present in events instead of default: "SENSITIVE_DATA_REMOVED".
repo.event2.mapper.overrideDefaultProperties (default value: empty). Allows to override default list of properties that will have their values replaced instead of default: usr:salt, usr:passwordHash, usr:password and trx:password.

Configuration examples:
Example 1:
-Drepo.event2.mapper.enabled=false

No property mapping will be done. we are back to before this ticket.
_________________
Example 2:
-Drepo.event2.mapper.overrideDefaultProperties=usr:*
-Drepo.event2.mapper.overrideReplacementText=HIDDEN_BY_SECURITY_MEASURES

All properties from user model will have their values replaced by "HIDDEN_BY_SECURITY_MEASURES" text. Property trx:password will NOT have its value replaced, because default config was overridden to only map usr:*.
___________________
Example 3:
-Drepo.event2.filter.nodeProperties=usr:salt,usr:passwordHash,usr:password,trx:password

The properties above will be filtered out completely before any mapping occurs and will not be present in activemq.
___________________
Example 4:
-Drepo.event2.filter.nodeProperties=trx:*

The properties above will be filtered out. Properties usr:salt,usr:passwordHash,usr:password will be mapped due to default configuration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,7 +337,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['10.2.18', '10.4', '10.5']
+        version: ['10.5', '10.6']
     steps:
       - uses: actions/checkout@v4
       - uses: Alfresco/alfresco-build-tools/.github/actions/get-build-info@v8.2.0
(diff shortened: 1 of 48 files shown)


pzhyland (approved):

kmagdziarz (approved):

damianujma (commented):
LGTM. Great work! 
Some classes needs the license update.

PR #3155, opened 2025-01-17
[fix/MNT-24172-fixRecreationOfRendition2] Fix of recreation of rendition2

mohit-singh4:
https://hyland.atlassian.net/browse/MNT-24172

Based on the implemented code, below are the observations encountered during the execution of the RenditionService2 implementation via the POST- nodes/{nodeId}/renditions and DELETE- nodes/{nodeId}/renditions/{renditionId} endpoints:

- The POST endpoint assigns the rn:rendition2 aspect to the respective rendition nodes (e.g., cm:pdf).
- After executing the DELETE endpoint, the CONTENT_PROPERTY is removed from the rendition nodes.
- When attempting to open the document from Share, a condition for Qname-cm:pdf checks for ASPECT_RENDITION2, during which the CONTENT_PROPERTY is retrieved. However, it returns null, causing the document not to open and resulting in a NullPointerException. Since the CONTENT_PROPERTY is removed during the DELETE operation, the renditions for ASPECT2 are returning null content, which leads to NullPointerExceptions.

**`Note`** This validation, however, is not implemented for the normal rendition when previewing the document via Share. In that case, the method always returns true, regardless of whether the rendition has been deleted.

**[Solution]:** Instead of throwing a NullPointerException, we are logging the error. Additionally, we are removing the existing rendition node (since the content data is null) and assigning a newly created rendition node.


SathishK-T (approved):
LGTM

purusothaman-mm (approved):
LGTM

code4uuuu (approved):

vedantmehra9 (approved):

kavitshah-gl (changes requested):
We will need to create the test cases for these changes.

mohit-singh4:
> We will need to create the test cases for these changes.

**Remarks:** In my implementation, the method I modified is private. Instead of throwing a NullPointerException, we are now returning the newly created node. Since this method is private, we cannot create a test case for it.
However, there is already a method in **RenditionServiceIntegrationTest** that checks whether the size is empty or not post-executing deleting rendition node.

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
```diff
@@ -1,1131 +1,1135 @@
...
+                logger.debug(msg);
+            }
+            // Check if the node has the applied renditioned aspect, and if it does,
+            // remove the existing rendition node and assign the newly created rendition node.
+            if (nodeService.hasAspect(sourceNode, RenditionModel.ASPECT_RENDITIONED))
+            {
+                List<ChildAssociationRef> renditions = nodeService.getChildAssocs(sourceNode, RenditionModel.ASSOC_RENDITION, renditionQName);
+                if (!renditions.isEmpty())
+                {
+                    nodeService.removeChild(sourceNode, renditions.get(0)
```
kavitshah-gl, line 1008: Can you make the renditions.get(0) more descriptive? 
For e.g.
ChildAssociationRef existingRendition = renditions.get(0);
nodeService.removeChild(sourceNode, existingRendition.getChildRef());
renditionAssoc = renditionNode;

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
```diff
@@ -1,1131 +1,1135 @@
...
+                            .getChildRef());
+                    renditionAssoc = renditionNode;
+                }
+            }
+            if (logger.isDebugEnabled())
```
kavitshah-gl, line 1013: Shouldn't this if condition be inside the if (!renditions.isEmpty()) condition because the deletion is happening inside that condition, in case the if condition returns false then operation will not occur but the log will be printed which can be misleading.

kavitshah-gl (changes requested):
There are some changes suggested.

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
mohit-singh4: Updated
Done.

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
mohit-singh4: Modified.
Done.

kavitshah-gl (approved):
Approved. Kindly get it reviewed with ACS Lead as well.

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
```diff
@@ -1,1131 +1,1136 @@
...
+            {
+                logger.debug(msg);
+            }
+            // Check if the node has the applied renditioned aspect, and if it does,
+            // remove the existing rendition node and assign the newly created rendition node.
+            if (nodeService.hasAspect(sourceNode, RenditionModel.ASPECT_RENDITIONED))
+            {
+                List<ChildAssociationRef> renditions = nodeService.getChildAssocs(sourceNode, RenditionModel.ASSOC_RENDITION, renditionQName);
+                if (!renditions.isEmpty())
```
kmagdziarz, line 1006: What if `renditions` is empty?

mohit-singh4: The same condition is implemented in the `getRenditionByName` method, where we check for both the **rendition** and **rendition2** aspects. However, this validation is not applied to the standard rendition (i.e., the `rendition` aspect) when previewing the document via the Share feature. In this case, the method always returns `true` from the `isRenditionAvailable` method, regardless of whether the rendition has been deleted.

On the other hand, for `rendition2`, the `isRenditionAvailable` method returns `null`, along with the previous `rendition2` node. If it's `null`, a log message is generated instead of throwing an exception.

mohit-singh4:
> > > We will need to create the test cases for these changes.
> > 
> > 
> > **Remarks:** In my implementation, the method I modified is private. Instead of throwing a NullPointerException, we are now returning the newly created node. Since this method is private, we cannot create a test case for it. However, there is already a method in **RenditionServiceIntegrationTest** that checks whether the size is empty or not post-executing deleting rendition node.
> 
> Since this method is private, can't we add an e2e test to test described scenario?

mohit-singh4:
> > > We will need to create the test cases for these changes.
> > 
> > 
> > **Remarks:** In my implementation, the method I modified is private. Instead of throwing a NullPointerException, we are now returning the newly created node. Since this method is private, we cannot create a test case for it. However, there is already a method in **RenditionServiceIntegrationTest** that checks whether the size is empty or not post-executing deleting rendition node.
> 
> Since this method is private, can't we add an e2e test to test described scenario?

We cannot directly write an "end-to-end" (E2E) test for a private method because the primary principle is to test only public methods.

pzhyland:
> > > > We will need to create the test cases for these changes.
> > > 
> > > 
> > > **Remarks:** In my implementation, the method I modified is private. Instead of throwing a NullPointerException, we are now returning the newly created node. Since this method is private, we cannot create a test case for it. However, there is already a method in **RenditionServiceIntegrationTest** that checks whether the size is empty or not post-executing deleting rendition node.
> > 
> > 
> > Since this method is private, can't we add an e2e test to test described scenario?
> 
> We cannot directly write an "end-to-end" (E2E) test for a private method because the primary principle is to test only public methods.

@mohit-singh4 You've changed the current behaviour. There is no single test confirming that it works correctly. It doesn't matter if the change is in the private method. The primary principle for fixing bug is to write a test, the test should fail, fix the code without changing the test, the test should be successful. Sometimes the Unit test for a single method is not enough, this is why we relay on Integration & E2E tests.

repository/src/main/java/org/alfresco/repo/rendition/executer/AbstractRenderingEngine.java
```diff
@@ -1,1131 +1,1136 @@
...
+        // Handle the rendition aspects
+        manageRenditionAspects(sourceNode, renditionNode);
+
+        // Verify that everything has gone to plan, and nothing got lost on the way!
+        ChildAssociationRef renditionAssoc = renditionService.getRenditionByName(sourceNode, renditionQName);
+        if (renditionAssoc == null)
+        {
+            String msg = "A rendition of name: " + renditionQName + " should have been created for source node: "
+                    + sourceNode;
+            if (logger.isDebugEnabled())
```
pzhyland, line 997: From the description I understand that this issue is caused by the `DELETE` operation not correctly removing the rendition leaving it in some inconsistent state (content is removed, aspect isn't). Have you consider fixing it there?

pzhyland (approved):

PR #3165, opened 2025-01-28
[fix/MNT-23559-FixForRecordsManagementSearch] Fix for limited 500 items in RecordManagement Search

mohit-singh4:
https://hyland.atlassian.net/browse/MNT-23559

**[Concern]** When performing a records search on the RM site, the number of results returned is limited to a maximum of 500 items.

**[Solution]** As per the implementation, it retrieves a static value from the property file, which is why it is restricted to 500 items. However, upon reviewing the code, I noticed that the same method is being called from RMSearchGet, where we have the privilege to set a custom maximum number of items. Therefore, instead of using the default value from the property file, we are now setting the value from the caller method's arguments.

### DIFF
--- a/amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/search/RecordsManagementSearchServiceImpl.java
+++ b/amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/search/RecordsManagementSearchServiceImpl.java
@@ -27,9 +27,16 @@
 
 package org.alfresco.module.org_alfresco_module_rm.search;
 
+import static org.alfresco.module.org_alfresco_module_rm.model.RecordsManagementModel.ASPECT_SAVED_SEARCH;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.springframework.extensions.surf.util.I18NUtil;
 
 import org.alfresco.error.AlfrescoRuntimeException;
 import org.alfresco.model.ContentModel;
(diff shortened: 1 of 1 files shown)


SathishK-T (approved):
LGTM

vedantmehra9 (approved):

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/search/RecordsManagementSearchServiceImpl.java
```diff
@@ -207,31 +213,35 @@ public void setRecordsManagementSearchParameters(RecordsManagementSearchParamete
...
+                .append(")");
 
         // create the search parameters
         SearchParameters searchParameters = new SearchParameters();
         searchParameters.setQuery(fullQuery.toString());
         searchParameters.setLanguage(SearchService.LANGUAGE_FTS_ALFRESCO);
         searchParameters.addStore(StoreRef.STORE_REF_WORKSPACE_SPACESSTORE);
-        searchParameters.setMaxItems(recordsManagementSearchParameters.getMaxItems());
+        int maxItems = rmSearchParameters.getMaxItems();
+        if (maxItems > 0)
```
kavitshah-gl, line 237: What will happen getMaxItems() is zero?

kavitshah-gl (commented):
Requested information.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/search/RecordsManagementSearchServiceImpl.java
mohit-singh4: It will not set any value and will provide the actual data instead.

SathishK-T (approved):
LGTM

code4uuuu (approved):

vedantmehra9 (approved):

purusothaman-mm (approved):
LGTM

kavitshah-gl (approved):
Approved. Lets get it reviewed with ACS Lead as well once.

PR #3190, opened 2025-02-06
ACS 9185 implement rest api for reauthorizing a single user

gerardolenski:
Implementation according to https://github.com/Alfresco/rest-api-explorer/pull/222

In community edition endpoins return `HTTP 501 Not implemented`

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/people/PeopleEntityResource.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/people/PeopleEntityResource.java
@@ -34,6 +34,7 @@ import org.springframework.beans.factory.InitializingBean;
 
 import org.alfresco.model.ContentModel;
 import org.alfresco.rest.api.People;
+import org.alfresco.rest.api.model.AuthCode;
 import org.alfresco.rest.api.model.AuthKey;
 import org.alfresco.rest.api.model.Client;
 import org.alfresco.rest.api.model.PasswordReset;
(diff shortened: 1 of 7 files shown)


remote-api/src/main/java/org/alfresco/rest/api/people/PeopleEntityResource.java
```diff
@@ -245,4 +247,29 @@ public void deauthorizeUser(String personId, Void body, Parameters parameters, W
...
+     * Get the authorization code.
+     *
+     * Not supported in community edition.
+     */
+    @Operation("reauthorization-code")
+    @WebApiDescription(title = "Get the reauthorization code", description = "Get the reauthorization code", successStatus = HttpServletResponse.SC_NOT_IMPLEMENTED)
+    public AuthCode getReauthorizationCode(String personId, Void body, Parameters parameters, WithResponse withResponse)
+    {
+        // functionality is not implemented in community edition
+        return null;
```
cezary-witkowski, line 261: In rest-api-explorer we say that this endpoint returns "application/json". Maybe we should return AuthCode with authorizationCode=null to fulfill the contract? On the other hand that would be an implementation and contradict with NOT_IMPLEMENTED. Something to think about.

gerardolenski: In the rest api explorer there is a information, that endpoint is available only in enterprise edition. In community edition we just implemented the same behavior (HTTP 501) as in deauthorize API.

cezary-witkowski (approved):
Nice work

kmagdziarz (approved):

PR #3260, opened 2025-03-12
[APPS-3242] Fix for facet query for elasticsearch

mohit-singh4:
https://hyland.atlassian.net/browse/APPS-3246

When requesting the size details of a newly created empty folder, the API response returns a status of "FAILED" instead of successfully returning `"sizeInBytes": 0`.

[solution] Previously, we set the faceLimit to 0 for empty folders.
Now, we have set the default Elasticsearch value to 100 to avoid a parsing exception.


SathishK-T (approved):
LGTM

pzhyland (approved):

PR #3303, opened 2025-04-08
MNT-24308 On click of 'My Favorites' under 'Documents' 

DurgDineshsai:
When user clicks on **My Favorites** under **Document** in **Document Library** page, all the links under Documents node are appear grayed out while using elastic search. As elastic search not allowing to search with empty string literal. So handled this fix by passing "default" value instead of empty string literal

### DIFF
--- a/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
+++ b/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
@@ -189,7 +189,7 @@ var Filters =
                filterQuery += "ID:\"" + favourite + "\"";
             }
             
-            if (filterQuery.length !== 0)
+            if (filterQuery.length !== 0 && filterQuery !== "ID:\"\"")
             {
                filterQuery = "+(" + filterQuery + ")";
                // no need to specify path here for all sites - IDs are exact matches
@@ -201,7 +201,7 @@ var Filters =
             else
             {
                // empty favourites query
-               filterQuery = "+ID:\"\"";
+               filterQuery = "+ID:\"default\"";
             }
             
             filterParams.query = filterQuery;


amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
```diff
@@ -189,7 +189,7 @@ var Filters =
                filterQuery += "ID:\"" + favourite + "\"";
             }
             
-            if (filterQuery.length !== 0)
+            if (filterQuery.length !== 0 && filterQuery !== "ID:\"\"")
```
kmagdziarz, line 192: Is possible to add test scenerio for that?

kavitshah-gl: Do we need to hardcode the value?

kavitshah-gl (commented):
I have raised one query.

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
DurgDineshsai: While using elastic search, it is not allowing to search with empty string literal For example **ID:""**. So handled this fix by passing **"default"** value instead of empty string literal. In some scenario's, filterQuery value will be ID:"", to avoid appending the query to filterQuery variable, added the condition in literals.

DurgDineshsai: > Is possible to add a test scenerio for that?

Since the [alfresco-community-repo](https://github.com/Alfresco/alfresco-community-repo) does not contain any existing frontend test scenario's. so, it is currently not possible to add a test scenario for this specified fix.

kmagdziarz (commented):
It is possible to test it somehow to prove that fix is working and not breaking anything?

kavitshah-gl (approved):
Approved but need to test it thoroughly.

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
pzhyland, line 192: It seems it's caused by the differences between the Search Services (solr) and Search Enterprise (elasticsearch). Shouldn't we somehow fix it at the lower level (building the ES query)? Current fix will solve only this case but the general fix migh solve other similar use cases as well.

DurgDineshsai: @pzhyland - You're right -- addressing this at lower level would provide a more generic solution. However to avoid blocking the current progress, we can proceed with current fix for now and can plan the generic approach( handling at ES query level) in upcoming sprint by creating a new story.

PR #3310, opened 2025-04-10
[MNT-24859] Basic Auth still possible with Keycloak enabled

SathishK-T:
https://hyland.atlassian.net/browse/MNT-24859.

Keycloak authentication only handled for Alfresco Administration Console. Now we implemented for Alfresco WebScripts Home as well.

### DIFF
--- a/remote-api/src/main/resources/alfresco/web-scripts-application-context.xml
+++ b/remote-api/src/main/resources/alfresco/web-scripts-application-context.xml
@@ -224,6 +224,7 @@
          <list>
             <value>AdminConsole</value>
             <value>AdminConsoleHelper</value>
+            <value>Index</value>
          </list>
       </property>
    </bean>
--- a/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
+++ b/repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
@@ -239,7 +239,12 @@ public class IdentityServiceAdminConsoleAuthenticator implements AdminConsoleAut
         try
         {
             URI originalUri = new URI(requestURL);
-            URI redirectUri = new URI(originalUri.getScheme(), originalUri.getAuthority(), identityServiceConfig.getAdminConsoleRedirectPath(), originalUri.getQuery(), originalUri.getFragment());
+            String redirectPath = identityServiceConfig.getAdminConsoleRedirectPath();
+            if (requestURL.contains("index"))
+            {
+                redirectPath = originalUri.getPath();
+            }
+            URI redirectUri = new URI(originalUri.getScheme(), originalUri.getAuthority(), redirectPath, originalUri.getQuery(), originalUri.getFragment());
             return redirectUri.toASCIIString();
         }
         catch (URISyntaxException e)
--- a/remote-api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts/index.get
+++ b/remote-api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts/index.get
@@ -1,0 +1,15 @@
+<import resource="classpath:alfresco/templates/webscripts/org/alfresco/repository/admin/admin-common.lib.js">
+
+/**
+ * Repository Admin Console
+ *
+ * Root page GET method
+ */
+function main()
+{
+   status.code = 301;
+   status.location = url.serviceContext + Admin.getDefaultToolURI();
+   status.redirect = true;
+}
+
+main();
--- a/remote-api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts/index.get.desc.xml
+++ b/remote-api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts/index.get.desc.xml
@@ -1,0 +1,11 @@
+<webscript>
+    <shortname>Index - root page</shortname>
+    <description><![CDATA[Index WebScript]]></description>
+    <url>/index</url>
+    <url>/index/</url>
+    <family>Index</family>
+    <format default="html">argument</format>
+    <authentication>sysadmin</authentication>
+    <lifecycle>internal</lifecycle>
+    <transaction allow="readonly">required</transaction>
+</webscript>


purusothaman-mm (approved):
LGTM

mohit-singh4 (approved):
LGTM

code4uuuu (approved):

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
```diff
@@ -70,6 +70,7 @@ public class IdentityServiceAdminConsoleAuthenticator implements AdminConsoleAut
     private static final String ALFRESCO_ACCESS_TOKEN = "ALFRESCO_ACCESS_TOKEN";
     private static final String ALFRESCO_REFRESH_TOKEN = "ALFRESCO_REFRESH_TOKEN";
     private static final String ALFRESCO_TOKEN_EXPIRATION = "ALFRESCO_TOKEN_EXPIRATION";
+    private static final String ALFRESCO_WEB_SCRIPT_PATH = "/alfresco/s/index";
```
kavitshah-gl, line 73: will the path be always coming as /alfresco/s/index for webscripts? Do we need to hardcode the value?

kavitshah-gl (commented):
Can you look into the comments please? Also we need to involve ACS Leads for further review as well.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
purusothaman-mm: Yes, for Alfresco WebScripts Home /alfresco/s/index is the correct URL to use. I’ve removed the hardcoded value and configured under property file

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
```diff
@@ -239,7 +239,12 @@ private String getRedirectUri(String requestURL)
         try
         {
             URI originalUri = new URI(requestURL);
-            URI redirectUri = new URI(originalUri.getScheme(), originalUri.getAuthority(), identityServiceConfig.getAdminConsoleRedirectPath(), originalUri.getQuery(), originalUri.getFragment());
+            String redirectPath = identityServiceConfig.getAdminConsoleRedirectPath();
```
kavitshah-gl, line 242: Can you verify it the existing test cases are covering this code?

kavitshah-gl (commented):
Test cases query

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
purusothaman-mm: The existing test cases do not cover this code. so, I have added two new test cases that cover the code changes.

repository/src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication.properties
```diff
@@ -12,6 +12,7 @@ identity-service.resource=alfresco
 identity-service.credentials.secret=
 identity-service.public-client=true
 identity-service.admin-console.redirect-path=/alfresco/s/admin/admin-communitysummary
+identity-service.webscript-home.redirect-path=/alfresco/s/index
```
kavitshah-gl, line 15: Kindly check to add this in the release notes if we get the approval to implement this.

kavitshah-gl (approved):
I have added comment about the release notes. Kindly get this tested e2e from community to repository to acs-packaging. Also add acs team lead fro reviewing since this is community repo.

damianujma (changes requested):
I tested this solution. 
1. We can enter the main page of Alfresco WebScripts using Keycloak, but **we cannot browse that page**. For example if we select "Browse 'Admin' Web Scripts" **it still requires Basic Auth instead of Keycloak.**
2. I don't like a way how it's implemented. We should not include the solution of the Alfresco WebScripts Auth in the Admin Console components. Alternatively, we can create an abstraction to simplify it.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/admin/IdentityServiceAdminConsoleAuthenticator.java
```diff
@@ -239,7 +239,12 @@ private String getRedirectUri(String requestURL)
...
+            if (originalUri.getPath().equals(identityServiceConfig.getWebScriptHomeRedirectPath()))
```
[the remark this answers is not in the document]
purusothaman-mm: I’ve addressed the requirement by creating a separate class to handle the Alfresco WebScripts Home authentication flow, rather than combining it with the Admin Console's authenticator component.

remote-api/src/main/resources/alfresco/web-scripts-application-context.xml
```diff
@@ -224,6 +224,7 @@
          <list>
             <value>AdminConsole</value>
             <value>AdminConsoleHelper</value>
+            <value>Index</value>
```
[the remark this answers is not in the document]
purusothaman-mm: I have removed Alfresco WebScripts Home from the Admin Console script family and created a separate script family specifically for Alfresco WebScripts Home

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/webscript/IdentityServiceWebScriptHomeAuthenticator.java
```diff
@@ -0,0 +1,329 @@
...
+        return remoteUserMapper.getRemoteUser(decorateBearerHeader(bearerToken, request));
+    }
+
+    @Override
+    public void requestAuthentication(HttpServletRequest request, HttpServletResponse response)
+    {
+        respondWithAuthChallenge(request, response);
+    }
+
+    public void respondWithAuthChallenge(HttpServletRequest request, HttpServletResponse response)
```
damianujma, line 118: This class looks almost the same as `IdentityServiceAdminConsoleAuthenticator`, can we avoid code duplication and abstract it to avoid changes in many places in the future?

damianujma (changes requested):
Why there was a need to add webscripts under remote-`api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts`? How it worked before?

purusothaman-mm:
> Why there was a need to add webscripts under remote-`api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts`? How it worked before?

In the RemoteUserAuthenticator class, the isAdminConsoleWebScript(WebScript webScript) method determines the authentication flow for each WebScript by inspecting the associated metadata.
Specifically, it checks whether the WebScript definition file (located at
remote-api/src/main/resources/alfresco/templates/webscripts/org/springframework/extensions/webscripts/) contains a family field.
If the family field is present: the user is redirected to the corresponding page after successful Keycloak authentication.
If the family field is missing: the system falls back to basic authentication, even if the user is already authenticated via Keycloak.
This logic is consistently applied to all WebScripts under the Admin Console.
If we no longer need to perform this family field check for every WebScript under WebScriptHome, we can simplify the configuration by removing all manually-added WebScript definitions from the aforementioned path.
This would make all WebScriptHome URLs can safely rely on keycloak based authentication without additional metadata-based filtering.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/webscript/IdentityServiceWebScriptHomeAuthenticator.java
purusothaman-mm: As mentioned, I’ve refactored the shared logic into a common structure to avoid duplication between IdentityServiceWebScriptAuthenticator and IdentityServiceAdminConsoleAuthenticator.

PR #3338, opened 2025-05-13
[MNT-24146] - Unable to update password for 'admin' user

DurgDineshsai:
Allowing admin user to update password when user is already enabled or trying to enable. If admin user tries to update the password by unchecking the enabled checkbox or send enabled flag as false through payload will throw error like "Admin authority cannot be disabled"

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/PeopleImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/PeopleImpl.java
@@ -712,7 +712,7 @@ public class PeopleImpl implements People
         Boolean isEnabled = person.isEnabled();
         if (isEnabled != null)
         {
-            if (isAdminAuthority(personIdToUpdate))
+            if (!isEnabled && isAdminAuthority(personIdToUpdate))
             {
                 throw new PermissionDeniedException("Admin authority cannot be disabled.");
             }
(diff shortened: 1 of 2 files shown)


HeinRagas (approved):
Logic seems good to me

MohinishSah (approved):
LGTM

remote-api/src/main/java/org/alfresco/rest/api/impl/PeopleImpl.java
```diff
@@ -712,7 +712,7 @@ public Person update(String personId, final Person person)
         Boolean isEnabled = person.isEnabled();
         if (isEnabled != null)
         {
-            if (isAdminAuthority(personIdToUpdate))
+            if (!isEnabled && isAdminAuthority(personIdToUpdate))
```
kavitshah-gl, line 715: Kindly add the test cases, there should be automated tests for PeopleImpl.

kavitshah-gl (changes requested):
Kindly look into the comments

remote-api/src/main/java/org/alfresco/rest/api/impl/PeopleImpl.java
DurgDineshsai: > Kindly add the test cases, there should be automated tests for PeopleImpl.
The unit test case is already handled previously for enabled flag scenario so no need to add new unit test case.

pzhyland (approved):

damianujma (approved):

kavitshah-gl (approved):
Approved

PR #3375, opened 2025-06-04
[ACS-9697] Backport to 23.N

SatyamSah5:
return appropriate error message when destination folder does not exist while updating folder rules.


bsayan2 (approved):
LGTM

PR #3387, opened 2025-06-16
ACS-6928 handling restricted nodes in queries API

jakubkochman:
Queries API is using following steps to return response:

- call search engine with given parameters
- sort the results
- materialize results into response object

The assumption here is that search engine will return node references to which user has access to, which is not guaranteed. This behavior was only built in Solr based search engine. This causes a bug, when Queries API is attempting to sort the results, because in the context of some user it tries to access restricted node attributes, which eventually are returned as 403 HTTP response.

The fix is to materialize search results and ignore any result to which we don't have access and then sort it.

### DIFF
--- a/remote-api/src/main/java/org/alfresco/rest/api/impl/QueriesImpl.java
+++ b/remote-api/src/main/java/org/alfresco/rest/api/impl/QueriesImpl.java
@@ -2,23 +2,23 @@
  * #%L
  * Alfresco Remote API
  * %%
- * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * Copyright (C) 2005 - 2025 Alfresco Software Limited
  * %%
- * This file is part of the Alfresco software. 
- * If the software was purchased under a paid Alfresco license, the terms of 
- * the paid license agreement will prevail.  Otherwise, the software is 
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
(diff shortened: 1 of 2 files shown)


jakubkochman:
Moved to draft, because I haven't confirmed the fix in enterprise edition.

remote-api/src/main/java/org/alfresco/rest/api/impl/QueriesImpl.java
```diff
@@ -412,34 +385,38 @@ public CollectionWithPagingInfo<T> find(Parameters parameters,
...
                 for (NodeRef nodeRef : nodeRefs)
                 {
-                    T t = convert(nodeRef, includeParam);
-                    collection.add(t);
+                    try
+                    {
+                        T t = convert(nodeRef, includeParam);
+                        collection.put(nodeRef, t);
+                    }
+                    catch (AccessDeniedException ade)
```
kushal-banik-hyland, line 413: Is there other possibilities like similar to `NodeNotFoundException`  kind of ? Are those handled separately(Not counting)

jakubkochman: In elastic search we have `nodeService.exists(refAndScore.nodeRef)`. Briefly checking in SolR there also appear to be something similar in place.

We are good here

kushal-banik-hyland:
Great Fix! @jakubkochman

kushal-banik-hyland (approved):

PR #3443, opened 2025-07-01
ACS-9788 test case to verify node content search

bsayan2:
This test case will ensure that search api is able to fetch files even if user search its content.


packaging/tests/tas-restapi/src/test/java/org/alfresco/rest/content/NodeContentTest.java
```diff
@@ -0,0 +1,72 @@
+package org.alfresco.rest.content;
```
damianujma: Could you please add a license header?

I would move this class to the `org.alfresco.rest.search` package rather than `org.alfresco.rest.content`.

bsayan2: Added license header however most of the test class in org.alfresco.rest is missing this.  
updated package

damianujma (approved):
LGTM. Thanks!

SatyamSah5 (approved):

PR #3446, opened 2025-07-02
Feature/prodsec 10326 update commons fileupload 7.4.N

jakubkochman:


jakubkochman:
@damianujma  I am not sure if master CI also needs to be updated for 7.4.N. I have copied the 25.N ci.yml file and commented out SAST & pre-commit as they didn't work out of the box.

.github/workflows/ci.yml
```diff
@@ -143,24 +245,25 @@ jobs:
...
         run: docker compose -f ./scripts/ci/docker-compose/docker-compose-db.yaml --profile mariadb up -d
         env:
           MARIADB_VERSION: ${{ matrix.version }}
       - name: "Run tests"
-        run: mvn -B test -pl repository -am -Dtest=AllDBTestsTestSuite -DfailIfNoTests=false -Ddb.name=alfresco -Ddb.url=jdbc:mariadb://localhost:3307/alfresco?useUnicode=yes\&characterEncoding=UTF-8 -Ddb.username=alfresco -Ddb.password=alfresco -Ddb.driver=org.mariadb.jdbc.Driver
+        run: mvn -B test -pl repository -am -Dtest=AllDBTestsTestSuite -DfailIfNoTests=false -Ddb.name=alfresco -Ddb.url=jdbc:mariadb://localhost:3307/alfresco?useUnicode=yes\&characterEncoding=UTF-8 -Ddb.username=alfresco -Ddb.password=alfresco -Ddb.driver=org.mariadb.jdbc.Driver # pragma: allowlist secret
       - name: "Clean Maven cache"
         run: bash ./scripts/ci/cleanup_cache.sh
 
-  repository_mariadb_10_6_tests:
```
damianujma, line 162: We should avoid changing tests. We need tests for supported versions, see:
https://docs.alfresco.com/content-services/7.4/support/

Bumping `alfresco-build-tools` is fine.

jakubkochman: There are two jobs defined in this CI. One is using exact version (in this PR it's `10.11`) and another job is using a list of Maria db versions: `['10.5', '10.6']`. Both have the same list of tests and general configuration, they only differ slightly in conditions whether the job should run. 

I'll add a commit with `[db]` marker in message to trigger the db tests.

I'll also double check for other dependencies.

jakubkochman: @damianujma I have reviewed the DB versions tested in 7.4, the DB versions in updated CI and finally cross referenced it with the page you have given. I have removed some tests for newer postgres.

I have kept the updated master as it would probably break post-merge since it also references the deprecated actions.

Can you have a look again?

.github/workflows/ci.yml
```diff
@@ -26,6 +26,9 @@ env:
   CI_WORKSPACE: ${{ github.workspace }}
   TAS_ENVIRONMENT: ./packaging/tests/environment
   TAS_SCRIPTS: ../alfresco-community-repo/packaging/tests/scripts
+  AUTH0_CLIENT_ID: ${{ secrets.AUTH0_OIDC_ADMIN_CLIENT_ID }}
+  AUTH0_CLIENT_SECRET: ${{ secrets.AUTH0_OIDC_CLIENT_SECRET }}
+  AUTH0_ADMIN_PASSWORD: ${{ secrets.AUTH0_OIDC_ADMIN_PASSWORD }}
```
damianujma, line 31: Can we remove AUTH0 vars?

jakubkochman: Auth0 props are used in tests. They don't work in 7.3.N so I removed those props in https://github.com/Alfresco/alfresco-community-repo/pull/3453

PR #3507, opened 2025-08-07
ACS-9844 added extension point for providing additional context configuration files

jakubkochman:
~~If someone wants to create integration tests in the enterprise project by extending `AbstractSingleNetworkSiteTest` , which is based on `EnterpriseTestApi` then it would be missing java beans that are backing some of the enterprise webscripts.~~

Added extension point for custom config location in `EnterpriseTestApi` which in turn are used when setting up Jetty via `EnterprisePublicApiTestFixture`.


bsayan2 (approved):

damianujma:
Please remember to update the license header.

PR #3516, opened 2025-08-12
[ACS-9991] added logic to execute text extraction for content indexing even if thumbnails are disabled.

SatyamSah5:
Whenever thumbnails were disabled, content indexing and metadata extract/embed was skipped (exception being thrown). Added logic to check if the given transform callback is a text extract transform for content indexing or metadata extract/embed and allow it. Remaining callbacks will result in exception being throw, if thumbnails are disabled.

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
+++ b/repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
@@ -78,7 +78,6 @@ import org.alfresco.util.PropertyCheck;
 public class RenditionService2Impl implements RenditionService2, InitializingBean, ContentServicePolicies.OnContentUpdatePolicy
 {
     public static final String TRANSFORMING_ERROR_MESSAGE = "Some error occurred during document transforming. Error message: ";
-    public static final String CONTENT_AVAILABILITY_REPLY_QUEUE = "org.alfresco.search.contentstore.event";
 
     public static final QName DEFAULT_RENDITION_CONTENT_PROP = ContentModel.PROP_CONTENT;
     public static final String DEFAULT_MIMETYPE = MimetypeMap.MIMETYPE_TEXT_PLAIN;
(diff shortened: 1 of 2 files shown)


repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -78,6 +78,7 @@
 public class RenditionService2Impl implements RenditionService2, InitializingBean, ContentServicePolicies.OnContentUpdatePolicy
 {
     public static final String TRANSFORMING_ERROR_MESSAGE = "Some error occurred during document transforming. Error message: ";
+    public static final String CONTENT_AVAILABILITY_REPLY_QUEUE = "org.alfresco.search.contentstore.event";
```
damianujma: Reply queue is configurable so the value cannot be hardcoded.

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -967,4 +968,16 @@ public void forceRenditionsContentHashCode(NodeRef sourceNodeRef)
...
+
+    // Checks if the given transform callback is a text extract transform for content indexing.
+    private boolean isTextExtractTransform(RenderOrTransformCallBack renderOrTransform)
+    {
+        if (!(renderOrTransform.getRenditionDefinition() instanceof TransformDefinition))
+        {
+            return false;
+        }
+        TransformDefinition def = (TransformDefinition) renderOrTransform.getRenditionDefinition();
+        return CONTENT_AVAILABILITY_REPLY_QUEUE.equals(def.getReplyQueue())
```
damianujma: What's the reason of checking queue? Can't we check only the mimetype?

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -288,7 +289,7 @@ private void requestAsyncTransformOrRendition(NodeRef sourceNodeRef, RenderOrTra
     {
         try
         {
-            if (!isEnabled())
+            if (!isEnabled() && !isTextExtractTransform(renderOrTransform))
```
damianujma: This will cause processing even if the `renditionService2.enabled` property is set to `false`. We should avoid it. Check whether the transform is a text extract should be taken into account if `thumbnailsEnabled` is set to `false`.

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -977,7 +976,19 @@ private boolean isTextExtractTransform(RenderOrTransformCallBack renderOrTransfo
             return false;
```
damianujma: Since the `RenditionDefinition2` interface provides `getTargetMimetype()` method I think we should not limit only to `TransformDefinition` which implements  `RenditionDefinition2`. This will allow us to avoid the need for changes in the future for new  `RenditionDefinition2`  implementations. Moreover, we can use pattern matching for `instanceof`.

bsayan2 (approved):

kushal-banik-hyland (approved):

PR #3555, opened 2025-08-29
MNT-24308 : On click of 'My Favorites' link under 'Documents' in the left panel of 'Document Library' page, all the links appear grayed out

debjit-chattopadhyay-hyland:
MNT-24308 : On click of 'My Favorites' link under 'Documents' in the left panel of 'Document Library' page, all the links appear grayed out

### DIFF
--- a/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
+++ b/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
@@ -91,6 +91,15 @@ function doclist_getAllNodes(parsedArgs, filterParams, query, totalItemCount)
    };
 }
 
+function sanitizeJunkFavouriteKeys(favourites){
+   for (var key in favourites) {
+      if (!key || key.trim() === "") {
+         delete favourites[key];
+      }
+   }
+   return favourites;
+}
+
 /**
  * Main entry point: Create collection of documents and folders in the given space
  *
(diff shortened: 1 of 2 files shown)


amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
```diff
@@ -91,6 +91,15 @@ function doclist_getAllNodes(parsedArgs, filterParams, query, totalItemCount)
    };
 }
 
+function sanitizeJunkFavouriteKeys(favourites){
+   for (var key in favourites) {
+      if (!key || key.trim() === "") {
```
bsayan2, line 96: Is it possible to use `!key.trim()` similar to `!key` .  Is it unnecessary as  `!key.trim()` will do other checks as well?

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
```diff
@@ -123,6 +132,28 @@ function doclist_main()
    
    if (logger.isLoggingEnabled())
       logger.log("doclist.lib.js - NodeRef: " + parsedArgs.nodeRef + " Query: " + query);
+
+   favourites = sanitizeJunkFavouriteKeys(favourites);
+
+   if(Object.keys(favourites).length === 0 && query === null)
```
bsayan2, line 138: is there any possibility in JS that `favourites` can be null? If no then fine , if yes whether we would require and additional null check!

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
debjit-chattopadhyay-hyland: it was observed in Share that when favourite key references are removed one by one, once the last reference is removed, the favourites object would be like this 

{"" : true}

Hence this utility function to remove this garbage keys and create a empty object when there are no favourites. 

with key.trim() === "" we are checking whether we have keys of the above mentioned pattern

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/doclist.lib.js
debjit-chattopadhyay-hyland: when there are no favourites marked in share, the object is of the pattern {"":true}

bsayan2 (approved):

PR #3590, opened 2025-09-25
Revert MNT-24776 as this fix is affecting other scenarios.

debjit-chattopadhyay-hyland:
Revert MNT-24776 as this fix is affecting other scenarios. 
Fix for MNT-24776 needs to be analysed with a safe approach


bsayan2 (approved):

Somnath-Deshmukh (commented):
Code looks fine

PR #3604, opened 2025-10-06
MNT-24776 adding if-else conditionals to avoid null values

debjit-chattopadhyay-hyland:
MNT-24776 adding if-else conditionals to avoid null values


debjit-chattopadhyay-hyland:
image

Somnath-Deshmukh (approved):

PR #3640, opened 2025-11-06
[MNT-25404] Query Performance - High performance cost in retrieving nodes/node properties for large result sets

jottley:
- Batch processing of queries for the V1 Search API
- Configuration property added for nodes.bulkLoad.batchSize
- - This property configures the size of the preload batching of the search results (defaults to 256)
- Configuration property added for nodes.bulkLoad.forceBatching
- - This property forces batch querying (IN vs =) in db queries.  This bypasses the check of the state of the caches.
- Post processing of the search results also use batch queries.  Everything should be in the caches but just in case we reduce the number of extra queries needed to build the response


jottley:
Closing.  Superseded by #3694

PR #3844, opened 2026-02-13
ACS-10951 bypass archive store using property "system.content.eagerOrphanCleanup=true"

b-ansari:


repository/src/main/java/org/alfresco/repo/node/db/DbNodeServiceImpl.java
gerardolenski: Please add the logic here. Please invoke cross-testing after the change.

PR #3989, opened 2026-04-08
ACS-11311 Authentication setup for Microsoft Entra

b-ansari:


repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/authentication/AbstractIdentityServiceAuthenticator.java
```diff
@@ -137,13 +137,17 @@ public String buildAuthRequestUrl(HttpServletRequest request)
...
+
         UriComponentsBuilder builder = UriComponentsBuilder.fromUriString(clientRegistration.getProviderDetails()
                 .getAuthorizationUri())
                 .queryParam("client_id", clientRegistration.getClientId())
                 .queryParam("redirect_uri", getRedirectUri(request.getRequestURL().toString()))
                 .queryParam("response_type", "code")
                 .queryParam("scope", String.join("+", getConfiguredScopes(clientRegistration)))
-                .queryParam("state", state.toString());
+                .queryParam("state", state.toString())
+                .queryParam("nonce", nonce);
```
gerardolenski: Based on the OIDC spec (https://openid.net/specs/openid-connect-core-1_0.html 3.1.2.1.  Authentication Request), the `nonce` param is optional. I'd recommend to implement it conditionally with additional paramer to be sure we won't introduce problems to other IDP providers.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -479,7 +484,8 @@ private ClientRegistration.Builder createBuilder(OIDCProviderMetadata metadata)
                     .userNameAttributeName(usernameAttribute)
                     .scope(getSupportedScopes(metadata.getScopes()))
                     .providerConfigurationMetadata(createMetadata(metadata))
-                    .authorizationGrantType(AuthorizationGrantType.TOKEN_EXCHANGE);
+                    .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
```
gerardolenski: Why you changed that? Is it backward compatible? I didn't find any information about this change in description of your ticket.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
```diff
@@ -302,6 +303,57 @@ public void shouldUseCustomIssuerAttribute()
         }
     }
 
+    // =====================================================
+    // MS Entra-specific test cases for scope validation toggle
```
gerardolenski: It is not Entra specific. This is generic configuration that can be used with any IDP provider. Pls refactor these tests to be generic.

gerardolenski (changes requested):
The changes in IdentityServiceFacadeFactoryBean looks dangerous and my be not backward compatible. We should not change it in this MNT. Please do minimal changes to IDP configuration that will enable MS Entra. Based on your investigation in ticket - it should be just implementation of  identity-service.scope-validation.disabled flag

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
b-ansari: removing this change, since the change is not affecting the flow and stamping the warning as :

 WARN  [client.registration.ClientRegistration$Builder] [http-nio-8080-exec-2] clientSettings.isRequireProofKey=true is only valid with authorizationGrantType=AuthorizationGrantType{value='authorization_code'}. Got authorizationGrantType=AuthorizationGrantType{value='urn:ietf:params:oauth:grant-type:token-exchange'}. Resetting to clientSettings.isRequireProofKey=false
 
Spring Security's [ClientRegistration.Builder] defaults requireProofKey to true when we don't explicitly set it. (PKCE) is only meaningful for the authorization_code grant type. When the builder sees you've set the grant type to [TOKEN_EXCHANGE], it detects the mismatch and auto-resets requireProofKey to false, logging that warning.
 
To suppress the warning, we can add [.clientSettings(ClientSettings.builder().requireProofKey(false).build())] before the grant type line — but it's purely cosmetic since Spring already handles it.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/authentication/AbstractIdentityServiceAuthenticator.java
b-ansari: tested the flow now, it works without the nonce param. Removing it as the param is optional

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/authentication/AbstractIdentityServiceAuthenticator.java
```diff
@@ -137,13 +137,19 @@ public String buildAuthRequestUrl(HttpServletRequest request)
...
```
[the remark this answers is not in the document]
b-ansari: fixed

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceConfig.java
```diff
@@ -78,6 +78,12 @@ public class IdentityServiceConfig
     private long jwtClockSkewMs;
     private String webScriptsHomeRedirectPath;
     private String webScriptsHomeScopes;
+    private boolean scopeValidationDisabled;
+
+    public boolean isScopeValidationDisabled()
```
gerardolenski, line 83: Can you move this getter method to the place where is a setter? I don't like both methods are completely separated in this class. Lets keep it together.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -510,6 +511,17 @@ private Builder configureClientAuthentication(Builder builder)
 
         private Set<String> getSupportedScopes(Scope scopes)
         {
+            if (config.isScopeValidationDisabled())
```
gerardolenski, line 513: Here is somethig starange in your code. If I understand correctly, original flow filters scopes retrieved from IDP provider with configured passwordGrantScopes. So originally we receive the subset of configured PasswordGrantScopes 

But in ScopeValidationDisabled block you added additional scopes:  AdminConsoleScopes, WebScriptsHomeScopes. So it is not only bypassing the filtering but also extending scopes. 
So why you don't return just only all PasswordGrantScopes?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -510,6 +511,17 @@ private Builder configureClientAuthentication(Builder builder)
...
+            {
+                // Bypass scope filtering against the IDP's scopes_supported discovery metadata.
+                // Required for IDPs like MS Entra that do not advertise custom API scopes
+                Set<String> allScopes = new LinkedHashSet<>();
```
arindam-roy-hld: If we really don't need to preserve the insertion order, we can use `HashSet`?

arindam-roy-hld (changes requested):

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
b-ansari: You're right — the intent of isScopeValidationDisabled is to bypass the IDP scope filtering, not to extend the scope set. I'll simplify the isScopeValidationDisabled block to just return all configured PasswordGrantScopes without filtering.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
b-ansari: proceeding with LinkedHashSet - Makes logs, debug output, and test assertions deterministic.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -510,6 +511,12 @@ private Builder configureClientAuthentication(Builder builder)
 
         private Set<String> getSupportedScopes(Scope scopes)
         {
+            if (config.isScopeValidationDisabled())
+            {
+                // Bypass scope filtering against the IDP's scopes_supported discovery metadata.
+                // Required for IDPs like MS Entra that do not advertise custom API scopes
+                return new LinkedHashSet<>(config.getPasswordGrantScopes());
```
gerardolenski: The `getPasswordGrantScopes` returns the immutable set. Why you changed it to mutable implementation?

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
```diff
@@ -302,6 +303,52 @@ public void shouldUseCustomIssuerAttribute()
...
+            // Simulate: scopes_supported does NOT include custom API scope
+            when(oidcResponse.getScopes()).thenReturn(new Scope("openid", "profile", "email", "offline_access"));
+            providerMetadata.when(() -> OIDCProviderMetadata.parse(any(String.class))).thenReturn(oidcResponse);
+
+            ClientRegistration clientRegistration = new ClientRegistrationProvider(config).createClientRegistration(
+                    restTemplate);
+
+            // Custom API scope should be included when scope validation is disabled
+            assertThat(clientRegistration.getScopes()).contains(CUSTOM_API_SCOPE);
+            assertThat(clientRegistration.getScopes()).contains("openid", "profile", "email");
```
gerardolenski: Shouldn't it rather be:
```
assertThat(clientRegistration.getScopes()).containsExactlyInAnyOrder("openid", "profile", "email", CUSTOM_API_SCOPE);
```

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/ClientRegistrationProviderUnitTest.java
```diff
@@ -302,6 +303,52 @@ public void shouldUseCustomIssuerAttribute()
...
+            // Simulate IDP: scopes_supported does NOT include custom API scope
+            when(oidcResponse.getScopes()).thenReturn(new Scope("openid", "profile", "email", "offline_access"));
+            providerMetadata.when(() -> OIDCProviderMetadata.parse(any(String.class))).thenReturn(oidcResponse);
+
+            ClientRegistration clientRegistration = new ClientRegistrationProvider(config).createClientRegistration(
+                    restTemplate);
+
+            // Custom API scope should be filtered out when scope validation is enabled
+            assertThat(clientRegistration.getScopes()).doesNotContain(CUSTOM_API_SCOPE);
+            assertThat(clientRegistration.getScopes()).contains("openid", "profile", "email");
```
gerardolenski: Same here, I think we should be sure what exactly is returned, e.g. if we don't have more elements:
```
assertThat(clientRegistration.getScopes()).containsExactlyInAnyOrder("openid", "profile", "email");
```

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/authentication/AbstractIdentityServiceAuthenticator.java
gerardolenski: What with headers update? Pls do it here and in other classes

gerardolenski (changes requested):

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/authentication/AbstractIdentityServiceAuthenticator.java
```diff
@@ -156,6 +156,11 @@ public String buildAuthRequestUrl(HttpServletRequest request)
 
     private Set<String> getConfiguredScopes(ClientRegistration clientRegistration)
     {
+        if (identityServiceConfig.isScopeValidationDisabled())
```
gerardolenski, line 159: If I'm not mistaken, you didn't write any test for this code change.

gerardolenski (changes requested):

gerardolenski (approved):
Finally LGTM. Please remember to prepare documentation request for new parameter.

cezary-witkowski (approved):
Looks good

PR #4011, opened 2026-04-14
ACS-11448 Node Size API fix for over 1k facets

evasques:
* Correction of the logic for when we have more than the defaultItems of facets
* Added unit test


tiagosalvado10 (approved):
LGTM

PR #4061, opened 2026-05-04
ACS-11422 - CVE Vulnerabilities in ACS 25.x

arindam-roy-hld:


repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -827,6 +829,10 @@ public CustomJOSEObjectTypeVerifier(JOSEObjectType... allowedTypes)
         @Override
         public void verify(JOSEObjectType type, SecurityContext context) throws BadJOSEException
         {
+            if (type == null)
+            {
+                return;
+            }
             super.verify(type, context);
```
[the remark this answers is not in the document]
arindam-roy-hld: existing tests handles that. Rather, this fix was added because the existing tests were failing.

`java.lang.NullPointerException: Cannot invoke "Object.equals(Object)" because "o" is null at java.base/java.util.ImmutableCollections$Set12.contains(ImmutableCollections.java:817) at com.nimbusds.jose.proc.DefaultJOSEObjectTypeVerifier.verify(DefaultJOSEObjectTypeVerifier.java:144) at org.alfresco.repo.security.authentication.identityservice.IdentityServiceFacadeFactoryBean$CustomJOSEObjectTypeVerifier.verify(IdentityServiceFacadeFactoryBean.java:833)failing in: static class CustomJOSEObjectTypeVerifier extends DefaultJOSEObjectTypeVerifier<SecurityContext> { public CustomJOSEObjectTypeVerifier(JOSEObjectType... allowedTypes) { super(Set.of(allowedTypes)); } @Override public void verify(JOSEObjectType type, SecurityContext context) throws BadJOSEException { super.verify(type, context); } }`

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -827,6 +829,10 @@ public CustomJOSEObjectTypeVerifier(JOSEObjectType... allowedTypes)
         @Override
         public void verify(JOSEObjectType type, SecurityContext context) throws BadJOSEException
         {
+            if (type == null)
```
gerardolenski: `blocker` Looks like this change introduces security breach. I see in the base class conditional bypassing type verification:
```
if (type == null && ! allowedTypes.contains(null)) {
			throw new BadJOSEException("Required JOSE header typ (type) parameter is missing");
		}
```
And this condition is never used in our solution when OAuth2.0 token is configured, because we always configured it this way:
```
new CustomJOSEObjectTypeVerifier(JOSEObjectType.JWT, AT_JWT)
```
which means JWT typ header validation accepts both Standard JSON Web Token and OAuth 2.0 Access Token. But never accepts null type.
Now you bypasssed verification globally. I can't understand why you did it.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
```diff
@@ -625,6 +625,7 @@ private NimbusJwtDecoder buildJwtDecoder(RestOperations rest, ProviderDetails pr
                 final RSAPublicKey publicKey = parsePublicKey(config.getRealmKey());
                 return NimbusJwtDecoder.withPublicKey(publicKey)
                         .signatureAlgorithm(DEFAULT_SIGNATURE_ALGORITHM)
+                        .jwtProcessorCustomizer(this::reconfigureJWKSCache)
```
gerardolenski: This was not changed even on master where Spring Security 7 was introduced. Quite don't understand the reason of it.

gerardolenski (changes requested):

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
arindam-roy-hld: removed.

gerardolenski (approved):
Now LGTM.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceFacadeFactoryBean.java
arindam-roy-hld: Discussed and resolved.

PR #4073, opened 2026-05-11
Implement caching technique for limiting the ACS/KeyCloak communication

Somnath-Deshmukh:
Added Cache implementation to ACS/KeyCloak communication with distributed, near cache enabled, modifiable configuration.


repository/src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication-context.xml
```diff
@@ -1,274 +1,300 @@
...
+      or access/refresh tokens. The cache is disabled by default and fails closed when the shared
+      secret is not configured.
+
+      Backing cache region: cache.identityServiceCredentialValidationCache (cache-context.xml,
+      caches.properties).
+   -->
+   <bean id="identityServiceCredentialValidationCache"
+         class="org.alfresco.repo.security.authentication.identityservice.cache.DefaultCredentialValidationCache">
+      <constructor-arg ref="identityServiceCredentialValidationBackingCache" />
+      <constructor-arg value="${identity-service.authentication.credentialValidationCache.enabled:}" />
```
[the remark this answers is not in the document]
Somnath-Deshmukh: Agreed. Resolved.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponent.java
```diff
@@ -81,6 +96,24 @@ public void authenticateImpl(String userName, char[] password) throws Authentica
...
+        {
+            final Optional<CredentialValidationCacheEntry> cached = credentialValidationCache.get(userName, password);
+            if (cached.isPresent())
+            {
+                if (LOGGER.isDebugEnabled())
+                {
+                    LOGGER.debug("Credential validation cache HIT for user '" + userName + "'. Skipping authorization request.");
+                }
+                setCurrentUser(cached.get().getNormalizedUsername());
+                return;
```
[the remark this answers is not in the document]
Somnath-Deshmukh: Added helper method cachedPrincipalIsStillProvisioned to handle this.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
```diff
@@ -0,0 +1,230 @@
...
+ * <b>Encapsulation:</b> all cache behaviour (key derivation, lookup, validity check, invalidation, population) lives entirely in this class. Consumers see only the {@link UserTokenProvider} contract; they never observe a cache, an entry, or a key. This makes it safe to swap caching on/off at the wiring layer without touching consumer code.
+ * </p>
+ *
+ * <p>
+ * <b>Storage scope:</b> the cached access token is a bearer credential and MUST NOT leave the JVM that obtained it. The backing {@link SimpleCache} is therefore expected to be configured as {@code cluster.type=local}. Cache keys are SHA-256 digests of {@code userName + 0x00 + password} encoded as base64url, so the password and user name never appear in cleartext in the cache.
+ * </p>
+ */
+public class CachingUserTokenProvider implements UserTokenProvider
+{
+    private static final Log LOGGER = LogFactory.getLog(CachingUserTokenProvider.class);
```
pzhyland: Please switch to Slf4j and use template logging. It allows to remove the if debug enabled checks.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
```diff
@@ -0,0 +1,230 @@
...
+        return fresh;
+    }
+
+    private UserToken readCache(String cacheKey)
+    {
+        try
+        {
+            return backingCache.get(cacheKey);
+        }
+        catch (RuntimeException ex)
```
pzhyland: Don't hide RuntimeException from the Cache, let it fail

Somnath-Deshmukh: Okay. Done

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
Somnath-Deshmukh: Agreed. Done

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserToken.java
```diff
@@ -0,0 +1,99 @@
...
+        return Objects.hash(normalizedUsername, tokenString);
+    }
+
+    /**
+     * Deliberately omits the token value: only the principal name is exposed so accidental {@code toString()} logging cannot leak the bearer credential.
+     */
+    @Override
+    public String toString()
+    {
+        return "UserToken{normalizedUsername='" + normalizedUsername + "'}";
```
gerardolenski: ACS has quite different standard: it uses `[]` instead of `{}`. Eg:
```
public String toString()
    {
        return "NodeMetaData [id=" + id + ", nodeRef=" + nodeRef + ", type=" + type + ", aclId=" + aclId
                + ", properties=" + properties + ", aspects=" + aspects + ", paths=" + paths + "]";
    }
```

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserToken.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * <p>
+ * Carries the normalized principal name (the canonical user identifier resolved by the IdP, which may differ from the credentials supplied by the caller) and the access-token string returned by the IdP. The token is exposed so that components (e.g., a caching {@link UserTokenProvider}) can re-validate it locally without contacting the IdP again.
+ * </p>
+ *
+ * <p>
+ * <b>Bearer-credential handling:</b> {@link #getTokenString()} returns a bearer credential. Callers MUST NOT log, persist, or replicate it across processes. {@link #toString()} deliberately omits the token to reduce the chance of accidental leakage.
+ * </p>
+ */
+public final class UserToken implements Serializable
```
gerardolenski: Looks like it could be converted to a record

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserTokenProvider.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param userName
+     *            the user name presented by the caller
+     * @param password
+     *            the password presented by the caller; never stored as-is by any implementation
+     * @return a validated {@link UserToken}
+     * @throws org.alfresco.repo.security.authentication.identityservice.IdentityServiceFacade.IdentityServiceFacadeException
+     *             if the Identity Service rejects the credentials or cannot be reached
+     */
+    UserToken getUserToken(String userName, char[] password);
```
gerardolenski: It is always a good approach to use value objects when designing interfaces,instead of list of primitives. Eg in this situation you could introduce something similar to
```
record UserTokenRequest(String username, char[] password) {}

UserToken getUserToken(UserTokenRequest tokeRequest);
```
Then even such value object could vaildate itself if fields are correctly set in constructor.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
```diff
@@ -0,0 +1,208 @@
...
+     * Derive the cache key as base64url(SHA-256(userName || 0x00 || password)). The single-byte separator prevents accidental collisions between e.g. ("ab", "cd") and ("a", "bcd"). The intermediate password byte buffer is wiped after use.
+     */
+    @SuppressWarnings("PMD.UseVarargs")
+    private static String deriveCacheKey(String userName, char[] password)
+    {
+        byte[] passwordBytes = null;
+        try
+        {
+            final MessageDigest digest = MessageDigest.getInstance(DIGEST_ALGORITHM);
+            digest.update(safe(userName).getBytes(StandardCharsets.UTF_8));
```
gerardolenski: So I'm thinking when exactly we can have a situation when username/password is null. And should we even allow to build such cache key? 

BTW. If you create value object for UserTokenRequest then it would be validated before ;)
And I see also 
```
org.alfresco.repo.security.authentication.identityservice.IdentityServiceFacade.AuthorizationGrant#password

return new AuthorizationGrant(requireNonNull(username), requireNonNull(password)
```

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
```diff
@@ -0,0 +1,208 @@
...
+ * </p>
+ *
+ * <p>
+ * <b>Storage scope:</b> the cached access token is a bearer credential and MUST NOT leave the JVM that obtained it. The backing {@link SimpleCache} is therefore expected to be configured as {@code cluster.type=local}. Cache keys are SHA-256 digests of {@code userName + 0x00 + password} encoded as base64url, so the password and user name never appear in cleartext in the cache.
+ * </p>
+ */
+public class CachingUserTokenProvider implements UserTokenProvider
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(CachingUserTokenProvider.class);
+    private static final String DIGEST_ALGORITHM = "SHA-256";
```
gerardolenski, line 63: In Jira description you were refering HMAC-SHA256 algorithm. Here you are using just plain hash. What is the reason of this change? Is it intentional?

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserTokenProviders.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Spring helper that selects the {@link UserTokenProvider} implementation to expose to consumers based on the {@code identity-service.authentication.userTokenCache.enabled} flag.
+ *
+ * <p>
+ * Both the direct and the caching providers are constructed by Spring at startup; this helper merely returns one of them. Switching the flag therefore requires only a context reload, not a code change. Consumers receive {@code UserTokenProvider} and remain unaware of which implementation backs them.
+ * </p>
+ */
+public final class UserTokenProviders
```
gerardolenski, line 38: To me this name is confusing. Maybe rather UserTokenProviderFactory, UserTokenProviderSelector, UserTokenProviderConfig or sth similar?

repository/src/main/resources/alfresco/caches.properties
```diff
@@ -1,738 +1,762 @@
-#
```
gerardolenski, line 1: What has happened here and other config files?

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProviderTest.java
```diff
@@ -0,0 +1,211 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import org.alfresco.repo.cache.MemoryCache;
+import org.alfresco.repo.security.authentication.identityservice.IdentityServiceFacade.TokenDecodingException;
+
+/**
+ * Pure unit tests for {@link CachingUserTokenProvider}: cache hits, misses, invalidation triggered by failed local re-validation, and the property that the SHA-256 cache key is sensitive to both user name and password.
+ */
+public class CachingUserTokenProviderTest
```
gerardolenski, line 64: I doubt that you can relay only on unit tests for this feature. Shouldn't integration and E2E tests be considered as well? I think for such feature we should run at least selected tests on CI for community, maybe also for enterprise and packaging by enabling introduced cache property. Right now you even don't know if context configuration you introduced is working. Did you check what kind of existing IT/E2E tests would be suitable for it? I suggest checking it and then consider what can be reused just in the testing environment with enabled cache.

gerardolenski (commented):
It is important performance feature in security area. Overall looks good to me, however I left few comments suggesting few possible improvements

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserToken.java
Somnath-Deshmukh: Okay. Done

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
Somnath-Deshmukh: My bad :) . Forgot to update this in the JIRA. Will update right way. And yes, it was intentional as doing HMAC will be overkill for this.

repository/src/main/resources/alfresco/caches.properties
Somnath-Deshmukh: In this caches.properties these properties have been added. L694 -701. Somehow the git changes are missing the indentation

cache.identityServiceUserTokenCache.maxItems=10000
cache.identityServiceUserTokenCache.timeToLiveSeconds=900
cache.identityServiceUserTokenCache.maxIdleSeconds=0
cache.identityServiceUserTokenCache.cluster.type=local
cache.identityServiceUserTokenCache.backup-count=1
cache.identityServiceUserTokenCache.eviction-policy=LRU
cache.identityServiceUserTokenCache.merge-policy=com.hazelcast.spi.merge.PutIfAbsentMergePolicy
cache.identityServiceUserTokenCache.readBackupData=false

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserTokenProviders.java
Somnath-Deshmukh: This class is named like this as decorator pattern is followed here and as a utility class this deciding the Token Provider.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserToken.java
Somnath-Deshmukh: Great Suggestion. Yes. This can be a record. Refactored this.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/UserTokenProvider.java
Somnath-Deshmukh: Great suggestion! Earlier I kept this as this convention was followed in the earlier authentication code. However, I have refactored this as follows:

- Added a new UserTokenRequest record - compact constructor requireNonNulls both fields (mirrors AuthorizationGrant.password). toString() overridden to suppress the password.
- DirectUserTokenProvider — accepts the request, extracts username() / password() for the existing AuthorizationGrant.password call.
- CachingUserTokenProvider — accepts the request; safe() helper deleted; chars == null branch in charsToBytes deleted; the request guarantees both fields are non-null so the dead code is gone.
- IdentityServiceAuthenticationComponent — wraps (userName, password) in a UserTokenRequest once at the entry point. The actual record-construction null-check happens before any cache key derivation, so any null reaching this component fails fast with a clear NullPointerException instead of silently producing a stable "all-nulls" cache key.
- Tests — call sites updated; three new tests assert UserTokenRequest validates inputs and doesn't echo the password in toString(). Mock stubs use a small forUser(...) matcher (argThat(req -> username.equals(req.username()))) so verifications still pin to a specific user without caring about the password's char[] identity.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
Somnath-Deshmukh: I kept this as a fail-safe earlier. However, I have refactored this as suggested.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProviderTest.java
Somnath-Deshmukh: Fair point. Happy to extend coverage. However, In the CI AppContext05TestSuite already covers the IdentityServiceAuthenticationComponentTest. But as you correctly pointed out the gap on the testing I have added the IdentityServiceUserTokenWiringTest class which loads the identity-service subsystem context in isolation, stubs the external dependencies, and asserts the factory selects the right UserTokenProvider implementation based on the property.

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private final UserTokenProvider delegate;
+    private final SimpleCache<String, UserToken> backingCache;
+    private final IdentityServiceFacade identityServiceFacade;
+
+    public CachingUserTokenProvider(
+            UserTokenProvider delegate,
+            SimpleCache<String, UserToken> backingCache,
```
gerardolenski, line 71: Consider using dedicated class eg `UserTokenKey` instead of `String`. This class should be immutable and could conatin the logic of key building inside constructor.

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProviderTest.java
```diff
@@ -0,0 +1,252 @@
...
```
gerardolenski, line 64: Consider adding unit tests for concurrent setup. Eg:
- few threads are constantly reading Token
- then token expires
- then all threads starting receiving new token

repository/src/main/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProvider.java
Somnath-Deshmukh: I agree in this principle, but the key is purely internal to CachingUserTokenProvider (never leaves the class, never appears in any public API)

repository/src/test/java/org/alfresco/repo/security/authentication/identityservice/CachingUserTokenProviderTest.java
Somnath-Deshmukh: MemoryCache is ConcurrentHashMap-backed, so it is thread-safe. However added concurrentSetupTestonRefTokenExpire test for the concurrent setup

gerardolenski (approved):
LGTM

pzhyland (approved):

PR #4077, opened 2026-05-12
Blocking REST API calls that attempt to inject malicious files

SanjoyHyland2025:
Fix: Block REST API modifications to system-managed repository paths (ACS-11595)

Problem
An authenticated admin could exploit FreeMarker template SSTI via the REST API to overwrite system web script files (e.g., Data Dictionary/Web Scripts/...), enabling sandbox escape and full RCE as the alfresco system user.

Solution
Added checkNotSystemPath(NodeRef) guard in NodesImpl that walks the ancestor chain of the target node. If any ancestor is a system/special node (Data Dictionary, Sites, etc.), a PermissionDeniedException is thrown — blocking the operation before it reaches the repository.
The check is applied to:
deleteNode
createNode
updateContent
moveOrCopyNode (both source and target)

image


remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3961,6 +3966,43 @@ public void validateProperties(Map<String, Object> properties, List<String> excl
...
+            if (current.equals(repositoryHelper.getCompanyHome()))
+            {
+                break;
+            }
+
+            QName type = nodeService.getType(current);
+            if (isSpecialNode(current, type))
+            {
+                throw new PermissionDeniedException(
+                        "Cannot perform operation on a system path node"
```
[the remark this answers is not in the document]
AbhraPoddar: This is a good suggestion

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3961,6 +3966,43 @@ public void validateProperties(Map<String, Object> properties, List<String> excl
...
+        while (current != null)
+        {
+            if (current.equals(repositoryHelper.getCompanyHome()))
+            {
+                break;
+            }
+
+            QName type = nodeService.getType(current);
+            if (isSpecialNode(current, type))
+            {
```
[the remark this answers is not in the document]
AbhraPoddar: Please check this use case. This is very important ,it seems

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -4000,6 +3966,39 @@ public void validateProperties(Map<String, Object> properties, List<String> excl
         }
     }
 
+    /**
+     * Blocks any REST API modification on system-managed repository paths. Walks up the parent chain and throws PermissionDeniedException if any ancestor is a system/special node (Data Dictionary, Sites, etc.)
```
AbhraPoddar: Please make this method's description in multi-lines, for better readability perspective.

remote-api/src/test/java/org/alfresco/rest/api/impl/NodesImplSystemPathTest.java
```diff
@@ -0,0 +1,173 @@
...
+        {
+            nodesImpl.deleteNode(childRef.getId(), parameters);
+        }
+        catch (PermissionDeniedException e)
+        {
+            fail("Normal node under Company Home should not throw PermissionDeniedException");
+        }
+        catch (Exception e)
+        {
+            // Other exceptions (e.g. NullPointerException from unmocked services deeper in deleteNode)
```
AbhraPoddar: Can this 'catch' block written in different way ? Without any operation , an empty catch block may incur some confusions. It seems the actual intended test case around checkNotSystemPath , would not be hampered by this exception , but better to have some message  or handle it differently would be more clear-cut.

remote-api/src/test/java/org/alfresco/rest/api/impl/NodesImplSystemPathTest.java
```diff
@@ -0,0 +1,173 @@
...
+        when(nodeService.exists(companyHomeRef)).thenReturn(true);
+
+        assertThrows(PermissionDeniedException.class,
+                () -> nodesImpl.deleteNode(companyHomeRef.getId(), parameters));
+
+        verify(nodeService, never()).getPrimaryParent(companyHomeRef);
+    }
+
+    @Test
+    public void testDirectParentIsSpecialNode_ThrowPermissionDenied()
```
AbhraPoddar, line 144: Would be better if only 'PermissionDenied' added as a suffix , instead of 'ThrowPermissionDenied'.

AbhraPoddar (commented):
Couple of in general observations - 
1. in setup() method -the objects (dataDictionaryRef  , comanyHomeRef ) will be generated each time before executing each test. For three / four tests in this class , creation of those objects repeatedly  may not be impacted and but they will remain in memory even after executing the tests. Now in future if more test cases need to be added, in this test class , then it might not be a good thing . Rather create a tearDown() method with @After and make those objects null inside that. I believe that could a good approach.
2. Although there are no rigid rules, but in all test methods after underscore, if suffix can be started with small case and then maintain with that suffix with camelCase , then that would be better. I have seen in the same package, under different test classes that has been followed. Ex : testNodeWithSpecialAncestor_permissionDenied or testNodeWithSpecialAncestor_deniedPermission.

Please check co-pilot review comments also, if relevant and need be.

remote-api/src/test/java/org/alfresco/rest/api/impl/NodesImplSystemPathTest.java
SanjoyHyland2025: done

remote-api/src/test/java/org/alfresco/rest/api/impl/NodesImplSystemPathTest.java
SanjoyHyland2025: done

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
SanjoyHyland2025: because of the precommit this creates a single line. but any ways i have reduced the size.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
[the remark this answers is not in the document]
SanjoyHyland2025: done

AbhraPoddar (commented):
Please check test method's naming convention is parity with other existing test cases , especially which having underscore

AbhraPoddar (approved):

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3806,14 +3774,14 @@ public QName createQName(String qnameStr)
...
+        String prefix = PARAM_INCLUDE_PROPERTIES + "/";
 
         List<QName> result = new ArrayList<>(qnameStrList.size());
         for (String str : qnameStrList)
         {
-            if (str.startsWith(PREFIX))
+            if (str.startsWith(prefix))
             {
-                str = str.substring(PREFIX.length());
+                str = str.substring(prefix.length());
```
jottley: This will/should fail build checks. Please evaluate how you can address this.

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -2186,6 +2187,7 @@ public Node createNode(String parentFolderNodeId, Node nodeInfo, Parameters para
 
         // check that requested parent node exists and it's type is a (sub-)type of folder
         NodeRef parentNodeRef = validateOrLookupNode(parentFolderNodeId);
+        checkNotSystemPath(parentNodeRef);
```
jottley: Evaluate Co-Pilots review.  If the logic is correct add a comment to the code to explain why this works or should be done how you have it so that co-pilot will understand the logic.

jottley (changes requested):

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3961,6 +3966,43 @@ public void validateProperties(Map<String, Object> properties, List<String> excl
...
     }
 
+    /**
+     * Blocks any REST API modification on system-managed repository paths.
+     * Walks up the parent chain and throws PermissionDeniedException
+     * if any ancestor is a system/special node (Data Dictionary, Sites, etc.)
+     */
+    private void checkNotSystemPath(NodeRef nodeRef)
+    {
+        NodeRef current = nodeRef;
```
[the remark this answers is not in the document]
SanjoyHyland2025: The allow/deny matrix is already covered by the unit tests added in NodesImplSystemPathTest.java:
Deny: write under Data Dictionary (special ancestor) → testNodeWithSpecialAncestor_PermissionDenied
Deny: direct special parent → testDirectParentIsSpecialNode_ThrowPermissionDenied
Allow: normal site content path → testNodeWithNormalAncestors_Succeed
Boundary: loop stops at Company Home → testCompanyHome_ShouldNotTraverseFurther

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -2186,6 +2187,7 @@ public Node createNode(String parentFolderNodeId, Node nodeInfo, Parameters para
 
         // check that requested parent node exists and it's type is a (sub-)type of folder
         NodeRef parentNodeRef = validateOrLookupNode(parentFolderNodeId);
```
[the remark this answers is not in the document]
SanjoyHyland2025: fixed this

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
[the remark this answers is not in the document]
SanjoyHyland2025: alraedy done this one

remote-api/src/test/java/org/alfresco/rest/api/impl/NodesImplSystemPathTest.java
```diff
@@ -0,0 +1,174 @@
...
+            nodesImpl.deleteNode(childRef.getId(), parameters);
+        }
+        catch (PermissionDeniedException e)
+        {
+            fail("Normal node under Company Home should not throw PermissionDeniedException");
+        }
+        catch (Exception e)
+        {
+            assertFalse("checkNotSystemPath should not block a normal node path",
+                    false);
```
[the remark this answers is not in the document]
SanjoyHyland2025: changed the asserting method

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
```diff
@@ -3767,14 +3774,14 @@
...
+        String prefix = PARAM_INCLUDE_PROPERTIES + "/";
 
         List<QName> result = new ArrayList<>(qnameStrList.size());
         for (String str : qnameStrList)
         {
-            if (str.startsWith(PREFIX))
+            if (str.startsWith(prefix))
             {
-                str = str.substring(PREFIX.length());
+                str = str.substring(prefix.length());
```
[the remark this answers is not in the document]
SanjoyHyland2025: done

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
SanjoyHyland2025: fixed this one

remote-api/src/main/java/org/alfresco/rest/api/impl/NodesImpl.java
SanjoyHyland2025: completed

jottley (approved):

PR #4080, opened 2026-05-13
ACS-11289 Property Extender Interceptor

gerardolenski:
More information about this change: https://hyland.atlassian.net/browse/ACS-11289

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
+++ b/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
@@ -45,5 +45,6 @@ public record CalculationResult(Map<QName, Serializable> additionalProperties)
     public CalculationResult
     {
         Objects.requireNonNull(additionalProperties);
+        additionalProperties = Collections.unmodifiableMap(additionalProperties);
     }
 }
--- a/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
+++ b/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
@@ -1,0 +1,51 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2026 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.repo.node.propertyextender;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+
+import org.alfresco.service.namespace.QName;
+
+/**
+ * The result of the property extender calculation.
+ *
+ * @param calculatedProperties
+ *            the additional properties calculated by the property extender, which will be added to the node together with the new properties
+ */
+public record CalculationResult(Map<QName, Serializable> calculatedProperties)
+{
+    public static final CalculationResult NO_OP = new CalculationResult(Collections.emptyMap());
+
+    public CalculationResult
+    {
+        calculatedProperties = Optional.ofNullable(calculatedProperties)
+                .map(Collections::unmodifiableMap)
+                .orElse(Collections.emptyMap());
+    }
+}
--- a/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
+++ b/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
@@ -28,7 +28,7 @@ package org.alfresco.repo.node.propertyextender;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
-import java.util.Optional;
+import java.util.Objects;
 
 import org.alfresco.service.namespace.QName;
 
@@ -44,8 +44,6 @@ public record CalculationResult(Map<QName, Serializable> additionalProperties)
 
     public CalculationResult
     {
-        additionalProperties = Optional.ofNullable(additionalProperties)
-                .map(Collections::unmodifiableMap)
-                .orElse(Collections.emptyMap());
+        Objects.requireNonNull(additionalProperties);
     }
 }
--- a/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
+++ b/repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
@@ -35,16 +35,16 @@ import org.alfresco.service.namespace.QName;
 /**
  * The result of the property extender calculation.
  *
- * @param calculatedProperties
- *            the additional properties calculated by the property extender, which will be added to the node together with the new properties
+ * @param additionalProperties
+ *            the additional properties calculated by the property extender, which will be applied to the node together with the original property changes.
  */
-public record CalculationResult(Map<QName, Serializable> calculatedProperties)
+public record CalculationResult(Map<QName, Serializable> additionalProperties)
 {
     public static final CalculationResult NO_OP = new CalculationResult(Collections.emptyMap());
 
     public CalculationResult
     {
-        calculatedProperties = Optional.ofNullable(calculatedProperties)
+        additionalProperties = Optional.ofNullable(additionalProperties)
                 .map(Collections::unmodifiableMap)
                 .orElse(Collections.emptyMap());
     }
--- a/repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyCalculator.java
+++ b/repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyCalculator.java
@@ -1,0 +1,115 @@
+/*
+ * #%L
+ * Alfresco Repository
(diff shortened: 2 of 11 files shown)


repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyCalculator.java
```diff
@@ -0,0 +1,115 @@
...
+        if (LOGGER.isTraceEnabled())
+        {
+            LOGGER.trace("Calculating additional properties for: {}", propertyChanges.keySet().stream()
+                    .map(QName::toPrefixString)
+                    .toList());
+        }
+        Map<QName, Serializable> calculated = extenders.stream()
+                .map(ext -> runCalculation(ext, propertyChanges))
+                .flatMap(result -> result.additionalProperties().entrySet().stream())
+                .collect(HashMap::new, (m, e) -> m.put(e.getKey(), e.getValue()), HashMap::putAll);
```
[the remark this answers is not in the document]
gerardolenski: N/A

repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyExtender.java
```diff
@@ -25,63 +25,29 @@
...
-     *            the context of the calculation, which contains the properties that are being added/updated on a node
-     * @return result containing calculated properties mapping. Empty map when no relevant properties were provided in the context.
+     *            the context of the calculation, which contains the property changes on a node.
+     * @return result containing calculated additional properties mapping. Empty map when no relevant properties were provided in the context.
      * @throws AlfrescoRuntimeException
      *             for expected calculation process failures.
      * @throws RuntimeException
      *             for any unexpected errors during the calculation process.
      */
     CalculationResult calculate(CalculationContext context);
```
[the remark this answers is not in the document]
gerardolenski: N/A

repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyExtenderInterceptor.java
```diff
@@ -66,60 +66,55 @@ public PropertyExtenderInterceptor(NodeService nodeService, PropertyExtendersHol
...
-    {
-        try
+        else if (methodName.equals("setProperty") && args.length == 3 &&
+                args[0] instanceof NodeRef nodeRef &&
+                args[1] instanceof QName propertyQName &&
+                args[2] instanceof Serializable propertyValue)
         {
-            return extender.calculate(new CalculationContext(newProperties));
+            var extendedProps = calculateProperties(extendersHolder.getExtenders(), Collections.singletonMap(propertyQName, propertyValue));
+            if (extendedProps.isExtended())
```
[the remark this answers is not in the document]
gerardolenski: Nice catch, fixed

repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
```diff
@@ -0,0 +1,49 @@
...
+ *            the additional properties calculated by the property extender, which will be applied to the node together with the original property changes.
+ */
+public record CalculationResult(Map<QName, Serializable> additionalProperties)
+{
+    public static final CalculationResult NO_OP = new CalculationResult(Collections.emptyMap());
+
+    public CalculationResult
+    {
+        Objects.requireNonNull(additionalProperties);
+    }
```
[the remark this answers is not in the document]
gerardolenski: fixed

repository/src/test/java/org/alfresco/repo/node/propertyextender/PropertyExtenderInterceptorTest.java
```diff
@@ -0,0 +1,438 @@
...
+        // when
+        sut.invoke(invocation);
+
+        // then
+        verify(nodeService).addProperties(NODE_REF, properties);
+    }
+
+    // --- helpers ---
+
+    private MethodInvocation mockInvocation(String methodName, Object[] args) throws NoSuchMethodException
```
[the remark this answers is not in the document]
gerardolenski: fixed

repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyExtendersHolder.java
```diff
@@ -25,25 +25,39 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(PropertyExtendersHolder.class);
+
+    private final Set<PropertyExtender> extenders = ConcurrentHashMap.newKeySet();
 
-    public List<PropertyExtender> getExtenders()
+    public Set<PropertyExtender> getExtenders()
     {
-        return unmodifiableList(extenders);
+        return unmodifiableSet(extenders);
     }
```
cezary-witkowski, line 50: I think I preferred version with List. I could imagine a scenario where the order of those extenders could change the outcome. With set we have lost control that we had with injections of beans. 

Imagine a scenario where the next Extender after yours will be needed to trigger yours. So this new Extender e.g. on adding some property will add another one which will be a key for your Extender. For that to work they would need to make sure they register their before yours and thats it. With Set this might be not possible.

repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyExtendersHolder.java
```diff
@@ -25,25 +25,39 @@
...
 
     public void registerExtender(PropertyExtender extender)
     {
-        extenders.add(extender);
+        if (extenders.add(extender))
```
cezary-witkowski, line 54: nice usage of the return boolean

cezary-witkowski (approved):

repository/src/main/java/org/alfresco/repo/node/propertyextender/PropertyExtendersHolder.java
gerardolenski: But right now it is not possible even in list:
- You have no controll of the registration order. If it will be needed, then contract of PropertyExtender should have eg. `priority` field. 
- One extender is not related to other, each calculates based on original input, not on input after previous calculation.

repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
```diff
@@ -0,0 +1,50 @@
...
+ *            the additional properties calculated by the property extender, which will be applied to the node together with the original property changes.
+ */
+public record CalculationResult(Map<QName, Serializable> additionalProperties)
+{
+    public static final CalculationResult NO_OP = new CalculationResult(Collections.emptyMap());
+
+    public CalculationResult
+    {
+        Objects.requireNonNull(additionalProperties);
+        additionalProperties = Collections.unmodifiableMap(additionalProperties);
```
arindam-roy-hld, line 48: Collections.unmodifiableMap() will be read-only, but since it is accepting params from outside, any other reference from outside can change the data using put(existing-key, new-value).
Can we use Map.copyOf()? It will discard the external reference and will be 100% immutable. (If no null key or value is expected to be stored)
(* Here and in CalculationContext class.)

repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * The result of the property extender calculation.
+ *
+ * @param additionalProperties
+ *            the additional properties calculated by the property extender, which will be applied to the node together with the original property changes.
+ */
+public record CalculationResult(Map<QName, Serializable> additionalProperties)
+{
+    public static final CalculationResult NO_OP = new CalculationResult(Collections.emptyMap());
```
arindam-roy-hld, line 43: Seems, NO_OP has been used in the Test classes. Should this be in this file? or, we can move this to the test classes.

repository/src/test/java/org/alfresco/repo/node/propertyextender/PropertyExtenderInterceptorTest.java
```diff
@@ -0,0 +1,491 @@
...
+        nodeService = mock(NodeService.class);
+        extendersHolder = spy(new PropertyExtendersHolder());
+        sut = new PropertyExtenderInterceptor(nodeService, extendersHolder);
+        captor = ArgumentCaptor.forClass(Map.class);
+    }
+
+    // --- addProperties ---
+
+    @Test
+    public void testAddProperties_noExtenders_shouldPassOriginalProperties() throws Throwable
```
arindam-roy-hld, line 92: Is "`throws Throwable`" required at all? May be we can add only "`throws NoSuchMethodException`" or "`throws Exception`" at maximum?

repository/src/test/java/org/alfresco/repo/node/propertyextender/PropertyExtenderInterceptorTest.java
```diff
@@ -0,0 +1,491 @@
...
+    {
+        // given
+        var invocation = mockInvocation("setProperty", NODE_REF, PROP_NAME, "test");
+
+        // when
+        sut.invoke(invocation);
+
+        // then
+        verify(nodeService).setProperty(NODE_REF, PROP_NAME, "test");
+        verify(nodeService, never()).addProperties(any(), any());
```
arindam-roy-hld, line 250: If we can add the class name here, it will be better.
`addProperties(any(), any());` -> `addProperties(any(NodeRef.class), anyMap());`

arindam-roy-hld (changes requested):

repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
gerardolenski: NO_OP is also used in ent-repo in production code. see CDLinker implementation

repository/src/test/java/org/alfresco/repo/node/propertyextender/PropertyExtenderInterceptorTest.java
gerardolenski: Yes - it is needed, see: `org.aopalliance.intercept.MethodInterceptor#invoke`

repository/src/main/java/org/alfresco/repo/node/propertyextender/CalculationResult.java
gerardolenski: My intention here is to ensure client won't modify it. It is hard to assume that the PropertyExtender will keep reference to it after calculation. If we assume that, then we should also consider that values of this maps should be deeply copied (which will be overkill ). For performance reasons (AOP interceptor) I rather want to avoid it.

arindam-roy-hld (approved):
LGTM - Good job

PR #4133, opened 2026-06-12
[MNT-25701] fix OOM in Share Edit Properties for large content

tathagta15:


repository/src/main/java/org/alfresco/repo/jscript/MetaDataExtractAction.java
```diff
@@ -162,6 +165,22 @@ public boolean isContentChanged(String itemId, FormData formData)
...
+            for (int i = 0; i < expected.length(); i++)
+            {
+                if (content.read() != expected.charAt(i))
+                {
+                    return false;
+                }
+            }
+            return content.read() == -1;
+        }
+    }
```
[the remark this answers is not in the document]
tathagta15: Keeping it minimal as null/missing-reader cases are already handled by the outer catch (same as the original code), and expected from String.valueOf(...) is never null. BufferedReader buffers internally, so the single-char loop is fine.

repository/src/main/java/org/alfresco/repo/jscript/MetaDataExtractAction.java
```diff
@@ -131,26 +136,23 @@ public ScriptAction create(boolean isContentChanged)
      */
     public boolean isContentChanged(String itemId, FormData formData)
     {
+        // Skip reading the existing content when the form has no content field (metadata-only save)
+        FormData.FieldData fieldData = formData.getFieldData("prop_cm_content");
+        if (fieldData == null || fieldData.getValue() == null)
+        {
+            return false;
+        }
```
kmagdziarz, line 144: What's benefit of moving that check outside try?

repository/src/main/java/org/alfresco/repo/jscript/MetaDataExtractAction.java
```diff
@@ -162,6 +164,22 @@ public boolean isContentChanged(String itemId, FormData formData)
...
+            for (int i = 0; i < expected.length(); i++)
+            {
+                if (content.read() != expected.charAt(i))
+                {
+                    return false;
+                }
+            }
+            return content.read() == -1;
+        }
+    }
```
kmagdziarz, line 181: So this method is only code that is run inside try on line #155 and have another try in it?

kmagdziarz (changes requested):

repository/src/main/java/org/alfresco/repo/jscript/MetaDataExtractAction.java
tathagta15: getFieldData is a map lookup returning null if absent, so keeping it within the try wont make much of sense
image

repository/src/main/java/org/alfresco/repo/jscript/MetaDataExtractAction.java
tathagta15: Outer try catches errors and returns false whereas the Inner try closes the stream

kmagdziarz (approved):

goleksyalfresco (approved):

PR #4137, opened 2026-06-15
ACS-11674 Elasticsearch Error Logging

jaredgrayhyland:


repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/contentmodelsync/ContentModelSynchronizer.java
```diff
@@ -195,14 +195,21 @@ boolean updateSettings(JSONObject settings, boolean preserveExisting) throws IOE
...
-        Response response = httpClientFactory.getElasticsearchClient()
-                .generic()
-                .execute(requests);
-
-        OpenResponse openResponse = indices.open(new OpenRequest.Builder().index(indexName)
-                .build());
+        try (Response response = httpClientFactory.getElasticsearchClient().generic().execute(requests))
+        {
+            OpenResponse openResponse = indices.open(new OpenRequest.Builder().index(indexName)
+                    .build());
```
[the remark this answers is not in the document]
kmagdziarz: Agree. Opening the index should be moved out of this try. This step should always happen.

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/contentmodelsync/ContentModelSynchronizer.java
```diff
@@ -195,14 +195,21 @@ boolean updateSettings(JSONObject settings, boolean preserveExisting) throws IOE
...
 
-        return response.getStatus() == 200 && openResponse.acknowledged();
+            boolean success = response.getStatus() == 200 && openResponse.acknowledged();
+            if (!success)
+            {
+                String rawBody = response.getBody().map(Body::bodyAsString).orElse("{}");
+                LOGGER.error("Failed to update analyser settings on index {}: status={} acknowledged={} reason={}", indexName,
+                        response.getStatus(), openResponse.acknowledged(), extractErrorReason(rawBody));
```
kmagdziarz: Please add commas to error message. Additionally, this generic setting upgrade method.

kmagdziarz (changes requested):

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/contentmodelsync/ContentModelSynchronizer.java
```diff
@@ -195,14 +195,55 @@ boolean updateSettings(JSONObject settings, boolean preserveExisting) throws IOE
...
+        finally
+        {
+            try
+            {
+                openResponse = indices.open(new OpenRequest.Builder().index(indexName)
+                        .build());
+            }
+            catch (IOException openException)
+            {
+                openFailure = openException;
```
adecatur72: Recommendation: add a log message inside the catch.  Maybe something like: LOGGER.error("Failed to reopen index {} after settings update; index may be left closed", indexName, openException);

A failed reopen is not good here.  The index is now closed and search/indexing will be broken until a restart, yet there's no log line that says so.

adecatur72: Also, in case the exception could be a RuntimeException (that the generic OpenSearch client could throw), you change this catch block to something  like this:
catch (Exception openException)
    {
        <log message from my earlier message>
        openFailure = (openException instanceof IOException ioe)
                ? ioe
                : new IOException("Failed to reopen index " + indexName + " after settings update", openException);
    }

adecatur72 (changes requested):
Looks good Jared.  I added a couple of comments.  I would also consider adding a little testing coverage in ContentModelSynchronizerTest.java for these updates.  You probably just need a couple of unit tests.

adecatur72 (approved):

kmagdziarz (approved):

PR #4225, opened 2026-07-24
ACS-11733 SDK Out-of-Process Java REST API - 'listActivitiesForPerson' Method does not Deserialize json.

b-ansari:


remote-api/src/main/resources/alfresco/public-rest-context.xml
```diff
@@ -342,7 +342,11 @@
         <property name="registry" ref="activitySummaryParser" />
         <property name="eventTypes">
             <list>
+                <value>org.alfresco.discussions.post-created</value>
```
gerardolenski, line 345: If I'm not mistaken, editing the eventTypes list affects every consumer of /people/{id}/activities — the Java SDK, the JS-API, and raw HTTP clients. So stripping params from more discussion types would change the payload for all of them, not just your SDK.

gerardolenski: all right, it is not the contract change. In RESt API explorer we have:

Image

So looks good

gerardolenski (dismissed):

cezary-witkowski (dismissed):
The analysis and solution look good to me. Good work. Please consider adding at least one test.

cezary-witkowski (approved):
Thanks for adding test. LGTM

PR #4229, opened 2026-07-27
 Implement new download content audit event type and Custom flag for Configuration

kushal-banik-hyland:
This pull request introduces a new `OnContentDownloadPolicy` to the Alfresco repository, enabling the system to audit and handle content downloads distinctly from content reads. The changes span the content service, auditing, and event generation components, ensuring that downloads can be tracked separately and optionally treated as read events, based on configuration.

**Key changes include:**

### Content Service Enhancements

* Added a new `OnContentDownloadPolicy` interface to `ContentServicePolicies`, and updated `ContentServiceImpl` to fire this policy when content is downloaded as an attachment. The method `getReader` now accepts an `attach` parameter and triggers the appropriate policy (`OnContentReadPolicy` or `OnContentDownloadPolicy`) depending on the context and configuration. [[1]](diffhunk://#diff-4a64337e5b8ae71bb153b0443d3622f0541b6d79566bf1a9d5517863f520bc17R96-R108) [[2]](diffhunk://#diff-5949f37e586852c704ddc6e10ed480c214234e455ca1050ee6c7e655be214c53L392-R412) [[3]](diffhunk://#diff-5949f37e586852c704ddc6e10ed480c214234e455ca1050ee6c7e655be214c53L421-R451) [[4]](diffhunk://#diff-e293cc891a39ecbb23afb336aa362e64efa6a5994d00f80f79a1c1d1fcab9d3aL278-R285)

* Introduced the `auditModelRegistry` dependency in `ContentServiceImpl`, with support for a new configuration property that determines whether content downloads should also be treated as reads for auditing purposes.

### Auditing and Policy Integration

* Updated the `AccessAuditor` and `NodeChange` classes to implement and handle the new `OnContentDownloadPolicy`. This includes recording download actions in the audit log and distinguishing them from read actions.
### Event Generation

* Registered the new `OnContentDownloadPolicy` with the event generation system to allow event listeners to respond to download actions, similar to other content-related events.
### Configuration

* Added a new configuration property `.contentDownloadsAsRead` to control whether downloads should also trigger read audit events, providing flexibility for different auditing requirements
---

These changes provide more granular auditing and policy enforcement for content downloads, supporting both operational and compliance needs.

Example of the Audit Entry when we have set flag `audit.enable.contentDownloadsAsRead` as true
`   {
        "_index": "audit-event-index",
        "_id": "812c271c-8fc5-48e8-a670-9c39aa8895b8",
        "_score": null,
        "_source": {
          "type": "org.alfresco.event.audit.entry.Created",
          "id": "812c271c-8fc5-48e8-a670-9c39aa8895b8",
          "time": "2026-07-27T11:51:45.628636468Z",
          "data": {
            "entryType": "alfresco-access",
            "username": "admin",
            "entryDBId": 1044,
            "entryList": [
              {
                "key": "transaction.sub-actions",
                "value": "readContent downloadContent"
              },
              {
                "key": "transaction.path",
                "value": "/app:company_home/app:guest_home/cm:botPic.PNG"
              },
              {
                "key": "transaction.action",
                "value": "READ"
              },
              {
                "key": "transaction.user",
                "value": "admin"
              },
              {
                "key": "transaction.type",
                "value": "cm:content"
              }
            ]
          }
        }
      },
      {
        "_index": "audit-event-index",
        "_id": "776a5948-f79d-4388-89e0-793b58f5d1f0",
        "_score": null,
        "_source": {
          "type": "org.alfresco.event.audit.entry.Created",
          "id": "776a5948-f79d-4388-89e0-793b58f5d1f0",
          "time": "2026-07-27T11:51:45.631933338Z",
          "data": {
            "entryType": "alfresco-access",
            "username": "admin",
            "entryDBId": 1045,
            "entryList": [
              {
                "key": "transaction.sub-actions",
                "value": "readContent downloadContent"
              },
              {
                "key": "transaction.path",
                "value": "/app:company_home/app:guest_home/cm:botPic.PNG"
              },
              {
                "key": "transaction.action",
                "value": "DOWNLOAD"
              },
              {
                "key": "transaction.user",
                "value": "admin"
              },
              {
                "key": "transaction.type",
                "value": "cm:content"
              }
            ]
          }
        }
  }`
and when we are just having download with flag `audit.enable.contentDownloadsAsReads=false`

`   {
        "_index": "audit-event-index",
        "_id": "776a5948-f79d-4388-89e0-793b58f5d1f0",
        "_score": null,
        "_source": {
          "type": "org.alfresco.event.audit.entry.Created",
          "id": "776a5948-f79d-4388-89e0-793b58f5d1f0",
          "time": "2026-07-27T11:51:45.631933338Z",
          "data": {
            "entryType": "alfresco-access",
            "username": "admin",
            "entryDBId": 1045,
            "entryList": [
              {
                "key": "transaction.sub-actions",
                "value": "downloadContent"
              },
              {
                "key": "transaction.path",
                "value": "/app:company_home/app:guest_home/cm:botPic.PNG"
              },
              {
                "key": "transaction.action",
                "value": "DOWNLOAD"
              },
              {
                "key": "transaction.user",
                "value": "admin"
              },
              {
                "key": "transaction.type",
                "value": "cm:content"
              }
            ]
          }
        }`

### DIFF
--- a/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
+++ b/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
@@ -182,7 +182,7 @@ public class AccessAuditor implements InitializingBean,
     private static final String TRANSACTION = "transaction";
     private static final String AUDIT_SUB_ACTIONS = "audit.alfresco-access.sub-actions.enabled";
     private static final String ACTION = "action";
-    private static final String READ_ACTION =  "read";
+    private static final String READ_ACTION = "read";
 
     private Properties properties;
     private PolicyComponent policyComponent;
@@ -588,26 +588,32 @@ public class AccessAuditor implements InitializingBean,
     /**
      * Listen for commit to audit gathered audit activity for the current user transaction.
      */
-    private class AccessTransactionListener extends TransactionListenerAdapter {
+    private class AccessTransactionListener extends TransactionListenerAdapter
+    {
         @Override
-        public void afterCommit() {
+        public void afterCommit()
+        {
             // Note: auditComponent.recordAuditValues(...) creates a transaction to record
             // audit messages, so there is no need to create our own. dod5015 still
             // does (not sure why).
             final Map<NodeRef, NodeChange> changedNodes = TransactionalResourceHelper.getMap(this);
-            for (Map.Entry<NodeRef, NodeChange> entry : changedNodes.entrySet()) {
+            for (Map.Entry<NodeRef, NodeChange> entry : changedNodes.entrySet())
+            {
                 NodeChange nodeChange = entry.getValue();
-                if (!nodeChange.isTemporaryNode()) {
+                if (!nodeChange.isTemporaryNode())
+                {
                     publishAuditForTheTransaction(nodeChange);
                 }
             }
         }
 
-        private void publishAuditForTheTransaction(NodeChange nodeChange) {
+        private void publishAuditForTheTransaction(NodeChange nodeChange)
+        {
             // When both READ and DOWNLOAD sub-actions are present (i.e., the
             // contentDownloadsAsRead flag is enabled), emit two separate audit entries.
             Map<String, Serializable> auditMap = nodeChange.getAuditData(false);
-            if (nodeChange.hasReadAndDownloadActions()) {
+            if (nodeChange.hasReadAndDownloadActions())
+            {
                 Map<String, Serializable> readAuditMap = new HashMap<>(auditMap); // This Must Come will Download as higher precedence
                 readAuditMap.put(ACTION, READ_ACTION);
                 recordAuditValues(TRANSACTION, readAuditMap);
--- a/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
+++ b/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
@@ -32,6 +32,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 
+import org.alfresco.repo.content.ContentServicePolicies;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.InitializingBean;
@@ -43,6 +44,7 @@ import org.alfresco.repo.coci.CheckOutCheckInServicePolicies.OnCancelCheckOut;
 import org.alfresco.repo.coci.CheckOutCheckInServicePolicies.OnCheckIn;
 import org.alfresco.repo.coci.CheckOutCheckInServicePolicies.OnCheckOut;
 import org.alfresco.repo.content.ContentServicePolicies.OnContentReadPolicy;
+import org.alfresco.repo.content.ContentServicePolicies.OnContentDownloadPolicy;
 import org.alfresco.repo.content.ContentServicePolicies.OnContentUpdatePolicy;
 import org.alfresco.repo.copy.CopyServicePolicies.OnCopyCompletePolicy;
 import org.alfresco.repo.node.NodeServicePolicies.BeforeDeleteNodePolicy;
@@ -175,7 +177,7 @@ public class AccessAuditor implements InitializingBean,
 
         OnCopyCompletePolicy,
 
-        OnCheckOut, OnCheckIn, OnCancelCheckOut
+        OnCheckOut, OnCheckIn, OnCancelCheckOut, ContentServicePolicies.OnContentDownloadPolicy
 {
     /** Logger */
     private static Log logger = LogFactory.getLog(AccessAuditor.class);
@@ -265,6 +267,7 @@ public class AccessAuditor implements InitializingBean,
         policyComponent.bindClassBehaviour(OnRemoveAspectPolicy.QNAME, this, new JavaBehaviour(this, "onRemoveAspect"));
 
         policyComponent.bindClassBehaviour(OnContentUpdatePolicy.QNAME, this, new JavaBehaviour(this, "onContentUpdate"));
+        policyComponent.bindClassBehaviour(OnContentDownloadPolicy.QNAME, this, new JavaBehaviour(this, "onContentDownload"));
         policyComponent.bindClassBehaviour(OnContentReadPolicy.QNAME, this, new JavaBehaviour(this, "onContentRead"));
 
         policyComponent.bindClassBehaviour(OnCreateVersionPolicy.QNAME, ContentModel.TYPE_CONTENT, new JavaBehaviour(this, "onCreateVersion"));
@@ -568,6 +571,17 @@ public class AccessAuditor implements InitializingBean,
         else
         {
             sb.append(value);
+        }
+    }
+
+    /**
+     * @param nodeRef the node reference
+     */
+    @Override
+    public void onContentDownload(NodeRef nodeRef) {
+        if (auditEnabled())
+        {
+            getNodeChange(nodeRef).onContentDownload(nodeRef);
         }
     }
 
--- a/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
+++ b/repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
@@ -598,29 +598,33 @@ public class AccessAuditor implements InitializingBean,
             // Note: auditComponent.recordAuditValues(...) creates a transaction to record
             // audit messages, so there is no need to create our own. dod5015 still
             // does (not sure why).
-
             final Map<NodeRef, NodeChange> changedNodes = TransactionalResourceHelper.getMap(this);
             for (Map.Entry<NodeRef, NodeChange> entry : changedNodes.entrySet())
             {
                 NodeChange nodeChange = entry.getValue();
                 if (!nodeChange.isTemporaryNode())
                 {
-                    // When both READ and DOWNLOAD sub-actions are present (i.e., the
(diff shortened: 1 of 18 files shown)


repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
```diff
@@ -582,17 +598,34 @@ public void afterCommit()
...
+                    publishAuditForTheTransaction(nodeChange);
                 }
             }
         }
+
+        private void publishAuditForTheTransaction(NodeChange nodeChange)
+        {
+            // When both READ and DOWNLOAD sub-actions are present (i.e., the
+            // contentDownloadsAsRead flag is enabled), emit two separate audit entries.
+            if (nodeChange.hasReadAndDownloadActions())
```
kushal-banik-hyland, line 620: Basically this is the solution if we need to have the extra audit entry for download. Normally there is a precedence rule by which we only get single audit action for any transaction , for any specific node. If we want to make sure that both actions are triggered then we need to throw two specific entries specifically.
Here we get `read` and `download` both as sub-action when the attachment = `true` and the introduced flag `audit.enabled.contentDownloadsAsRead` set as true.

kushal-banik-hyland: Also if you find any other more optimized method feel free to suggest!

repository/src/main/java/org/alfresco/repo/model/filefolder/FileFolderServiceImpl.java
```diff
@@ -1632,7 +1632,7 @@ public ContentReader getReader(NodeRef nodeRef)
         {
             throw new InvalidTypeException("Unable to get a content reader for a folder: " + fileInfo);
         }
-        return contentService.getReader(nodeRef, ContentModel.PROP_CONTENT);
+        return contentService.getReader(nodeRef, ContentModel.PROP_CONTENT, true); // explicitly making sure all the CMIS and WebDav has this method selected as DOWNLOAD
```
damianujma: Making every caller `getReader(NodeRef nodeRef)` a "download" is a too broad. For example `RecordsManagementSearchServiceImpl.getSavedSearches/getSavedSearch` call this method - to read a saved-search config JSON - not a download. 
The actual CMIS download (`CMISConnector.getContentStream`) still uses the 2-arg reader.

damianujma (changes requested):
See potential improvements: https://github.com/Alfresco/alfresco-community-repo/compare/master...feature/ACS-12294-du-Implement-New-downloadContent-Audit-Event-Type

repository/src/main/java/org/alfresco/repo/model/filefolder/FileFolderServiceImpl.java
kushal-banik-hyland: Changed

repository/src/main/java/org/alfresco/repo/audit/access/NodeChange.java
```diff
@@ -186,6 +188,10 @@ else if (subActions.contains(CREATE_NODE))
         {
             action = "CREATE";
         }
+        else if (subActions.contains(DOWNLOAD_CONTENT))
+        {
+            action = "DOWNLOAD";
+        }
```
jakubkochman, line 195: It seems that `READ` has a special treatment when using `runAs`. With this change it's possible to loose that:

```
    @BeforeClass
    public static void setUpAuthenticationUtil() throws Exception
    {
        // AuthenticationUtil.runAs()/setRunAsUser() need this static init, normally done by Spring.
        new AuthenticationUtil().afterPropertiesSet();
    }



@Test
    public final void testOnContentReadAndDownload_runAsUserAttributionLostForSplitReadEntry()
    {
        AuthenticationUtil.setFullyAuthenticatedUser("bob");
        try
        {
            AuthenticationUtil.runAs(() -> {
                nodeChange.onContentRead(content1);
                nodeChange.onContentDownload(content1);
                return null;
            }, "alice");

            assertTrue("Precondition: both sub-actions must be present", nodeChange.hasReadAndDownloadActions());

            Map<String, Serializable> auditMap = nodeChange.getAuditData(false);

            assertEquals("DOWNLOAD", auditMap.get("action"));
            assertEquals("MNT-8810: the actual reader should be attributed, not the ambient user",
                    "alice", auditMap.get("user"));
        }
        finally
        {
            AuthenticationUtil.clearCurrentSecurityContext();
        }
    }
```

In `getDerivedAction()` download is resolved first so it does not enter the read block to preserve the runAs.

repository/src/main/java/org/alfresco/repo/events/EventGenerationBehaviours.java
```diff
@@ -149,6 +150,8 @@ public void init()
             return;
         }
 
+        bindClassPolicy(ContentServicePolicies.OnContentDownloadPolicy.QNAME, NodeContentPutEvent.EVENT_TYPE);
```
damianujma: Why download policy is gated on `NodeContentPutEvent`?

repository/src/main/java/org/alfresco/repo/audit/access/NodeChange.java
damianujma, line 202: Shouldn't we set `keepRunAsUser = true`, see `else if` block for read content.

repository/src/main/java/org/alfresco/repo/events/EventGenerationBehaviours.java
damianujma: Why `NodeContentPutEvent.EVENT_TYPE`?

repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
```diff
@@ -582,17 +601,30 @@ public void afterCommit()
...
+        private void publishAuditForTheTransaction(NodeChange nodeChange)
+        {
+            // When both READ and DOWNLOAD sub-actions are present (i.e., the
+            // contentDownloadsAsRead flag is enabled), emit two separate audit entries.
+            Map<String, Serializable> auditMap = nodeChange.getAuditData(false);
+            if (nodeChange.hasReadAndDownloadActions())
+            {
+                Map<String, Serializable> readAuditMap = new HashMap<>(auditMap); // This Must Come Download as higher precedence
+                readAuditMap.put(ACTION, READ_ACTION);
+                recordAuditValues(TRANSACTION, readAuditMap);
```
damianujma, line 624: For READ action I can see read and download in sub-actions. Is it valid?

repository/src/main/java/org/alfresco/repo/audit/access/NodeChange.java
kushal-banik-hyland: Good Catch. Actually there is a flag named `keepRunAsUser`. That is primarily keeps the user intact after calculating the derived action (which actual user was fetched during the action invocation). Making this default false wipe out the user and again fetch this after hand which is wrong. So need to make the flag `keepRunAsUser` true in download also.
Also added this test. Thanks!

repository/src/main/java/org/alfresco/repo/audit/access/AccessAuditor.java
kushal-banik-hyland: Basically both are in same transaction. Here any download comes with read so in this sub-actions both will appear for read and download.

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
```diff
@@ -389,11 +397,17 @@ public ContentReader getRawReader(String contentUrl)
...
     public ContentReader getReader(NodeRef nodeRef, QName propertyQName)
     {
-        return getReader(nodeRef, propertyQName, true);
+        return getReaderImpl(nodeRef, propertyQName, true, false);
+    }
+
+    @Override
+    public ContentReader getReader(NodeRef nodeRef, QName propertyQName, boolean attachment)
+    {
+        return getReaderImpl(nodeRef, propertyQName, true, attachment);
```
SatyamSah5: Shouldn't we keep this to something like
`return getReaderImpl(nodeRef, propertyQName, isDownloadAsRead, attachment);`

So that based on the new parameter `downloadAsRead` we would decide to trigger read policy as well for download scenarios, instead of always keeping it true.

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
```diff
@@ -389,11 +397,17 @@ public ContentReader getRawReader(String contentUrl)
...
     }
 
     @SuppressWarnings("unchecked")
-    private ContentReader getReader(NodeRef nodeRef, QName propertyQName, boolean fireContentReadPolicy)
+    private ContentReader getReaderImpl(NodeRef nodeRef, QName propertyQName, boolean fireContentReadPolicy, boolean fireContentDownloadPolicy)
```
SatyamSah5, line 409: is there a need of changing core method's name, pls check all places where it was used.

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
kushal-banik-hyland: Ya its a good suggestion! Can unify the logic!

repository/src/main/java/org/alfresco/repo/content/ContentServiceImpl.java
kushal-banik-hyland: Basically this method is new. Made sure the actual `getReader(...)` line 404 has the same contract. The new one  basically an deep level logical seperation

PR #4230, opened 2026-07-27
Fix/review and restrict global secrets

joyjit-hyland:
We reverted this PR because there was some test failure after merging this PR.
Even though the failure was due to S3 bucket missing, not related to the changes of this PR we decided to test the changes further before merging. 

I have run the branch couple of times in CI, everything seems to be fine.


gionn (changes requested):
fix branch conflicts

gionn (approved):

PR #4272, opened 2026-08-11
ACS-12472 Export on AGS record search should additionally support csv file

debjit-chattopadhyay-hyland:
https://hyland.atlassian.net/browse/MNT-23479

This pull request adds support for embedding a CSV file of search results into exported ACP or ZIP archives in the Alfresco Records Management module. The new functionality allows users to include the exact tabular data they see on screen as an additional file in the export, improving traceability and usability. The implementation includes new logic for handling the optional `items` parameter in export requests, a utility class for generating CSV files from tabular data, and comprehensive tests to ensure correct behavior.

**Export Functionality Enhancements**
* Added support for an optional `items` parameter in the export API, allowing a displayed search-result table (with `headers` and `rows`) to be embedded as a CSV file inside the exported archive (`ExportPost.java`, `rm-community-repo/config/alfresco/templates/webscripts/org/alfresco/rma/export.post.desc.xml`). [[1]](diffhunk://#diff-b01fc9e717f6631a49ae7262647e798806389de7abaf145e8141c546ba4eafbaL7-R11) [[2]](diffhunk://#diff-96f09a0da34131aefba199fcda93ca82d7b1511bb4a7c8028f9f6d938ebdd6a5R76-R84) [[3]](diffhunk://#diff-96f09a0da34131aefba199fcda93ca82d7b1511bb4a7c8028f9f6d938ebdd6a5R114) [[4]](diffhunk://#diff-96f09a0da34131aefba199fcda93ca82d7b1511bb4a7c8028f9f6d938ebdd6a5R127-R133) [[5]](diffhunk://#diff-96f09a0da34131aefba199fcda93ca82d7b1511bb4a7c8028f9f6d938ebdd6a5R145-R150) [[6]](diffhunk://#diff-96f09a0da34131aefba199fcda93ca82d7b1511bb4a7c8028f9f6d938ebdd6a5R183-R188)

**CSV Generation Utility**
* Introduced the `SearchResultsCSVWriter` class to build a CSV file from the supplied tabular data, ensuring values are sanitized and formatted correctly for spreadsheet applications (`SearchResultsCSVWriter.java`).

**Archive Manipulation**
* Implemented methods to add the generated CSV file as a new entry at the root of an existing ZIP-based archive, using Java NIO filesystem APIs for safe file operations (`ExportPost.java`).

**Testing**
* Added a new test case to verify that exporting with the `items` parameter correctly embeds a search-results CSV file in the archive, and a helper method to check for the presence of the CSV entry (`RmRestApiTest.java`).

**Documentation**
* Updated the API documentation to describe the new `items` parameter and the structure of the expected search-results table (`export.post.desc.xml`).


amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
```diff
@@ -0,0 +1,237 @@
...
+            {
+                LOG.debug("Created temporary CSV file: " + csvFile.getAbsolutePath());
+            }
+
+            return csvFile;
+        }
+        catch (JSONException je)
+        {
+            deleteTempFile(csvFile);
+            throw new WebScriptException(Status.STATUS_BAD_REQUEST, "Could not parse CSV data from request body.", je);
```
jottley, line 107: This is approach will work.  If this was a common approach across the code (ie two times in a single class is not significant enough to require a change) at that point I would suggest something like this

```java
catch (JSONException je) {
    throw buildException(csvFile, Status.STATUS_BAD_REQUEST, "Could not parse CSV data from request body.", je);
}
catch (IOException ioe) {
    throw buildException(csvFile, Status.STATUS_INTERNAL_SERVER_ERROR, "Failed to create CSV file.", ioe);
}

private WebScriptException buildException(File csvFile, int status, String message, Exception cause) {
    deleteTempFile(csvFile);
    return new WebScriptException(status, message, cause);
}
```

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Deletes the given temporary file if it exists, ignoring any failure. Prevents partially written CSV files from accumulating on disk when the CSV cannot be created successfully.
+     *
+     * @param file
+     *            the file to delete
+     */
+    private void deleteTempFile(File file)
+    {
+        if (file != null && file.exists() && !file.delete() && LOG.isDebugEnabled())
```
jottley, line 124: Can we change the logging to use `warn`?  If debug is off they will never see that this is failing. If it fails we can slowly leak undeleted files and the customer would never know.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/ExportPost.java
jottley, line 197: There is this single JSONException handling that currently exists...but now we have two possible cases where the exception can be thrown: one from the body and one from the parameters being passed.  It would be better to have a separate handling at the point of the new code so that any failures specific to parsing the parameter can be  called out at that time.

amps/ags/rm-community/rm-community-repo/config/alfresco/templates/webscripts/org/alfresco/rma/export.post.desc.xml
```diff
@@ -4,7 +4,11 @@
   Exports one or more nodes within an ACP file and streams the archive back to the client.<br/>
   The body of the post should be in the form<br/>
   {<br/>
-  &nbsp;&nbsp;&nbsp;"nodeRefs" : array of nodeRefs to export<br/>
+  &nbsp;&nbsp;&nbsp;"nodeRefs" : array of nodeRefs to export,<br/>
+  &nbsp;&nbsp;&nbsp;"items" : { (optional) displayed search-result table embedded as a timestamped AGS_Search_Results_&lt;timestamp&gt;.csv inside the archive<br/>
```
jottley: Do we know what performance will be like if there is a large amount of data being passed? Any concerns with not just parsing, but writing?

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
```diff
@@ -0,0 +1,237 @@
...
+     *
+     * @param items
+     *            the object holding the tabular data
+     * @return the headers array
+     */
+    private JSONArray getHeaders(JSONObject items)
+    {
+        if (items == null)
+        {
+            throw new WebScriptException(Status.STATUS_BAD_REQUEST, "Mandatory 'items' parameter was not provided in request body");
```
jottley: Understanding that when this code executes it is expected (requiring) the `items` parameter. But in the overall execution it is thought of as an optional parameter.  Can we clean up the language so that we don't confuse the two. I think just removing `Mandatory` will work. Or changing it to `Expected`? This would be the same for the use of `Mandatory` in the messages below.

jottley (changes requested):

amps/ags/rm-community/rm-community-repo/config/alfresco/templates/webscripts/org/alfresco/rma/export.post.desc.xml
debjit-chattopadhyay-hyland: The `items` payload is the rendered results grid, which is capped by `rm.recordSearch.maxItems` property (default 500). https://github.com/Alfresco/alfresco-community-repo/blob/6839fab6038e8a347ff85815bd1afc9b157f0bf9/amps/ags/rm-community/rm-community-repo/config/alfresco/module/org_alfresco_module_rm/alfresco-global.properties
 
Parsing reads the items JSON once into memory via org.json. At these volumes it's negligible; the parsed structure is the only part held in memory.

Writing is streamed, not buffered: each row is sanitised and written one at a time to a temp file through a buffered CSVPrinter, then the finished CSV is injected into the ACP zip. So write-side heap stays proportional to a single row and scales linearly — the CSV is never assembled in memory as a whole.

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/ExportPost.java
debjit-chattopadhyay-hyland: done

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
debjit-chattopadhyay-hyland: done

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
debjit-chattopadhyay-hyland: noted

amps/ags/rm-community/rm-community-repo/source/java/org/alfresco/module/org_alfresco_module_rm/script/SearchResultsCSVWriter.java
debjit-chattopadhyay-hyland: done

jottley (approved):

PR #4278, opened 2026-08-12
ACS-11473 Switch nodesSharedCache to invalidating and fix stale cache in rendition

SatyamSah5:
When `nodesSharedCache` was switched from `invalidating `to `fully distributed`, one of the customers observed search performance degrading when in cluster environment. The reason being that with more the size of ACS nodes, more the number of network calls made to access the shared cache in cluster. For ex: A cluster with 10 nodes, almost 90 % of cache calls( for `nodesSharedCache` ) would be network calls which leads to increased latency and hence search performance degradation.

This PR reverts the `nodesSharedCache` back to `invalidating` and fixes the issue raised in `ACS-6670`
For more details see `ACS-11473` and `ACS-6670`

The core change is `nodeDAO.setCheckNodeConsistency();` which forces db check for the ongoing transaction, hence bypassing the cache.


repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -350,6 +357,21 @@ private void requestAsyncTransformOrRendition(NodeRef sourceNodeRef, RenderOrTra
...
+                            // ContentData (size=0) from initial node creation because cache invalidation from the
+                            // uploading node hasn't propagated yet. Bypass the shared cache to read directly from DB.
+                            nodeDAO.setCheckNodeConsistency();
+                            contentDataForTransform = DefaultTypeConverter.INSTANCE.convert(
+                                    ContentData.class, nodeService.getProperty(sourceNodeRef, PROP_CONTENT));
+                            if (contentDataForTransform != null && contentDataForTransform.getSize() > 0)
+                            {
+                                sourceContentHashCode = getSourceContentHashCode(sourceNodeRef);
+                            }
+                        }
```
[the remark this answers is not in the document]
SatyamSah5: Added `renderOrTransform.throwIllegalStateExceptionIfAlreadyDone(sourceContentHashCode);`

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -112,6 +113,7 @@
 
     private TransactionService transactionService;
     private NodeService nodeService;
+    private NodeDAO nodeDAO;
```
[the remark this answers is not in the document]
SatyamSah5: Will move it to top of the class.

mousumi-podder (dismissed):

kushal-banik-hyland:
Can you give the reports here which you found during benchmarking or is there any confluence page regarding this?

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -96,22 +97,9 @@ public class RenditionService2Impl implements RenditionService2, InitializingBea
 
     private static Log logger = LogFactory.getLog(RenditionService2Impl.class);
 
-    // As Async transforms and renditions are so similar, this class provides a way to provide the code that is different.
-    private abstract static class RenderOrTransformCallBack
```
kushal-banik-hyland, line 100: Is this is the same class in down below?

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -350,6 +358,22 @@ private void requestAsyncTransformOrRendition(NodeRef sourceNodeRef, RenderOrTra
...
                     {
+                        ContentData contentDataForTransform = DefaultTypeConverter.INSTANCE.convert(
+                                ContentData.class, nodeService.getProperty(sourceNodeRef, PROP_CONTENT));
+                        if (contentDataForTransform != null && contentDataForTransform.getSize() == 0)
+                        {
+                            // In a cluster with invalidating cache, the shared cache may hold stale
+                            // ContentData (size=0) from initial node creation because cache invalidation from the
+                            // uploading node hasn't propagated yet. Bypass the shared cache to read directly from DB.
+                            nodeDAO.setCheckNodeConsistency();
+                            contentDataForTransform = DefaultTypeConverter.INSTANCE.convert(
```
kushal-banik-hyland: If in case for some time the invalidation is propagated or the config is changed this transformation block will not executed as there will be some `contentDataForTransformation` whose size will be greater than zero

repository/src/main/resources/alfresco/caches.properties
```diff
@@ -164,7 +164,7 @@ cache.node.nodesSharedCache.tx.statsEnabled=${caches.tx.statsEnabled}
 cache.node.nodesSharedCache.maxItems=250000
 cache.node.nodesSharedCache.timeToLiveSeconds=300
 cache.node.nodesSharedCache.maxIdleSeconds=0
-cache.node.nodesSharedCache.cluster.type=fully-distributed
+cache.node.nodesSharedCache.cluster.type=invalidating
```
kushal-banik-hyland, line 167: Getting the intension to avoid fully-distributed during reading from the cluster as this distributed shared keys may present in different nodes causes the roundtrip network delay. Here in our benchmark setup we have seen the invalidating gives better result as the stale data are no longer present and those are directly fetched from the db. Means `time To Fetch From Cache Shard` > `Time To Fetch From DB`. But in production scenario if there is some more bottleneck for fetching this from Db (Considering the network or read write cycles per second). Which can't be determined without setup and also I guess it can be vary from env to env. 
Composition of making key stale makes the performance of cache worse or the stats of cache hits reduced. Maybe near cache can be a good option if we are not trying to make db more engaging.
The operations in any node for `nodeSharedCache` will invalidate the keys always for any change. For high concurrent operations db can come to pressure as mostly the cache layer can be will not handle the calls.
Maybe I am missing some real condition or something but found out nearcache is there to solve this exact problem!

SatyamSah5: I actually tested with near cache with fully distributed mode and though it gives some performance improvements w.r.t using only fully-distributed mode, invalidating was far better in terms of performance. I agree that it was tested on my local system and might vary env to env.

Also, near cache is similar to invalidating as in case of write/update, remote messages needs to be sent to other nodes local cache(nearCache). But with near cache and fully distributed mode, the miss path becomes `local near-cache → IMap (network hop) → DB`. The `IMap` layer adds a network round-trip that provides no value for this specific cache — after an invalidation, the `IMap` partition on the remote node holds the same stale data (or nothing at all), so we end up at DB anyway, just with extra latency. With `InvalidatingCache`, the miss path is simply: `local cache → DB` — one fewer hop.

With near-cache + IMap(fully-distributed), we need to tune and maintain TWO cache sizes — the near-cache capacity (per node) AND the IMap capacity (distributed). Eviction policies, memory pressure, and GC impact need to be considered at both layers. With InvalidatingCache, there's a single local cache to size. Given that `nodesSharedCache` is the highly critical cache in Alfresco, A second capacity layer would make cache management more complex.

IMO, `nodesSharedCache` is accessed on virtually every operation, every property read, every permission check, every search result resolution. It's overwhelmingly read-heavy (reads vastly outnumber writes).

Also, in this scenario, the fix does NOT add DB calls on every read. The normal path remains: `local cache → return (no DB)`. The DB fallback only triggers when stale data is explicitly detected in the rendition callback — specifically when `contentData.getSize() == 0`. Only then does` nodeDAO.setCheckNodeConsistency() `bypass the shared cache for that single transaction to re-read from DB. Also, this stale cache issue was found internally when running [dtas tests](https://github.com/Alfresco/alfresco-deployment-test-automation-scripts/blob/master/tests/scenarios/test_transformation.py)

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
SatyamSah5: yes

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
SatyamSah5: The bypass only activates in the narrow cross-node race window where the shared cache still holds stale ContentData with size=0.

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -350,6 +358,22 @@ private void requestAsyncTransformOrRendition(NodeRef sourceNodeRef, RenderOrTra
 
                     if (sourceContentHashCode != SOURCE_HAS_NO_CONTENT)
                     {
+                        ContentData contentDataForTransform = DefaultTypeConverter.INSTANCE.convert(
+                                ContentData.class, nodeService.getProperty(sourceNodeRef, PROP_CONTENT));
```
damianujma: Can we extract a `getSourceContentData(NodeRef)` private helper method + an overload `getSourceContentHashCode(ContentData)` that accepts pre-fetched data?

```java
private ContentData getSourceContentData(NodeRef sourceNodeRef) {
    return DefaultTypeConverter.INSTANCE.convert(ContentData.class,
        nodeService.getProperty(sourceNodeRef, PROP_CONTENT));
}

private int getSourceContentHashCode(ContentData contentData) {
    // logic moved here
}

private int getSourceContentHashCode(NodeRef sourceNodeRef) {
    return getSourceContentHashCode(getSourceContentData(sourceNodeRef));
}
```

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -48,6 +48,7 @@
 import org.alfresco.repo.content.ContentServicePolicies;
```
damianujma, line 48: Please update the license header.

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
SatyamSah5: Nice suggestion 👍 , updated

repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
SatyamSah5: Updated

mousumi-podder (approved):

repository/src/main/resources/alfresco/caches.properties
kushal-banik-hyland: That's a valid path and I believe for this reason the Invalidating mode was introduced manually beside the `nearCache` as it also fetches to shad based node in case of miss and it goes to db. Then we can directly go to db from this node. Making `nodeSharedCache` to invalidating could solve that performance problem. Maybe if this change comes with another kind of problem we may use `ReplicatedMap` kind of implementation in future for async replication. It does the same like staling but adds the updated data. For now I guess its fine.

kushal-banik-hyland (approved):

damianujma (approved):

PR #4284, opened 2026-08-12
ACS-12299 updated v1 namespace-prefix api

bsayan2:
## Summary

Implements a modern v1 REST API endpoint to retrieve the complete namespace URI ↔ prefix mapping from Alfresco's NamespaceService. This replaces the legacy v0 web script implementation with a clean, maintainable REST endpoint following the standard Alfresco REST API pattern used by Types, Aspects, and Actions.

## API

**Endpoint:** `GET /alfresco/api/-default-/public/alfresco/versions/1/namespace-prefix`

**Response:**
```json
{
  "list": {
    "pagination": {
      "count": 1,
      "hasMoreItems": false,
      "totalItems": 1
    },
    "entries": [
      {
        "entry": {
          "prefixUriMap": {
            "http://www.alfresco.org/model/content/1.0": "cm",
            "http://www.alfresco.org/model/system/1.0": "sys"
          }
        }
      }
    ]
  }
}

### CODE REVIEW

File: remote-api/src/main/java/org/alfresco/rest/api/impl/NamespacePrefixesImpl.java [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+                String uri = namespaceService.getNamespaceURI(prefix);
+
+                if (uri == null)
+                {
+                    LOGGER.debug("Skipping prefix '{}' with null URI", prefix);
+                    continue;
+                }
+
+                String existing = prefixUriMap.put(uri, prefix);
+                if (existing != null && !existing.equals(prefix))
```
dpatel-alfresco, line 77: You can check if this can be the real case or not and decide keep or not

remote-api/src/main/java/org/alfresco/rest/api/namespace/NamespacePrefixesEntityResource.java
```diff
@@ -0,0 +1,63 @@
...
+import org.alfresco.rest.framework.resource.EntityResource;
+import org.alfresco.rest.framework.resource.actions.interfaces.EntityResourceAction;
+import org.alfresco.rest.framework.resource.parameters.CollectionWithPagingInfo;
+import org.alfresco.rest.framework.resource.parameters.Parameters;
+import org.alfresco.util.ParameterCheck;
+
+/**
+ * REST API v1 endpoint for namespace-prefix mapping.
+ */
+@EntityResource(name = "namespace-prefix", title = "Namespace Prefixes")
```
dpatel-alfresco, line 42: The name is not similar to PR description, so please update OR description

dpatel-alfresco (approved):
Looks good to me just added some comment for minor changes

remote-api/src/main/java/org/alfresco/rest/api/namespace/NamespacePrefixesEntityResource.java
bsayan2: PR description updated, Thanks for noticing

remote-api/src/main/java/org/alfresco/rest/api/impl/NamespacePrefixesImpl.java
bsayan2: Might not be a real case , however if such occurrence happens it will log the info .

AbhraPoddar (approved):

PR #4296, opened 2026-08-14
[MNT-25111] Shared Links API with where clause does not work with Search Enterprise

ShreasiK:
https://hyland.atlassian.net/browse/MNT-25111


repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/ElasticsearchSearchService.java
```diff
@@ -115,6 +123,7 @@ public ResultSet query(StoreRef store, QName queryId, QueryParameter[] queryPara
     public ResultSet query(SearchParameters searchParameters)
     {
         adjustSearchParameters(searchParameters);
+        parameteriseQuery(searchParameters);
```
dpatel-alfresco, line 122: Can you check the similar method is also available in SolrSearchService. Is it possible to share one implementation for both the classes ?

dpatel-alfresco (commented):
I have added one suggestion please check that and 
also please check if we can add more test cases like missing parameter definition , definition without a default and no value and unclosed ${...} placeholder

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/ElasticsearchSearchService.java
ShreasiK: Done

dpatel-alfresco (dismissed):

bsayan2 (dismissed):

dpatel-alfresco (approved):

bsayan2 (approved):

PR #4297, opened 2026-08-14
[MNT-25799] Fix Document Library tag filter for tags containing spaces

SanjoyHyland2025:
The Share Document Library tag filter returned no results (or, after intermediate fixes, every document) when filtering by a tag whose value contains a space, e.g. "long tag".

Root cause: the slingshot doclist webscript built the filter query as +TAG:"<ISO9075(value)>" using the legacy "lucene" language. ISO9075 encoding turned the space into "_x0020_" (matching nothing), and querying the TAG text field with a quoted multi-word phrase is not reliable on the Elasticsearch connector (the mandatory clause is dropped, so every document is returned).

Fix: resolve the tag to its category node under /cm:categoryRoot/cm:taggable (by ISO9075-encoded, lower-cased local name) and match tagged content by that exact nodeRef using the AFTS exact-match operator: +=cm:taggable:"<nodeRef>" with the "fts-alfresco" language. This mirrors the existing "category" filter, is immune to tag-text tokenisation, and works on both Solr and Elasticsearch. An unresolved tag now returns no results instead of every document.

FYI:

https://github.com/user-attachments/assets/9f94208c-1148-447a-b001-e4c07cb1f588

### DIFF
--- a/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
+++ b/amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
@@ -233,15 +233,30 @@ var Filters =
             filterParams.query = "+ID:\"" + parsedArgs.nodeRef + "\"";
             break;
 
-          case "tag":
-              // Remove any trailing "/" character
-              if (filterData.charAt(filterData.length - 1) == "/")
-              {
-                  filterData = filterData.slice(0, -1);
-              }
-              filterQuery = this.constructPathQuery(parsedArgs);
-              filterParams.query = filterQuery + " +TAG:\"" + search.ISO9075Encode(filterData) + "\"";
-              break;
+         case "tag":
+            // Remove any trailing "/" character
+            if (filterData.charAt(filterData.length - 1) == "/")
+            {
+               filterData = filterData.slice(0, -1);
+            }
+            filterParams.language = "fts-alfresco";
+            var encodedTag = search.ISO9075Encode(filterData.toLowerCase());
+            var tagNodes = search.luceneSearch('+PATH:"/cm:categoryRoot/cm:taggable//cm:' + encodedTag + '"');
+            if (tagNodes && tagNodes.length > 0)
+            {
+               filterParams.query = '+=cm\\:taggable:"' + tagNodes[0].nodeRef.toString() + '"';
+            }
+            else
+            {
+               // Unknown tag - return no results rather than every document
+               return {
+                  query: null,
+                  limitResults: 0,
+                  sort: [],
+                  language: "fts-alfresco"
+               };
+            }
+            break;
 
           case "category":
 
--- a/amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
+++ b/amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
@@ -76,7 +76,7 @@ public class FiltersLibTest
     private Scriptable scope;
 
     @Before
-    public void setUp() throws Exception
+    public void setUp() throws IOException
     {
         cx = Context.enter();
         // Interpretive mode - no bytecode generation required for a simple script evaluation
@@ -198,7 +198,7 @@ public class FiltersLibTest
         return "'" + value.replace("\\", "\\\\").replace("'", "\\'") + "'";
     }
 
-    private String readResource(String path) throws Exception
+    private String readResource(String path) throws IOException
     {
         try (InputStream is = getClass().getClassLoader().getResourceAsStream(path))
         {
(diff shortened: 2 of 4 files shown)


amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
```diff
@@ -233,15 +233,30 @@ var Filters =
...
+            if (filterData.charAt(filterData.length - 1) == "/")
+            {
+               filterData = filterData.slice(0, -1);
+            }
+            filterParams.language = "fts-alfresco";
+            var encodedTag = search.ISO9075Encode(filterData.toLowerCase());
+            var tagNodes = search.luceneSearch('+PATH:"/cm:categoryRoot/cm:taggable//cm:' + encodedTag + '"');
+            if (tagNodes && tagNodes.length > 0)
+            {
+               filterParams.query = '+=cm\\:taggable:"' + tagNodes[0].nodeRef.toString() + '"';
```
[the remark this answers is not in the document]
SanjoyHyland2025: The concern is understandable (the previous code did prepend constructPathQuery), but restoring that scoping re-introduces the bug this PR fixes. On the Elasticsearch connector, composing a +PATH:"…//*" scope with the +=cm:taggable:"<nodeRef>" match causes the connector to drop the mandatory tag constraint, so the query returns every node under the path.
Verified live with tag space tag11:
=cm:taggable:"workspace://SpacesStore/0363e859-…" → 1 result (the correctly tagged node).
+PATH:"/app:company_home//*" +=cm:taggable:"workspace://SpacesStore/0363e859-…" → 429 results (all content under the path).

Somnath-Deshmukh (changes requested):
I see no test coverage has been added for this. Can we please add some tests for this?

amps/share-services/src/main/resources/alfresco/templates/webscripts/org/alfresco/slingshot/documentlibrary-v2/filters.lib.js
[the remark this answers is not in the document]
Somnath-Deshmukh: Have we tested this with other sites?

amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
```diff
@@ -0,0 +1,208 @@
...
+        Scriptable result = runTagFilter("mytag", false);
+
+        assertEquals("fts-alfresco", getString(result, "language"));
+        assertEquals("+=cm\\:taggable:\"" + TAG_NODEREF + "\"", getString(result, "query").trim());
+        // The tag value is looked up under the category root, ISO9075-encoded
+        assertEquals("+PATH:\"/cm:categoryRoot/cm:taggable//cm:mytag\"", getLastLuceneQuery());
+    }
+
+    /**
+     * MNT-25799: a tag containing a space must be ISO9075-encoded (space -> _x0020_) when the tag node is located, and must still yield a valid membership query.
```
AbhraPoddar: Please remove MNT number . Its not meaningful much here. try to provide an example in this comment like query=long tag - then this could be easily understandable in future.

amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
```diff
@@ -0,0 +1,208 @@
...
+    public void tearDown()
+    {
+        Context.exit();
+    }
+
+    /**
+     * Baseline: a single-word tag resolves the tag node and produces a well-formed {@code +=cm:taggable:"<nodeRef>"} membership query using the fts-alfresco language.
+     */
+    @Test
+    public void tagQueryForSingleWordTag()
```
AbhraPoddar, line 100: Just a suggestion. Can we provide some specific edge cases related tests?  It's understandable that single word may cater - **Long_Tag , Long-Tag, Tom&Jerry , Tag#1, Tag$1, Base@India** etc. But it will be good if we segregate all these types of test cases separately (for better understanding), make sure these types of edge cases can be handled (please think about end user point of view). If those test cases will be implemented, provide also a good comment against each test case. Since this test class is newly introduced, we should adhere best practices as much we can.

amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
SanjoyHyland2025: updated

amps/share-services/src/test/java/org/alfresco/slingshot/web/scripts/FiltersLibTest.java
SanjoyHyland2025: done

Somnath-Deshmukh (dismissed):
LGTM

AbhraPoddar (dismissed):

jottley (commented):
Looks Good. Could we add an integration test so that we can prove it working against an elastic server?

SanjoyHyland2025:
> Looks Good. Could we add an integration test so that we can prove it working against an elastic server?

added the integration test for this changes

AbhraPoddar (approved):

Somnath-Deshmukh (approved):

debjit-chattopadhyay-hyland (approved):

jottley (approved):

PR #4314, opened 2026-08-18
ACS-12541 Configure safe-outputs.create-pull-request.allowed-base-branches for backport workflow

damianujma:


gionn (commented):
try also adding:

```
checkout:
  - fetch-depth: 0
```

PR #4322, opened 2026-08-19
ACS-11473 Switch nodesSharedCache to invalidating and fix stale cache in rendition backport 25.N

SatyamSah5:
Cherry Picked https://github.com/Alfresco/alfresco-community-repo/commit/104ed1c3fa461ebbca25d9c7e6c429ed1580e3fa


repository/src/main/java/org/alfresco/repo/rendition2/RenditionService2Impl.java
```diff
@@ -350,6 +359,19 @@ private void requestAsyncTransformOrRendition(NodeRef sourceNodeRef, RenderOrTra
...
+                            // ContentData (size=0) because cache invalidation from the uploading node
+                            // hasn't propagated yet. Bypass the shared cache to read directly from DB.
+                            nodeDAO.setCheckNodeConsistency();
+                            sourceContentData = getSourceContentData(sourceNodeRef);
+                            if (sourceContentData != null && sourceContentData.getSize() > 0)
+                            {
+                                sourceContentHashCode = getSourceContentHashCode(sourceContentData);
+                                renderOrTransform.throwIllegalStateExceptionIfAlreadyDone(sourceContentHashCode);
+                            }
+                        }
```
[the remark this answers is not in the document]
SatyamSah5: This scenario fixed here is very rare and happens only during the small time window between an update being made and the corresponding cache invalidation event being received by that node.
checkSupported() with size=0 is the most permissive check (size limits are upper bounds), so any transform that's unsupported at the real size would also be unsupported at size=0 — we'd never reach this code path. Additionally, the transform service re-validates size limits during actual transformation, so there's no risk of an invalid transform proceeding.

mousumi-podder (approved):

repository/src/test/java/org/alfresco/repo/rendition2/RenditionService2Test.java
```diff
@@ -293,4 +300,45 @@ public void overriddenRendition()
...
+    {
+        ContentData emptyContentData = new ContentData("store://empty-url", "application/pdf", 0L, "UTF-8");
+
+        // Always return 0 size — this is genuinely empty content, not stale cache
+        when(nodeService.getProperty(nodeRef, ContentModel.PROP_CONTENT)).thenReturn(emptyContentData);
+
+        renditionService2.render(nodeRef, TEST_RENDITION);
+
+        verify(nodeDAO).setCheckNodeConsistency();
+        verify(transformClient, times(1)).transform(any(), any(), nullable(String.class), anyInt());
```
kushal-banik-hyland, line 342: Can we check if the renderTransform throwIllegalStateExceptionIfAlreadyDone is not being called!

kushal-banik-hyland (approved):

repository/src/test/java/org/alfresco/repo/rendition2/RenditionService2Test.java
SatyamSah5: this test is scoped to verifying that a genuinely 0-byte source still attempts the transform, which the existing transform(...) verify already covers. I dont think this is needed.

PR #4392, opened 2026-09-10
[ACS-11964][run tests]-Replace package sweep with explicit class list in elasticsearch-e2e-part2-suite

mousumi-podder:
This pull request simplifies the `elasticsearch-e2e-part2-suite.xml` test suite configuration by removing package-level sweeping and explicit suppression of certain test classes. The main effect is to make the suite more explicit about which tests are run, reducing the risk of unintended test executions.

Test suite configuration simplification:

* Removed the `<packages>` section, which previously included all tests in `org.alfresco.rest.search.*` except for two subpackages, in favor of explicitly listing test classes.
* Eliminated explicit suppression of "part1" test classes by removing `<class ...><methods><exclude name=".*"/></methods></class>` entries, ensuring only the intended test classes are referenced.

### DIFF
--- a/packaging/tests/tas-restapi/src/test/resources/test-suites/elasticsearch-e2e-part2-suite.xml
+++ b/packaging/tests/tas-restapi/src/test/resources/test-suites/elasticsearch-e2e-part2-suite.xml
@@ -7,12 +7,6 @@
         <listener class-name="org.alfresco.utility.testng.OSTestMethodSelector"/>
     </listeners>
     <test name="SearchTests-Part2">
-        <packages>
-            <package name="org.alfresco.rest.search.*">
-                <exclude name="org.alfresco.rest.search.fingerprint"/>
-                <exclude name="org.alfresco.rest.search.crosslocale"/>
-            </package>
-        </packages>
         <classes>
             <!-- 1) FacetRangeSearchTest: range facet gaps on ES -->
             <class name="org.alfresco.rest.search.FacetRangeSearchTest">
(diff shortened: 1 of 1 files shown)


packaging/tests/tas-restapi/src/test/resources/test-suites/elasticsearch-e2e-part2-suite.xml
```diff
@@ -7,12 +7,6 @@
         <listener class-name="org.alfresco.utility.testng.OSTestMethodSelector"/>
     </listeners>
     <test name="SearchTests-Part2">
-        <packages>
-            <package name="org.alfresco.rest.search.*">
-                <exclude name="org.alfresco.rest.search.fingerprint"/>
-                <exclude name="org.alfresco.rest.search.crosslocale"/>
-            </package>
-        </packages>
```
jakubkochman, line 15: Without the package it may miss the following classes, can you double check if we need to add them explicitly?

```
<!-- 10) SearchQueryPaginationTest: will include all the tests -->
<class name="org.alfresco.rest.search.SearchQueryPaginationTest"/>
<!-- 11) SearchPermissionsTest: will include all the tests -->
<class name="org.alfresco.rest.search.SearchPermissionsTest"/>
<!-- 12) SearchParentFieldTest: will include all the tests -->
<class name="org.alfresco.rest.search.SearchParentFieldTest"/>
<!-- 13) SearchSecondaryAssociationTest: will include all the tests -->
<class name="org.alfresco.rest.search.SearchSecondaryAssociationTest"/>
<!-- 14) SpecialCharacterSearchTest: will include all the tests -->
<class name="org.alfresco.rest.search.SpecialCharacterSearchTest"/>
<!-- 15) ExplicitRoutingTest: Solr-only explicit routing; also needs runtime custom model, ACS-12502 -->
<class name="org.alfresco.rest.search.ExplicitRoutingTest">
    <methods><exclude name=".*"/></methods>
</class>
<!-- 16) tracker.CascadingIntegrationTest: Solr tracker cascade; also needs runtime custom model, ACS-12502 -->
<class name="org.alfresco.rest.search.tracker.CascadingIntegrationTest">
    <methods><exclude name=".*"/></methods>
</class>
```

mousumi-podder: added it as suggested.Thanks @jakub.

jakubkochman: Instead of setting each class we could keep the package, but split the existing classes differently. If we rely on explicit class names it's very possible that we miss one test class from being executed. It's bigger problem in my opinion than having the same test execute as part of 2 test suites.

For example org.alfresco.rest.search.facets to have inside: FacetRange/Interval/Fields, Faceted, Pivot, Stats, FieldDefinition, ShardInfo, SearchAFTSInField. Then part one could take `<packages><package name="org.alfresco.rest.search"/></packages>`
and part two `<packages><package name="org.alfresco.rest.search.facets"/></packages>`. Because there is no `.*` , part one would not include sub-packages.

It does not even need to be `facets`, but explicit `part1`, `part2` package names would also work.

damianujma: @mousumi-podder can we use part1, part2 packages names?

mousumi-podder: yes @damianujma.
In that case, I will close this PR and create new PR with a new ACS task id to track this entire restructuring process to be tracked properly.

damianujma (approved):

PR #4399, opened 2026-09-17
ACS-12695 Search API: enable trashcan search via the `deleted-nodes` scope

b-ansari:


gerardolenski (changes requested):
I've reviewed the changes and would like to propose a slightly different approach.

Let's first look at the current behavior in the Community Edition:

- The ALIVE=true filter is not applied when building queries for Elasticsearch.
- When a user performs a search using the deleted-nodes scope, the API currently returns HTTP 500 Internal Server Error.

My suggestion is to align the behavior with the capabilities available in the Community Edition:

- Always apply the ALIVE=true filter in Community Edition searches.
- Since the deleted-nodes scope is not supported in Community Edition, the API should return HTTP 501 Not Implemented along with a clear and descriptive error message explaining the limitation.

This approach would make the API behavior more predictable and provide clearer feedback to consumers about unsupported functionality.

b-ansari:
> I've reviewed the changes and would like to propose a slightly different approach.
> 
> Let's first look at the current behavior in the Community Edition:
> 
> * The ALIVE=true filter is not applied when building queries for Elasticsearch.
> * When a user performs a search using the deleted-nodes scope, the API currently returns HTTP 500 Internal Server Error.
> 
> My suggestion is to align the behavior with the capabilities available in the Community Edition:
> 
> * Always apply the ALIVE=true filter in Community Edition searches.
> * Since the deleted-nodes scope is not supported in Community Edition, the API should return HTTP 501 Not Implemented along with a clear and descriptive error message explaining the limitation.
> 
> This approach would make the API behavior more predictable and provide clearer feedback to consumers about unsupported functionality.

yes, I agree to the above points.
Also, will introduce the AliveStateFilter for community defaulting to always true and can be implemented in enterprise as required.

remote-api/src/main/java/org/alfresco/rest/api/search/impl/SearchMapper.java
```diff
@@ -108,6 +109,9 @@ public class SearchMapper
     public static final String AFTS = "afts";
     private StoreMapper storeMapper;
 
+    /** Whether the {@code deleted-nodes} (trashcan) search scope is supported. Overridden by editions that support it. */
```
gerardolenski, line 112: I don't think this comment is necessary, as the field name is already self-explanatory.

remote-api/src/main/java/org/alfresco/rest/api/search/impl/SearchMapper.java
```diff
@@ -504,6 +508,12 @@ public void fromScope(SearchParameters sp, Scope scope, SearchRequestContext sea
             List<String> stores = scope.getLocations();
             if (stores != null && !stores.isEmpty())
             {
+                if (stores.contains(StoreMapper.DELETED) && !deletedNodesScopeSupported)
```
gerardolenski, line 511: 1. I suggest switching the order of these conditions. When the feature is enabled, evaluating the store-related condition is unnecessary.

```
if (!deletedNodesScopeSupported && stores.contains(StoreMapper.DELETED))
```

2. Can you extract entire if statement to private method, eg: validateDeletedNodesScopeSupported, checkDeletedNodesScopeSupported or sth simillar?

remote-api/src/test/java/org/alfresco/rest/api/search/SearchMapperTests.java
```diff
@@ -562,6 +567,29 @@ public void fromScope() throws Exception
...
+    @Test
+    public void fromScope_deletedNodesScopeNotSupported()
+    {
+        SearchMapper communitySearchMapper = new SearchMapper();
+        communitySearchMapper.setStoreMapper(new StoreMapper());
+        SearchParameters searchParameters = new SearchParameters();
+        communitySearchMapper.setDefaults(searchParameters);
+        SearchRequestContext searchRequestContext = SearchRequestContext.from(minimalQuery());
+
+        try
```
gerardolenski, line 584: Is there a reason for implementing it this way? Using Assertions.assertThatThrownBy(...) would make the test more concise and easier to understand.

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/query/filter/AliveStateFilterBuilder.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2026 Alfresco Software Limited
```
gerardolenski, line 5: It is new class, so we should have: Copyright (C) **2026** Alfresco Software Limited. Please also check otehr classes in your PR.

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/query/filter/AliveStateFilterBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+ * #L%
+ */
+package org.alfresco.repo.search.impl.elasticsearch.query.filter;
+
+import org.opensearch.client.opensearch._types.query_dsl.Query;
+
+import org.alfresco.service.cmr.search.SearchParameters;
+
+/**
+ * The default (Community) implementation always restricts results to live nodes.
```
gerardolenski, line 33: This Javadoc does not accurately describe the purpose of this interface.

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/query/SearchRequestBuilderService.java
```diff
@@ -229,8 +229,8 @@ public String getElasticIndex(List<StoreRef> stores)
         }
         return switch (stores.get(0).getProtocol())
         {
-        case StoreRef.PROTOCOL_WORKSPACE -> httpClientFactory.getIndexName();
-        case StoreRef.PROTOCOL_ARCHIVE -> httpClientFactory.getArchiveIndexName();
+        // ACS-12695: archive/deleted-nodes scope now targets the single unified "alfresco" index rather than the separate legacy archive index.
```
gerardolenski, line 232: We recently discussed our policy regarding comments in production code, including avoiding ticket numbers in comments. Could you please pay a bit more attention to this?

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/resultset/ElasticsearchResultSetBuilder.java
```diff
@@ -140,13 +141,14 @@ public ElasticsearchResultSet build(SearchParameters searchParameters, List<Hit<
                 null);
     }
 
-    private List<NodeRefAndScore> mapNodeRefsAndScores(List<Hit<Object>> hits, boolean isBulkFetchEnabled)
+    private List<NodeRefAndScore> mapNodeRefsAndScores(SearchParameters searchParameters, List<Hit<Object>> hits, boolean isBulkFetchEnabled)
```
gerardolenski, line 144: I see an issue with this refactoring. You added the searchParameters parameter, but isBulkFetchEnabled is also passed in even though it is derived directly from searchParameters.

To avoid redundant parameters and keep the API consistent, I think the method signature should be:
```
mapNodeRefsAndScores(List<Hit<Object>> hits, SearchParameters searchParameters)
```

The method can then determine isBulkFetchEnabled internally from searchParameters. This keeps the signature simpler and avoids passing the same information twice.

repository/src/main/java/org/alfresco/repo/search/impl/elasticsearch/resultset/ElasticsearchResultSetBuilder.java
```diff
@@ -155,6 +157,17 @@ private List<NodeRefAndScore> mapNodeRefsAndScores(List<Hit<Object>> hits, boole
         return results;
     }
 
+    /**
+     * ACS-12695: hits now come from a single unified index, so the store can no longer be assumed to always be the workspace store - archive ("deleted-nodes") scope results must resolve to the archive store so that restore-by-id and other archive-only flows keep working.
+     */
+    static StoreRef resolveStoreRef(SearchParameters searchParameters)
+    {
+        return searchParameters.getStores().stream()
```
gerardolenski, line 165: I cannot agree with this implementation because I find it misleading.

For Elasticsearch, the contract implicitly assumes exactly one store. This is clearly visible in StandardSearchStrategy#executeSearch, where requestBuilderService.getElasticIndex(...) is invoked. The implementation of that method assumes that SearchParameters contain exactly one store.

Could you adjust this implementation to reflect that contract more explicitly? Otherwise, the current approach may suggest that multiple stores are supported, while the underlying Elasticsearch execution path does not actually handle that scenario.

---
Just an optional suggestion for consideration: it might be worth extracting a dedicated component responsible for resolving the store and index from SearchParameters.Such a component could then be reused across:
- SearchRequestBuilderService
- ElasticsearchResultSetBuilder
- EnterpriseAliveStateFilterBuilder
- HighlightsHandler

This would centralize the logic in a single place, reduce duplication, and help keep the behavior consistent across all Elasticsearch-related components. You can investigate if this is possible and worth to do.

gerardolenski (changes requested):

